PR #6, opened 2021-10-20
Modal Models

a-sr:
See [LF PR #501](https://github.com/lf-lang/lingua-franca/pull/501)


edwardalee:
Where are we with this modal-models PR? Can we merge?

lhstrh (approved):
This looks pretty solid to me, but I think that some things in `reactor_common.c` would be better split off into a separate `modal.c`. I'm OK with merging this now and doing a refactoring later.

PR #9, opened 2021-10-21
Docs

lhstrh:
Commits authored by @arengarajan99 for documentation generation.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+  #    - '**'
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: macos-latest
```
lhstrh, line 15: In my experience, `ubuntu-latest` runs faster.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+    - name: Requirements
+      run: brew install doxygen
```
lhstrh, line 19: This way, you run a single command, which if it fails provides less clear output. You can split this in multiple commands using `run: |` followed by a newline and a single command for each subsequent line.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Build docs
+      run: cd docs
+        && make doxygen
+        && cd _build/html
+        && touch .nojekyll
+    - name: Deploy
+      uses: JamesIves/github-pages-deploy-action@releases/v3
+      with:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        BRANCH: gh-pages # The branch the action should deploy to.
```
lhstrh, line 36: This is the most important line. I've once accidentally pushed deployment stuff like this to the main branch and you can imagine the result :-)

lhstrh (commented):
This looks great! I left some minor comments.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
+name: Docs
+
+on:
+  #push:
+  #  branches-ignore:
+  #    - '**'
+  push:
+    branches: [ main ]
+  pull_request:
```
lhstrh, line 9: We should not run this for PRs
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+    branches: [ main ]
```
lhstrh, line 10:
```suggestion
```

PR #13, opened 2021-11-05
First attempt at refactoring

lhstrh:


core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
-    lf_mutex_lock(&mutex);
+bool _lf_worker_should_stop() {
+    // If this is not the very first step, notify that the previous step is complete
+    // and check against the stop tag to see whether this is the last step.
+    if (_lf_logical_tag_completed) {
+        logical_tag_complete(current_tag);
+        // If we are at the stop tag, do not call _lf_next()
+        // to prevent advancing the logical time.
+        if (compare_tags(current_tag, stop_tag) >= 0) {
+            // Break out of the while loop and notify other
```
edwardalee, line 1027: There is no while loop here to break out of.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+            return true;
+        }
+    }
+    return false;
+}
 
-    int worker_number = ++worker_thread_count;
-    LOG_PRINT("Worker thread %d started.", worker_number);
+/**
+ * Advance tag. This will also pop events for the newly acquired tag and puts
```
edwardalee, line 1041:
```suggestion
 * Advance tag. This will also pop events for the newly acquired tag and put
```

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+ * the triggered reactions on the reaction queue.
+ * 
+ * @return should_exit True if the worker thread should exit. False otherwise.
+ */
+bool _lf_worker_advance_tag(int worker_number) {
+    // Block other worker threads from doing that.
```
edwardalee, line 1047:
```suggestion
    // Block other worker threads from also advancing the tag.
```

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
 bool _lf_logical_tag_completed = false;
 
 /**
- * Worker thread for the thread pool.
- * This acquires the mutex lock and releases it to wait for time to
- * elapse or for asynchronous events and also releases it to execute reactions.
+ * Return true if the worker should stop now; false otherwise.
```
edwardalee, line 1017: Need to document whether this function assumes that the mutex is held. I think it does, in which case, we have by convention been adding the suffix `_locked` to the function name.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+        }
+    }
+    return false;
+}
 
-    int worker_number = ++worker_thread_count;
-    LOG_PRINT("Worker thread %d started.", worker_number);
+/**
+ * Advance tag. This will also pop events for the newly acquired tag and puts
+ * the triggered reactions on the reaction queue.
```
edwardalee, line 1042: Again, should the lock be held when this is called?

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+    _lf_next();
+    tracepoint_worker_advancing_time_ends(worker_number);
+    _lf_advancing_time = false;
+    DEBUG_PRINT("Worker %d: Done waiting for _lf_next().", worker_number);
+    return false;
+}
+
+/**
+ * Advance tag if there are no reactions in the reaction queue or in progress. If
+ * there are such reactions or if another thread is already advancing time, wait
```
edwardalee, line 1070:
```suggestion
 * there are such reactions or if another thread is already advancing the tag, wait
```

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+ * until something on the reaction queue is changed.
+ * 
+ * @return should_exit True if the worker thread should exit. False otherwise.
+ */
+bool _lf_worker_try_advance_tag_or_wait(int worker_number) {
```
edwardalee, line 1075: Need to document whether the mutex lock is assumed held.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+
+    return false;
+}
+
+/**
+ * Handle deadline violation for 'reaction'.
+ * 
+ * @return true if a deadline violation occurred. false otherwise.
+ */
+bool _lf_worker_handle_deadline_violation_for_reaction(int worker_number, reaction_t* reaction) {
```
edwardalee, line 1121: I'm guess this function must be called when the mutex lock is NOT held. This should be documented.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+    }
+    return violation_occurred;
+}
+
+/**
+ * Handle STP violation for 'reaction'.
+ * 
+ * @return true if an STP violation occurred. false otherwise.
+ */
+bool _lf_worker_handle_STP_violation_for_reaction(int worker_number, reaction_t* reaction) {
```
edwardalee, line 1158: Same issue.  Mutex lock probably has to be NOT held when this is called.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+    return violation_occurred;
+}
+
+/**
+ * Handle violations for 'reaction'. Currently limited to deadline violations
+ * and STP violations.
+ * 
+ * @return true if a violation occurred. false otherwise.
+ */
+bool _lf_worker_handle_violations(int worker_number, reaction_t* reaction) {
```
edwardalee, line 1203: Should the mutex be held?

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+    violation = _lf_worker_handle_deadline_violation_for_reaction(worker_number, reaction) ||
+                    _lf_worker_handle_STP_violation_for_reaction(worker_number, reaction);
+    return violation;
+}
+
+/**
+ * Invoke 'reaction' and schedule any resulting triggered reaction(s) on the
+ * reaction queue.
+ */
+void _lf_worker_invoke_reaction(int worker_number, reaction_t* reaction) {
```
edwardalee, line 1215: Document that mutex MUST NOT be held when this is called.

core/reactor_threaded.c
```diff
@@ -1014,18 +1014,227 @@ int worker_thread_count = 0;
...
+    // reactions into the queue or execute them immediately.
+    schedule_output_reactions(reaction, worker_number);
+}
 
+/**
+ * The main looping logic of each LF worker thread.
+ * 
+ * @param worker_number The number assigned to this worker thread
+ */
+void _lf_worker_do_work(int worker_number) {
```
edwardalee, line 1235: Should the mutex be held?

edwardalee (changes requested):
It is not possible for me to tell whether the sequence of mutex locks/unlocks is preserved in the refactoring, and the new functions all need to document their assumptions about whether the mutex is held when they are called.

Soroosh129:
> It is not possible for me to tell whether the sequence of mutex locks/unlocks is preserved in the refactoring

Other than documenting the assumptions about the mutex for each refactored function, is there anything I can do to make this sequence clearer?

edwardalee:
No need to add anything here. I was just saying that I couldn't (easily) check that the refactored does the same thing that the pre-refactoring code did.  But anyway, I'm confident you got it right... Just saying that I didn't check it.

Soroosh129:
> No need to add anything here. I was just saying that I couldn't (easily) check that the refactored does the same thing that the pre-refactoring code did. But anyway, I'm confident you got it right... Just saying that I didn't check it.

Okay, thank you for pointing out the issues in the refactoring.  I think I fixed them all.

Since all the tests pass in lf-lang/lingua-franca#728, we could have some confidence that the refactoring didn't introduce a major issue.

PR #14, opened 2021-11-10
Optimized ports reset

petervdonovan:
This is an optimization that only is beneficial when the set of all `is_present` values that need to be reset at the start of the next tag is sparse (i.e., mostly `false`).

Discussion is in [the PR at the main repository](https://github.com/lf-lang/lingua-franca/pull/736).


lhstrh (approved):
This looks great! I only left some suggestions for improving the comments.

PR #15, opened 2021-11-10
Avoid searching reaction queue.

petervdonovan:
The changes in this branch allow us to avoid having to search for reactions in the reaction queue in order to determine their status.

### DIFF
--- a/core/reactor.h
+++ b/core/reactor.h
@@ -331,6 +331,20 @@ typedef enum {no=0, token_and_value, token_only} ok_to_free_t;
 typedef enum {absent = false, present = true, unknown} port_status_t;
 
 /**
+ * Status of a given reaction at a given logical time.
+ *
+ * If a reaction is 'inactive', it is neither running nor queued.
+ * If a reaction is 'queued', it is going to be executed at the current logical time,
+ * but it has not started running yet.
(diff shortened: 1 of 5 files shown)


core/reactor.h
```diff
@@ -330,6 +330,20 @@ typedef enum {no=0, token_and_value, token_only} ok_to_free_t;
...
+ * Status of a given reaction at a given logical time.
+ *
+ * If a reaction is 'inactive', it is neither running nor queued.
+ * If a reaction is 'queued', it is going to be executed at the current logical time,
+ * but it has not started running yet.
+ * If a reaction is 'running', its body is being executed.
+ *
+ * @note inactive must equal zero because it should be possible to allocate a reaction
+ *  with default values using calloc.
+ * FIXME: The running state does not seem to be read.
```
lhstrh, line 343: Curious! This suggests we could just use a `bool`, but it shouldn't matter for performance. Let's leave it like this for now...

lhstrh (approved):

petervdonovan:
As for preliminary conclusions regarding performance impact: On my machine, with 3 threads, the mean execution time for Chameneos decreases from 4.9 seconds to 3.2 seconds, in both cases with a standard deviation of not more than a few hundred milliseconds.

PR #19, opened 2021-11-24
Remove redundant check for reaction->deadline > 0LL

CloverCho:
I think this check already exists in line 185, so I remove redundant check.

Also, I have a question about comments below (where in line 188~193).
            // Check for deadline violation.
            // There are currently two distinct deadline mechanisms:
            // local deadlines are defined with the reaction;
            // container deadlines are defined in the container.
            // They can have different deadlines, so we have to **check both**.
            // Handle the local deadline first.

Although the comments say we check "both", there is only one check, whichis "physical_time > current_tag.time + reaction->deadline" on line 194.
Is this check for local deadline or container deadline?
If so,do we need to update this comment as well?

### DIFF
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -191,7 +191,7 @@ int _lf_do_step() {
             // container deadlines are defined in the container.
             // They can have different deadlines, so we have to check both.
             // Handle the local deadline first.
-            if (reaction->deadline > 0LL && physical_time > current_tag.time + reaction->deadline) {
+            if (physical_time > current_tag.time + reaction->deadline) {
                 LOG_PRINT("Deadline violation. Invoking deadline handler.");
                 // Deadline violation has occurred.
                 violation = true;
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -176,6 +176,7 @@ int _lf_do_step() {
 
         bool violation = false;
 
+        // FIXME: These comments look outdated. We may need to update them.
         // If the reaction has a deadline, compare to current physical time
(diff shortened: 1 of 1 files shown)


core/reactor.c
```diff
@@ -191,7 +191,7 @@ int _lf_do_step() {
             // container deadlines are defined in the container.
             // They can have different deadlines, so we have to check both.
             // Handle the local deadline first.
-            if (reaction->deadline > 0LL && physical_time > current_tag.time + reaction->deadline) {
+            if (physical_time > current_tag.time + reaction->deadline) {
```
lhstrh, line 194: Not every reaction has a deadline. I suspect that the first check is there to make sure deadlines are only enforced for reactions that have one, but I'm not sure. The answer can be found in the code generator...

lhstrh: But you're right that it is redundant given the check on line 185. As for the comment, I think it is outdated. We used to have different kinds of deadlines but discontinued other kinds we experimented with due to their complexity.

CloverCho: Thank you for your response!
I will search about deadline in the code generator.
Thank you!

hokeun: @edwardalee Do you think you could take a look at the comments and let us if we need to update these comments? Maybe the two deadlines are not applicable anymore? Thanks!

edwardalee:
Yes, this comment is obsolete. Please remove.

hokeun:
> Yes, this comment is obsolete. Please remove.

Thanks, Edward, for confirming!

PR #21, opened 2021-12-07
Unit tests

lhstrh:
Attempt to get unit tests to work


petervdonovan:
The toy example seems to run now, and things should not be broken if you check out the `c-unit-tests` branch in Lingua Franca. Before adding more tests, I still need to clean up some of the changes I just pushed.

petervdonovan:
I failed to cleanly revert a merge, and now this branch is too complicated. This PR is closed so that the relevant changes can be cherry-picked into a clean new branch.

PR #22, opened 2021-12-19
Resizing array

petervdonovan:
This is the version currently in `c-new-scheduler`, minus two useful features, plus one cleanup. The two missing features are hard to cherry-pick, so they can be added later, when `c-new-scheduler` gets merged.

Versions of this data structure are in two different branches. The intent of this PR is to make the data structure accessible while we wait for one of those branches to be merged.

### DIFF
--- a/core/utils/vector.h
+++ b/core/utils/vector.h
@@ -1,7 +1,6 @@
-/*
- * This file defines a minimal vector (resizing array) data type.
- * It is intended to be the simplest way of storing a collection of
- * pointers that is frequently filled and then completely emptied.
+/**
+ * @author Peter Donovan (peterdonovan@berkeley.edu)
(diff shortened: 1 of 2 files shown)


core/utils/vector.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Remove and return some pointer that is contained in the given vector,
+ * or return NULL if the given vector is empty.
+ * @param v Any vector.
+ */
+void* vector_pop(vector_t* v);
+
+/**
+ * Vote on whether this vector ought to have a smaller memory footprint.
+ */
```
edwardalee, line 61: It is not clear from this comment what the effect of calling this function is.

edwardalee (approved):
This looks very useful to me.  I suggest adding author info, and, if the code is inspired by or derived from something out there, a citation to sources.

PR #23, opened 2021-12-19
Unit tests

petervdonovan:
This PR includes changes to the structure of reactor-c. The intention is to make unit testing easier.

For this to work, it appears that we will indeed have to build `reactor-c` as a library. Fortunately, this is probably not very hard -- we just have to parameterize the build process according to whether the execution is threaded, federated, federated and decentralized, etc. To my knowledge, it is an open question whether the library is destined to be used for production, or if it is only going to be used for testing.


core/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+    endif(NUMBER_OF_WORKERS)
+    foreach(FILE ${TEMP})
+        file(RELATIVE_PATH RELATIVE_TO_ROOT ${CORE_ROOT} ${CMAKE_CURRENT_SOURCE_DIR}/${FILE})
+        list(APPEND SOURCES ${RELATIVE_TO_ROOT})
+    endforeach()
+    set(SOURCES ${SOURCES} PARENT_SCOPE)
+endmacro()
+
+# Applies the X as a preprocessor definition if X is defined.
+macro(define X)
```
lhstrh, line 19: What is this for?

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include "core/utils/vector.h"
```
lhstrh, line 3: Maybe the directory of this test file should mirror the path of the code under test (i.e., `utils` instead of `data-structures`)?

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+ * 
+ * @param v A vector.
+ */
+void test_vote(vector_t* v) {
+    DEBUG_PRINT("vote.");
+    vector_vote(v);
+}
+
+/**
+ * @brief Runs a randomly selected test on `v`.
```
lhstrh, line 95:
```suggestion
 * @brief Run a randomly selected test on `v`.
```

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+            error_print_and_exit(
+                "Expected %p but got %p while randomly accessing a vector.",
+                mock[idx], v->start[idx]
+            );
+        }
+    }
+}
+
+/**
+ * @brief Checks that voting does not cause an error.
```
lhstrh, line 85:
```suggestion
 * @brief Check that voting does not cause an error.
```

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+    void** mock_start = mock + mock_size;
+    for (int i = 0; i < count; i++) {
+        mock[mock_size++] = mock - rand();
+    }
+    vector_pushall(v, mock_start, count);
+    return count;
+}
+
+/**
+ * @brief Checks that the result of a random access to `v` yields the
```
lhstrh, line 67:
```suggestion
 * @brief Check that the result of a random access to `v` yields the
```

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+    )) {
+        error_print_and_exit(
+            "Expected %p but got %p while popping from a vector.",
+            expected, found
+        );
+    }
+}
+
+/**
+ * @brief Tests the "push all" functionality of `v`.
```
lhstrh, line 50:
```suggestion
 * @brief Test the "push all" functionality of `v`.
```

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+ */
+void test_push(vector_t* v) {
+    DEBUG_PRINT("push.");
+    void* x = mock + rand();
+    vector_push(v, x);
+    mock[mock_size++] = x;
+}
+
+/**
+ * @brief Tests the pop functionality of `v`.
```
lhstrh, line 31:
```suggestion
 * @brief Test the pop functionality of `v`.
```

test/general/data-structures/vector_test.c
```diff
@@ -0,0 +1,139 @@
...
+#define N 5000
+#define RANDOM_SEED 1614
+
+static void* mock[CAPACITY];
+static size_t mock_size = 0;
+
+static int distribution[4] = {30, 50, 5, 15};
+
+/**
+ * @brief Tests the append functionality of `v`.
```
lhstrh, line 18:
```suggestion
 * @brief Test the append functionality of `v`.
```

test/Tests.cmake
```diff
@@ -0,0 +1,38 @@
...
+        ${DIR}/*${TEST_SUFFIX}
+    )
+    list(APPEND TEST_FILES ${TEST_FILES_FOR_DIR})
+    set(TEST_FILES ${TEST_FILES} PARENT_SCOPE)
+endfunction()
+
+# Add the appropriate directories for the provided build parameters.
+add_test_dir(${TEST_DIR}/general)
+if(NUMBER_OF_WORKERS)
+    add_test_dir(${TEST_DIR}/threaded)
```
lhstrh, line 23: I personally like `single-threaded` and `multi-threaded` more because `unthreaded` is a bit of a misnomer...

README.md
```diff
@@ -15,3 +15,17 @@ To generate and view documentation, follow the following steps:
   - `make html`
 - Point your browser to the generated HTML page:
   - `firefox _build/html/index.html`
+
+To create a new test, write a C program with a file name ending in "test.c"
+in the `test` directory.
+
+To run unthreaded tests, execute the following. Note that -U is required to undefine
```
lhstrh, line 22: There are certain tests that we'll want to have for both the multi-threaded and single-threaded runtime. Would we just have to replicate these tests? Or might there be a more elegant solution for this?

lhstrh (approved):
This looks great, @petervdonovan! Should this still be a draft? It looks to me like we could merge this... I left a few comments/questions.

petervdonovan:
> Should this still be a draft? It looks to me like we could merge this... 

Well, we seem to have a couple of options, in order of increasing complexity:
1. Merge in what we have now, so that it the build is configurable from the command line according to whether it is single-threaded or multi-threaded
2. Add more changes so that the build is configurable according to whether it is federated; add federated tests
3. Make the build fully configurable, so that it can replace a big chunk of the logic that currently exists in the code generator, including some of `CCmakeGenerator.xtend` as well as the parts that determine which files to include.

I lean toward addressing the comments and cleaning up one thing, and then making items 2 and 3 a separate PR.

lhstrh:
> > Should this still be a draft? It looks to me like we could merge this...
> 
> Well, we seem to have a couple of options, in order of increasing complexity:
> 
>     1. Merge in what we have now, so that it the build is configurable from the command line according to whether it is single-threaded or multi-threaded
> 
>     2. Add more changes so that the build is configurable according to whether it is federated; add federated tests
> 
>     3. Make the build fully configurable, so that it can replace a big chunk of the logic that currently exists in the code generator, including some of `CCmakeGenerator.xtend` as well as the parts that determine which files to include.
> 
> 
> I lean toward addressing the comments and cleaning up one thing, and then making items 2 and 3 a separate PR.

I agree with you! Please proceed as proposed :-)

test/Tests.cmake
petervdonovan: Now that you mention it, I see that you are right, at least in principle.

The code often includes the words "unthreaded," "non-threaded," and "threaded." Examples from grep:
```
core/reactor.h: * This file contains header information used by both the threaded and
core/reactor.h: * non-threaded versions of the C runtime. A generated C program will have
core/trace.h: * @param worker The thread number of the worker thread or 0 for unthreaded execution.
core/reactor.c:/* Runtime infrastructure for the non-threaded version of the C target of Lingua Franca. */
core/reactor.c:            tracepoint_reaction_starts(reaction, 0); // 0 indicates unthreaded.
core/reactor_threaded.c: * A struct representing a barrier in threaded 
core/reactor_common.c: * This is the internal implementation shared by both the threaded
core/reactor_common.c: * and non-threaded versions.
```
We could open an issue for this terminology. In principle, it seems like a problem, but there could be reasons for it. For example,
* The "unthreaded" runtime doesn't use a threading library.
* It must be distinguished from the multithreaded runtime in the case where `NUMBER_OF_WORKERS` equals 1.
* "unthreaded" and "threaded" have fewer syllables than "single-threaded" and "multithreaded."

README.md
petervdonovan: Currently, the tests under "general" are executed always, regardless of whether the runtime is single-threaded or multi-threaded. Does that suffice?

If tests become too expensive to run, then we could make an additional category for tests that can be run separately.

lhstrh: Not sure... There are differing implementations (reactor.c vs reactor_threaded.c) of various functions that should functionally behave the same and thus would require verbatim copies of certain tests, with the difference only being the imports. This seems like a situation where the Cmake potentially could help.

petervdonovan: I think that use case should be covered if those tests are placed in "general." In that case, you could run `cmake .. -UNUMBER_OF_WORKERS` to run the tests in "general", compiling and linking `.c` files from the single-threaded runtime, and you could run `cmake .. -DNUMBER_OF_WORKERS=2` (or something else nonzero) if you want to run the same tests, compiling and linking `.c` files from the multithreaded runtime.

It works on my machine, at least.

lhstrh: OK, yes, that sounds good. Thanks!

core/CMakeLists.txt
petervdonovan: The user passes arguments to CMake via the command line, like
```
cmake .. -DNUMBER_OF_WORKERS=4 -DFEDERATED=1
```

This macro passes those variables along to the compiler as preprocessor definitions. The changes I just added to the README will try to explain this.

test/general/data-structures/vector_test.c
petervdonovan: Sounds good -- the change has been applied.

test/Tests.cmake
petervdonovan: Well, this change has been applied -- no more misnomer. Since our use of the word "unthreaded" in other places is a separate issue, it seems okay to mark this one as resolved.

PR #27, opened 2022-01-04
Add integration tests

lhstrh:
This PR incorporates integration tests into our CI. A specific version of the compiler can be specified in a file called `lingua-franca-ref.txt`. If omitted, the default ref (for `lingua-franca` this is `master`) is used. For more info, see [documentation](https://github.com/lf-lang/lingua-franca/blob/master/.github/workflows/README.md).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  push:
+  pull_request:
+  # Also allow running this workflow manually from the Actions tab.
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
petervdonovan, line 12: For some reason, GitHub says they offer [more cores for MacOS virtual machines](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). It is unimportant and does not matter now, but we might revisit it later if we decide it would speed things up or help us detect parallelism bugs

petervdonovan (approved):

.github/workflows/ci.yml
lhstrh: More cores, yes, but there is more contention for MacOS runners (we get allocated fewer), so when there are a number of jobs running concurrently, progress tends to stall on our MacOS jobs. Also, Linux usually runs faster to begin with. My guess is that they're offering the extra cores to compensate for the poor performance...

PR #29, opened 2022-01-11
Update ci.yml

lhstrh:
CI gets a bit overwhelmed if we build for every push. Let's limit to PRs and pushes to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-  pull_request:
     branches:
       - master
+  pull_request:
   # Also allow running this workflow manually from the Actions tab.
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    branches:
+      - master
```
Soroosh129, line 6: Should this be `main`?

lhstrh: Ah, right. Nice catch!

lhstrh, line 6:
```suggestion
      - main
```

Soroosh129 (approved):

Soroosh129:
Looks good to me! Thanks for updating this.

PR #30, opened 2022-01-13
pass address, port, username of RTI in runtime

housengw:
allow passing hostname/IP, port and username of RTI as command line arguments to federates.

example:
`bin/federate_a -i 1 --host 127.0.0.1`


core/reactor_common.c
```diff
@@ -1726,48 +1772,67 @@ int process_args(int argc, char* argv[]) {
...
                 usage(argc, argv);
                 return 0;
             }
-            i++;
-            info_print("Federation ID for executable %s: %s", argv[0], argv[i]);
-            federation_id = argv[i++];
-        } else if (strcmp(argv[i], "--ros-args") == 0) {
+            federation_metadata.federation_id = argv[i++];
+            info_print("Federation ID for executable %s: %s", argv[0], federation_metadata.federation_id);
+        } else if (strcmp(arg, "--host") == 0) {
```
Soroosh129, line 1810: I think an argument like `--rti` or `-r` might be more descriptive. How hard do you think it would be to parse a full address in C? For example `--rti user@host:port`?

housengw: the primary reason I choose to separate --rti into --host --port and --user is to cut down the number of edge cases. Given we are parsing the address in C, doing --host --port and --user is a lot easier to get right.

Soroosh129: Right, I figured. Still, I think running a federate using a `--host` argument can be a bit misleading and hard to explain. Running a federate with a `--rti` argument I think is easier to understand and explain. It might be worth going the extra mile to parse a full host address.

I found a few promising ways to approach this problem just by googling around, including using [`sscanf`](https://stackoverflow.com/a/726214/783868) and/or using an [`isdigit()`](https://www.geeksforgeeks.org/isdigit-function-in-c-c-with-examples/) function that might be useful in `validate_port` as well.

If this is a dead-end, then I think `--rti --host xxx` is still a bit more descriptive that plain `--host`.

housengw:
I have replace the `--host`, `--port` and `--user` flags with `--rti`. I used `sscanf` to parse any string that looks like `a@b:c` into `user = a`, `host = b` and `port = c` and only performed validation on the port. While implementing the `--rti`, I noticed that it might be better to have a standard of parsing/validating rti address since we are doing it at 3 places (in the `.lf` file as `at`, compiler arguments, runtime arguments)

core/reactor_common.c
```diff
@@ -1655,47 +1658,156 @@ void usage(int argc, char* argv[]) {
...
+/**
+ * Extract the host, port and user from rti_addr.  
+ */
+void extract_rti_addr_info(char* rti_addr, rti_addr_info_t* rti_addr_info) {
+    int n = 0;
+    n = sscanf(rti_addr, "%255s@%255s:%5s", rti_addr_info->rti_userStr, rti_addr_info->rti_hostStr, rti_addr_info->rti_portStr);
+    if (n == 3) {
+        rti_addr_info->has_host = rti_addr_info->has_port = rti_addr_info->has_user = true;
+    } else {
+        n = sscanf(rti_addr, "%255s:%5s", rti_addr_info->rti_hostStr, rti_addr_info->rti_portStr);
```
Soroosh129, line 1727: I think the logic here can be improved a bit. First, `sscanf` can return `EOF` or `-1`, which are both error conditions that need to be dealt with. Second, it seems that a `user@ip` argument from the user can be problematic here (note that we do have a default port number). Would it be safe here to check if any of the fields (e.g., `rti_userStr`) have been set, and then try to re-parse the string accordingly?

core/reactor_common.c
```diff
@@ -1726,48 +1838,72 @@ int process_args(int argc, char* argv[]) {
...
             }
-            i++;
-            info_print("Federation ID for executable %s: %s", argv[0], argv[i]);
-            federation_id = argv[i++];
-        } else if (strcmp(argv[i], "--ros-args") == 0) {
+            federation_metadata.federation_id = argv[i++];
+            info_print("Federation ID for executable %s: %s", argv[0], federation_metadata.federation_id);
+        } else if (strcmp(arg, "--rti") == 0) {
+            if (argc < i + 1) {
+                error_print("--rti needs a string argument.");
```
Soroosh129, line 1878: Would be nice if this would explain the format of the string.

core/reactor_common.c
```diff
@@ -1726,48 +1838,72 @@ int process_args(int argc, char* argv[]) {
...
+                usage(argc, argv);
+                return 0;
+            }
+            parse_rti_code_t code = parse_rti_addr(argv[i++]);
+            if (code != SUCCESS) {
+                switch (code) {
+                    case INVALID_HOST:
```
Soroosh129, line 1885: It seems like `INVALID_HOST` and `INVALID_USER` are never returned. Perhaps `extract_rti_addr_info` is the best place to detect these two errors.

core/reactor_common.c
```diff
@@ -1655,47 +1658,156 @@ void usage(int argc, char* argv[]) {
...
+        }
+    }
+    int port_number = atoi(port);
+    return port_number >= 0 && port_number <= 65535;
+}
+
+/**
+ * A helper struct for passing rti_addr information betweem parse_rti_addr and extract_rti_addr_info
+ */
+typedef struct rti_addr_info_t {
```
Soroosh129, line 1698: These most likely belong in a separate file.

core/reactor_common.c
housengw: I tried to use sscanf and multiple passes but could not get it to work. Since we handle the rti's address using regex in compile time, I used regex to parse the rti's address instead. This would help standardize how we handle RTI's address at different places.

housengw:
seems like windows doesn't have regex... We can either revert to `sscanf` or try to use [PCRE](http://pcre.org), which implements a set of wrapper functions that correspond to the POSIX regular expression API

Soroosh129 (approved):
Looks great! Thanks for adding this great usability feature.

PR #31, opened 2022-01-13
Mixed radix support

edwardalee:
This branch adds functions for creating and manipulating permuted mixed-radix-numbers, which are used in C code generated by the scalability-banks branch of LF. Hence, this should be merged into main before scalability-banks is merged into master.


core/util.c
```diff
@@ -217,3 +217,48 @@ void register_print_function(print_message_function_t* function, int log_level)
     print_message_function = function;
     print_message_level = log_level;
 }
+
+/**
+ * Increment the mixed radix number by one according to the permutation matrix.
+ */
+void mixed_radix_incr(mixed_radix_int_t* mixed) {
+	int i = 0;
```
Soroosh129, line 225: Perhaps a check for `mixed` is warranted here and in the other function to make sure it is not null.

Soroosh129 (approved):
Looks good to me.

I have two relatively minor comments:
- Would it make sense to put these new mixed_radix-related stuff in a separate file, such as `core/utils/mixed_radix.h` and `core/utils/mixed_radix.h`?
- Would it make sense to add a function for creating a new `mixed_radix_int_t` and a function for subsequently freeing the memory used by it?

core/util.c
```diff
@@ -217,3 +217,48 @@ void register_print_function(print_message_function_t* function, int log_level)
...
+		mixed->digits[digit_to_increment]++;
+		if (mixed->digits[digit_to_increment] >= mixed->radixes[digit_to_increment]) {
+			mixed->digits[digit_to_increment] = 0;
+			i++;
+		} else {
+			return; // All done.
+		}
+	}
+	// If we get here, the number has overflowed. Wrap to zero.
+	mixed->digits[i - 1] = 0;
```
cmnrd, line 237: This is potentially dangerous. If `mixed->size <= 0`, then this will write `0` to some memory location.

core/util.c
```diff
@@ -217,3 +217,48 @@ void register_print_function(print_message_function_t* function, int log_level)
...
     print_message_level = log_level;
 }
+
+/**
+ * Increment the mixed radix number by one according to the permutation matrix.
+ */
+void mixed_radix_incr(mixed_radix_int_t* mixed) {
+	int i = 0;
+	while (i < mixed->size) {
+		int digit_to_increment = mixed->permutation[i];
```
cmnrd, line 227: Also this seems dangerous. What if permutation holds an invalid index? It is used below unchecked.

cmnrd (commented):
I found and commented on a few unguarded array accesses, where there can easily be a buffer over or underflow.

Also, I agree with Soroush that such a "class" should better be defined in its own file. I think a function for allocating and initializing a radix number would be very helpful, similar to a constructor in oo languages.

core/util.c
edwardalee: What do you suggest doing in such error conditions? Note that C does not have exceptions. This code is meant to be invoked only generated code, but it does end up in scope for people writing reactions, so there could be a security issue.  But what to do?  Silently do nothing? Print an error message and exit? Note that in some embedded applications, printing something doesn't make sense.

edwardalee:
> * Would it make sense to add a function for creating a new `mixed_radix_int_t` and a function for subsequently freeing the memory used by it?

I don't have a use case for this. The code generator puts the mixed radix struct on the stack. In fact, it puts all temporary data structures on the stack. It only allocates memory for data structures that have to persist for the lifetime of reactors. This ensures that the heap is pristine (no fragmentation) upon the start of execution.

core/util.c
cmnrd: I think the easiest way would be to use `assert`. This way we could at least verify all pre-conditions for debug builds (Assertions are disabled in release builds). If the assertion does not hold, the program would abort.

Typically, C functions use a return value to indicate errors. For instance, the function could return 1 on success and 0 on failure. This would leave it up to the user to handle the error.

PR #34, opened 2022-01-20
New scheduler unit tests

Soroosh129:
This implements a simple unit test for the scheduler API and further untangles parts of the c runtime (mainly tag.c).


lhstrh:
It would be really nifty if we could also get code coverage reports out of this. Here's a useful pointer: https://github.com/codecov/example-cpp11-cmake

core/tag.c
```diff
@@ -73,6 +76,44 @@ interval_t _lf_global_physical_clock_drift = 0LL;
...
+ * the current time, then increase the microstep. Otherwise, update the current
+ * time and set the microstep to zero.
+ * 
+ * @param next_time The time step to advance to.
+ */ 
+void advance_tag(instant_t next_time) {
+    // FIXME: The following checks that advance_tag()
+    // is being called correctly. Namely, check if logical time
+    // is being pushed past the head of the event queue. This should
+    // never happen if _lf_advance_logical_time() is called correctly.
```
edwardalee, line 91:
```suggestion
    // never happen if advance_tag() is called correctly.
```

core/tag.c
```diff
@@ -178,6 +219,19 @@ instant_t _lf_last_reported_physical_time_ns = 0LL;
...
+ * Return the most recent time reported by the physical clock
+ * when accessed by get_physical_time(). This will be an epoch time
+ * (number of nanoseconds since Jan. 1, 1970), as reported when
+ * you call lf_clock_gettime(CLOCK_REALTIME, ...). This differs from
+ * _lf_last_reported_physical_time_ns by _lf_global_physical_clock_offset
+ * plus any calculated drift adjustement, which are adjustments made
+ * by clock synchronization.
+ */
+instant_t get_last_reported_unadjusted_physical_time() {
+	return _lf_last_reported_unadjusted_physical_time_ns;
```
edwardalee, line 232:
```suggestion
    return _lf_last_reported_unadjusted_physical_time_ns;
```

core/tag.h
```diff
@@ -193,6 +202,17 @@ extern interval_t _lf_global_test_physical_clock_offset;
  */
 extern interval_t _lf_epoch_offset;
 
+/**
+ * Return the most recent time reported by the physical clock
+ * when accessed by get_physical_time(). This will be an epoch time
+ * (number of nanoseconds since Jan. 1, 1970), as reported when
+ * you call lf_clock_gettime(CLOCK_REALTIME, ...). This differs from
+ * _lf_last_reported_physical_time_ns by _lf_global_physical_clock_offset
```
edwardalee, line 210: Are the function names cited here still correct?

edwardalee (approved):
This PR realizes some much-needed changes to reduce the inclusion of C files and instead separately compile the C files.  This should be merged.

lhstrh:
@petervdonovan is there anything in this PR that is still useful, or can this be closed, you think?

petervdonovan:
You mentioned this PR to me multiple times and I never set aside the time to finish it. Partly this is because I was never excited about the idea of unit testing schedulers -- all non-trivial bugs in them are concurrency bugs, and the best way to catch concurrency bugs is integration tests; and furthermore, the schedulers are not deep, algorithmic modules, and so most of their complexity is in their integration with the whole system, rather than being intrinsic to them. It looks like this PR includes valuable refactoring, but with so many conflicts I doubt that reviving it would be faster than starting from scratch

lhstrh:
That makes sense. Maybe @erlingrj or @edwardalee are interested in picking up the thread, but if not, let's close this.

erlingrj:
At the moment I am not interested in picking up this PR

PR #38, opened 2022-02-11
Add check_deadline function.

hokeun:
See https://github.com/lf-lang/lingua-franca/issues/403 for more information.

### DIFF
--- a/core/reactor.h
+++ b/core/reactor.h
@@ -570,6 +570,7 @@ typedef struct allocation_record_t {
  */
 typedef struct self_base_t {
 	struct allocation_record_t *allocations;
+    struct reaction_t *executing_reaction;   // The currently executing reaction of the reactor.
 } self_base_t;
 
 //  ======== Function Declarations ========  //
--- a/core/util.c
+++ b/core/util.c
@@ -218,3 +218,20 @@ void register_print_function(print_message_function_t* function, int log_level) 
     print_message_level = log_level;
 }
 
+/**
+ * Check the deadline of the currently executing reaction against the
+ * current physical time. If the deadline has passed, invoke the deadline
+ * handler and return true. Otherwise, return false.
+ * 
+ * @param self The self struct of the reactor.
+ * @return true if the specified deadline has passed.
+ * @return false if the deadline has not passed yet.
+ */
+bool check_deadline(void* self) {
+	reaction_t* reaction = ((self_base_t*)self)->executing_reaction;
+	if (get_physical_time() > get_logical_time() + reaction->deadline) {
+		reaction->deadline_violation_handler(self);
+		return true;
+	}
+	return false;
+}


hokeun:
> Another thing is that user-facing API should be defined/prototyped in `ctarget.h`.

Thanks for the comments, Soroush. I hope I got your comments correctly in https://github.com/lf-lang/reactor-c/pull/38/commits/c97abe58a6c68dfc6c5649298c16b8e53a409d1d.

core/reactor.h
```diff
@@ -876,6 +877,17 @@ void _lf_notify_workers(void);
  */
 bool _lf_is_blocked_by_executing_reaction(void);
 
+/**
+ * Check the deadline of the currently executing reaction against the
+ * current physical time. If the deadline has passed, invoke the deadline
+ * handler and return true. Otherwise, return false.
+ * 
+ * @param self The self struct of the reactor.
+ * @return true if the specified deadline has passed.
```
edwardalee, line 886: I think usually there is only one @return statement, which would read like this:
```
@return True if the specified deadline has passed and false otherwise.

File: core/reactor_common.c [hunk 2]
```diff
@@ -1534,6 +1534,24 @@ lf_token_t* _lf_set_new_array_impl(lf_token_t* token, size_t length, int num_des
     return new_token;
 }
 
+/**
+ * Check the deadline of the currently executing reaction against the
+ * current physical time. If the deadline has passed, invoke the deadline
+ * handler and return true. Otherwise, return false.
+ * 
+ * @param self The self struct of the reactor.
+ * @return true if the specified deadline has passed.
```
edwardalee, line 1543: Same comment.

lib/ctarget.c
```diff
@@ -178,3 +178,16 @@ trigger_handle_t schedule_value(void* action, interval_t extra_delay, void* valu
...
     return _lf_schedule_value(action, extra_delay, value, (size_t)length);
 }
+
+/**
+ * Check the deadline of the currently executing reaction against the
+ * current physical time. If the deadline has passed, invoke the deadline
+ * handler and return true. Otherwise, return false.
+ * 
+ * @param self The self struct of the reactor.
+ * @return true if the specified deadline has passed.
```
edwardalee, line 188: Same comment.

include/ctarget.h
```diff
@@ -252,4 +252,15 @@ trigger_handle_t schedule_copy(void* action, interval_t offset, void* value, int
  */
 trigger_handle_t schedule_value(void* action, interval_t extra_delay, void* value, int length);
 
+/**
+ * Check the deadline of the currently executing reaction against the
+ * current physical time. If the deadline has passed, invoke the deadline
+ * handler and return true. Otherwise, return false.
+ * 
+ * @param self The self struct of the reactor.
+ * @return true if the specified deadline has passed.
```
edwardalee, line 261: Same comment.

edwardalee (approved):
Very elegant and simple! Let's merge this.

lhstrh (approved):
Very nice!

core/reactor_common.c
```diff
@@ -1534,6 +1534,24 @@ lf_token_t* _lf_set_new_array_impl(lf_token_t* token, size_t length, int num_des
...
+ * @return false if the deadline has not passed yet.
+ */
+bool _lf_check_deadline(self_base_t* self) {
+	reaction_t* reaction = self->executing_reaction;
+	if (get_physical_time() > get_logical_time() + reaction->deadline) {
+		reaction->deadline_violation_handler(self);
```
Soroosh129, line 1549: @edwardalee The deadline violation handler function is called here. Is this semantically okay?

Because of the way `check_deadline()` is supposed to work, the deadline handler could be called multiple times (if the caller doesn't terminate the reaction (e.g., by using `return`), upon `check_deadline()` returning true, and calls `check_deadline()` again later). What if the deadline handler had some side effect, including changing a state variable and setting to an output port?
This semantic seems a bit strange to me. Depending on the body of the reaction, the deadline violation handler could be called multiple times within the same tag. Even stranger, the number of times the deadline violation handler is called depends on many external factors, like the load on the system.

Soroosh129: Okay, got it.

I think the way the API is documented on the website might be important to make sure programmers fully understand the ramifications of using `check_deadline`.

hokeun: I just merged the PR as I believe we reached an agreement. Please let me know if there's any further suggestion, I'll be happy to address that.

PR #45, opened 2022-02-22
Priority driven schedulers

petervdonovan:
This is here to track progress implementing priority-driven schedulers other than (our version of) EDF.

As usual for a draft PR, it is not ready for review.

As we have discussed, our semantics may not fully support typical scheduling strategies. In particular, I am not sure what DM means and how it is different from EDF in the context of our mixed logical and physical timelines since our deadlines are given in logical time. We also do not have a way to specify execution times in cases where they are known.


petervdonovan:
I am closing this because it only provides a toy implementation and because it does not solve a real problem that we are facing right now.

PR #49, opened 2022-03-02
Hotfix: Added back tracepoints for scheduler-related events

Soroosh129:


Soroosh129:
> Does the visualization of a trace now still show parallel execution? It seems that in several places the `worker` argument has been replaced with `-1`.

Only for `scheduler_advancing_time_starts` and `scheduler_advancing_time_ends` events which will not be parallel anyway.

PR #50, opened 2022-03-09
Workers

Soroosh129:
See lf-lang/lingua-franca#993.


core/threaded/reactor_threaded.c
```diff
@@ -1137,8 +1137,23 @@ int lf_reactor_c_main(int argc, char* argv[]) {
 
     if (process_args(default_argc, default_argv)
             && process_args(argc, argv)) {
-        lf_mutex_lock(&mutex); // Sets start_time
-        initialize();
+        
+        // If _lf_number_of_threads is 0, it means that it was not provided on
+        // the command-line using the --threads argument.
```
edwardalee, line 1142: Is this comment correct? First, should the command-line argument name be changed to match the target property name? Second, what if this is specified as a target property rather than a command-line argument?

edwardalee (approved):
This looks like a really good solution to me.

core/platform.h
```diff
@@ -90,6 +90,11 @@ typedef _microstep_t microstep_t;
 
 #ifdef NUMBER_OF_WORKERS
 
+/**
+ * @brief Get the number of cores on the host machine.
+ */
+extern int lf_host_cores();
```
lhstrh, line 96: Maybe a more descriptive name would be `num_of_cores` or `available_cores`? I don't think "host" is very meaningful; this being part of the platform API makes it obvious that this pertains to the platform.

core/platform/lf_unix_syscall_support.c
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+#include <unistd.h>
+
+/**
+ * @brief Get the number of cores on the host machine.
+ */
+int lf_host_cores() {
+    return (int)sysconf(_SC_NPROCESSORS_ONLN);
+}
```
lhstrh, line 19: Missing newline.

core/threaded/reactor_threaded.c
```diff
@@ -1137,8 +1137,23 @@ int lf_reactor_c_main(int argc, char* argv[]) {
 
     if (process_args(default_argc, default_argv)
             && process_args(argc, argv)) {
-        lf_mutex_lock(&mutex); // Sets start_time
-        initialize();
+        
+        // If _lf_number_of_threads is 0, it means that it was not provided on
```
lhstrh, line 1141: Maybe better to put this in a separate function (and call it inside of `initialize()`).

lhstrh (approved):
Looks great! Thanks for turning this around so quickly!

core/threaded/reactor_threaded.c
lhstrh: It thought that if the target property is specified, this results in the `#define` to be set in the generated code. We have to make sure that the CLI arg still overrides it, but it seems like this would indeed be that case...

Soroosh129: > First, should the command-line argument name be changed to match the target property name?

Ah, yes, thank you. It should be workers. I'll make that change.

> Second, what if this is specified as a target property rather than a command-line argument

The command-line argument takes priority. This made the most sense to me at least.

lhstrh: There is a caveat to that approach: if there is a minimum number of workers required to accommodate blocking input control reactions, we don't want to go below that number. I think we need to encode that requirement separately.

core/threaded/reactor_threaded.c
Soroosh129: `initialize()` is in reactor_common.c, which is also used in the unthreaded runtime.

lhstrh: Ah, you're right; please ignore the latter part of my suggestion.

core/platform.h
Soroosh129: We don't have fully qualified names in C, so you would just see `available_cores()` in the code with no context where it is defined, unless you look for it. I think `lf_host_cores` is descriptive enough (maybe `lf_host_num_cores`?), but if you think `available_cores` is better, we could use that too.

lhstrh: What else would the number of reported available cores refer to but the number of available core on which the code executes? The number of available cores on a different machine? Who would assume that?

core/threaded/reactor_threaded.c
Soroosh129: Yes, good point. I applied your suggestion.

core/platform.h
Soroosh129: Sure, we can use `available_cores()`. That's fine with me.

lhstrh: With the `lf_` prefix though, right?

core/platform/lf_unix_syscall_support.c
lhstrh: Newline is still missing

lhstrh (approved):
🚀🚀🚀

core/platform/lf_unix_syscall_support.c
Soroosh129: Strange, I thought I added it though.

lhstrh (approved):

PR #51, opened 2022-03-11
Continuous benchmarking

LDeng0205:
This PR adds the [continuous benchmarking](https://github.com/benchmark-action/github-action-benchmark). It currently runs with Github hosted runners; we are in the process of setting up russell as self-hosted runner to run the benchmarks. This PR comes with another PR in the benchmarks repo that makes some modifications to the runner script.


petervdonovan:
There might be a little bit of meaningless noise in this PR while I try to setup GitHub Actions. There is an "unsubscribe" button in Github (on the right) for anyone who does not want useless messages.

petervdonovan:
The only thing holding this PR back right now is that the performance alerts are not appearing anymore, and I cannot figure out why not.

lhstrh:
It seems we're incredibly close with this one. Is this ready to merge yet?

petervdonovan:
I think it might actually be fine to merge this. The other PR has to get moved, this seems like it can just get merged here. I still need to set up the self-hosted runner, but it seems imprudent to wait for the backlog of tasks to clear.

petervdonovan:
I keep being reminded of this PR, which is still potentially useful and which is still on the brink of being completed. After letting it sit for a while, I am beginning to get the feeling that hooking benchmarks into CI just might not be the way to go. CI has been a constant time-drain, and it will additionally become a security hazard if/when we link it up with our own servers. We may eventually decide that the most time-saving option in the long run is to focus on making sophisticated tests and benchmarks easy to run manually, on a server that we can just manually ssh into when necessary.

lhstrh:
If we run benchmarks for feature branches on request and regularly (during times that CI is less busy) on `main`, I think the impact on wait times and security will be minimal. So I guess that makes it less "continuous" but that's OK because not all changes are expected to impact performance.

.github/workflows/continuous-benchmark.yml
```diff
@@ -0,0 +1,96 @@
...
+
+permissions:
+  contents: write
+  deployments: write
+
+
+jobs:
+  benchmark:
+    name: Run C Benchmarks
+    runs-on: ubuntu-latest #FIXME: change to self-hosted after russel is set up.
```
lhstrh, line 17:
```suggestion
    runs-on: Linux
```

PR #54, opened 2022-03-18
Factored out support for modal models into a separate file

### LINKED ISSUE #48: Factor out support for modal models into a separate file
I think that some things in `reactor_common.c` would be better split off into a separate `modal.c`. I'm OK with merging (the pull request) now and doing a refactoring later.

_Originally posted by @lhstrh in https://github.com/lf-lang/reactor-c/pull/6#pullrequestreview-896947812_

a-sr:
Closes #48
Relates to [PR #1054](https://github.com/lf-lang/lingua-franca/pull/1054)


Soroosh129:
This looks great!
Would you please also take a look at reactor.h to see if anything related to modal models could also be factored out into a  modal.h there?

In terms of organization of the `core` directory, I think it might be better if both `modal.c` and the potential `modal.h` were in a `core/modal_models` folder. For example, the Python target has a [`modal_models/definitions.h`](https://github.com/lf-lang/reactor-c-py/blob/main/include/modal_models/definitions.h) and a [`modal_models/impl.c`](https://github.com/lf-lang/reactor-c-py/blob/main/lib/modal_models/impl.c).

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+void _lf_trigger_reaction(reaction_t* reaction, int worker_number);
+event_t* _lf_create_dummy_events(trigger_t* trigger, instant_t time, event_t* next, microstep_t offset);
+extern reaction_t** _lf_startup_reactions;
+extern int _lf_startup_reactions_size;
+extern pqueue_t* event_q;
+
+/**
+ * Checks whether the given mode is currently considered active.
+ * This includes checking all enclosing modes.
+ * If any of those is inactive, then so is this one.
```
edwardalee, line 52: This doc is confusing. It implies that every enclosed mode within an active mode is active. But shouldn't only one of those be active?

Also, I would change to use imperative case. For example, the doc could be:
```
 * Return true if the given mode is active.
 * A given mode is active if it is marked active and all its enclosing modes are active.
```

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+                state = NULL;
+            }
+        }
+        //DEBUG_PRINT(" => Mode is active");
+    }
+    return true;
+}
+
+/**
+ * Handle startup reactions in modal reactors.
```
edwardalee, line 83:
```suggestion
 * Trigger startup reactions in modal reactors that should be triggered.
 * A startup reaction should be triggered if it is in an active mode and
 * either it has never been triggered before or it has not been triggered
 * since a reset transition was taken into the mode.
```

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+
+    new_suspended_event->event = event;
+    new_suspended_event->next = _lf_suspended_events_head; // prepend
+    _lf_suspended_events_num++;
+
+    _lf_suspended_events_head = new_suspended_event;
+}
+
+/**
+ * Removes the given node from the list of suspended events.
```
edwardalee, line 139:
```suggestion
 * Remove the given node from the list of suspended events and
```

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+        if (predecessor != NULL) {
+                predecessor->next = next; // Remove from linked list
+        }
+    }
+
+    return next;
+}
+
+/**
+ * Performs transitions in all modal reactors.
```
edwardalee, line 174:
```suggestion
 * Perform transitions in all modal reactors.
```
A bit more of an explanation of what this means would be welcome.

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+ *
+ * @param state An array of mode state of modal reactor instance Must be ordered hierarchically.
```
edwardalee, line 176:
```suggestion
 * @param state An array of mode state of modal reactor instance, which must be ordered hierarchically, where
```

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+ *              Enclosing mode must come before inner.
```
edwardalee, line 177:
```suggestion
 *      an enclosing mode must come before the inner mode.
```

core/modal.c
```diff
@@ -0,0 +1,371 @@
...
+        if (should_trigger_startup_reactions_at_next_microstep) {
+            // Insert a dummy event in the event queue for the next microstep to make
+            // sure startup reactions (if any) can be triggered as soon as possible.
+            pqueue_insert(event_q, _lf_create_dummy_events(NULL, current_tag.time, NULL, 1));
+        }
+    }
+}
+
+/**
+ * Releases internal data structures for modes.
```
edwardalee, line 348:
```suggestion
 * Release internal data structures for modes.
```

edwardalee (approved):
This is an excellent refactoring.  I've made some suggestions for the documentation, although these predate the refactoring, so feel free to ignore them for this PR.

core/modal.c
```diff
@@ -0,0 +1,374 @@
...
+
+// Forward declaration of functions and variables supplied by reactor_common.c
+void _lf_trigger_reaction(reaction_t* reaction, int worker_number);
+event_t* _lf_create_dummy_events(trigger_t* trigger, instant_t time, event_t* next, microstep_t offset);
+extern reaction_t** _lf_startup_reactions;
+extern int _lf_startup_reactions_size;
+extern pqueue_t* event_q;
+
+/**
+ * Checks whether the given mode is currently considered active.
```
lhstrh, line 50:
```suggestion
 * Return true if the given mode is active.
```

core/modal.c
lhstrh: I think the comment is correct, since it is talking about enclosing modes (up in the hierarchy), not enclosed modes (down in the hierarchy).

lhstrh (approved):
Much better!

a-sr:
Now, there is also a header for modes that factors out additions to `reactor.h` and both modal models files are in a separated directory now.

Soroosh129 (approved):
LGTM!

As I mentioned in the weekly meeting today, we should try to avoid including .c files directly if possible but it's not a critical issue. There are other .c files that we include as well, so we could make a concerted effort at a later time to fix this.

PR #58, opened 2022-03-28
Added const to function signatures to suppress warnings.

edwardalee:
Picky compilers issue a warning when a function `foo(char* s)` is called with a constant string, `foo("constant string");`
This PR adds `const` to a few function signatures to prevent this warning.


lhstrh (approved):

Soroosh129 (approved):
LGTM!

Issue #44, opened 2022-02-21
Enhanced tokens for more flexible memory management


edwardalee:
In our conversation with @rcakella today, it became clear that we need to enhance the token mechanism in the C infrastructure with a token that carries its own destructor. Specifically, you can currently use some of the SET functions, for example SET_NEW, to dynamically allocate memory for an outgoing message, and then a reference counting mechanism is used to later free this memory when it is no longer used. But this will not work if your message is, for example, a struct that contains pointers to dynamically allocated memory. Only the struct will be freed, not the pointed to memory. A simple solution, I think, would be to augment the token_t struct with a destructor, a function pointer. If this is NULL, then free() will be used, as now. If it is non-null, then the specified function will be invoked instead. We will then provide another variant of SET that takes the destructor as an argument.

housengw:
passing the destructor to `SET_NEW` might be tricky in the remote federated case since the `token` at the input port of the receiving federate is allocated using a network control reaction under the hood (in `federate.c`). 

I think a cleaner solution might be to make the destructor part of the struct instead of the token.

edwardalee:
I had not thought about using the token destructor across federates. In fact, I don't think it makes sense to use across federates. I think we should rely on the serialization/deserialization scheme to be able to work at the receiving end with the default destructor (which just calls free()).  The sending end could use the provided destructor after serializing the data.

housengw:
closed by #66

PR #60, opened 2022-04-02
Add checks for `token->ok_to_free` before freeing the token's value

housengw:
Soroush and I noticed that `token->ok_to_free` field is not checked before freeing the token's value (currently, token's value is always freed).

We ran into this issue because we used `shared_ptr` as the type of an action. Shared pointers are freed automatically, but our runtime also tries to free it at the start of the next time step. This creates a double free issue and causes a segfault. 

For now, we manually set `token->ok_to_free` to be `token_only` for the shared pointer physical action.
In the future, we could pattern match the type of actions and set `token->ok_to_free` to `token_only` automatically in `initialize_trigger_objects`.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -378,9 +378,9 @@ token_freed _lf_done_using(lf_token_t* token) {
     if (token->ref_count == 0) {
         if (token->value != NULL) {
             // Count frees to issue a warning if this is never freed.
-            // Do not free the value field if it is garbage collected.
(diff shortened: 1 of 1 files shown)


edwardalee (approved):
This looks OK to me, but there is a more general solution that we need to implement, see issue #44 . Interested in taking that on?

housengw:
I took a look at issue #44 and the design outlined in the issue looks promising. I can take this after the EMSOFT deadline. I will merge this temporary fix for now.

PR #61, opened 2022-04-04
Remove TAN messages and record in-transit messages in the RTI

Soroosh129:
This PR removes TAN messages entirely.

Instead, a federate with a physical action (that is connected to a network output) is going to periodically create a dummy event (with the period controlled by `coordination-options: {advance-message-interval: 10 msec}`) which forces the federate to advance its tag and allow downstream federates to make progress.

After fixing this bug, another bug was exposed in the RTI, in which the RTI could potentially lose track of a federate's actual earliest next event (see [this](https://github.com/lf-lang/lingua-franca/pull/1074#issuecomment-1141282021) comment for more detail). This caused the RTI to grant incorrect tag advance grant (TAG) messages. This bug was fixed by adding a queue to the RTI that keeps a record of all currently in-transit messages.


edwardalee (approved):
LGTM.

Soroosh129:
Actually, I realized that this fix is a bit of a red herring.

The added second check for `_lf_bounded_NET(&tag)` is checking a modified version of `tag`, where `tag.time = get_physical_time()` at some point in the near past, against the current physical time. So it always returns false.

This is effectively disabling TAN, and that "fixed" the issue. However, TAN messages are important to ensure that progress is made.

Background of the problem: TAN messages appear to cause incorrect Tag Advance Grants to be sent by the RTI, causing STP violations in centralized coordination. I'm not sure why. It's very clear they are the issue because commenting out the content of `handle_time_advance_notice` causes the STP violations to go away.

core/federated/RTI/rti.c
```diff
@@ -826,8 +826,10 @@ void handle_time_advance_notice(federate_t* fed) {
     // Check downstream federates to see whether they should now be granted a TAG.
     // To handle cycles, need to create a boolean array to keep
     // track of which upstream federates have been visited.
-    bool* visited = (bool*)calloc(_RTI.number_of_federates, sizeof(bool)); // Initializes to 0.
-    send_downstream_advance_grants_if_safe(fed, visited);
+    for (int i = 0; i < fed->num_downstream; i++) {
```
edwardalee, line 829: I think this change cannot possibly solve the problem we were seeing and contradicts the documentation for the `send_advance_grant_if_safe` function, which says:
```
 * This should be called whenever an immediately upstream federate sends to
 * the RTI an LTC (Logical Tag Complete), or when a transitive upstream
 * federate sends a TAN (Time Advance Notice) or a NET (Next Event Tag) message.
 * It is also called when an upstream federate resigns from the federation.
```
This change will result in the transitive part of this not being done upon receiving a TAN.

edwardalee (changes requested):
The cosmetic changes look good to me, but the one substantive change is probably not right. Before the change, a TAN would result in calling `send_downstream_advance_grants_if_safe` for all downstream federates, and after the change only for immediately downstream federates.  But at worst, it should be harmless to call it for all downstream federates because `send_downstream_advance_grants_if_safe` calls `send_advance_grant_if_safe`, which checks for each federate whether it is actually safe to send a TAG.  Since the docs for that latter function say clearly it should be called on _all_ downstream federates, I suspect there was a reason for that. I suggest reverting this change and merging in the cosmetic changes.

core/federated/RTI/rti.c
Soroosh129: Yes I agree. This change did not fix the issue anyway. I will revert this change. Please note that lf-lang/lingua-franca#1074 depends on this branch in reactor-c.

edwardalee:
The RTI, as defined in this branch, causes the following tests in federated to lock up and time out: LoopDistributedDouble.lf, PingPongDistributed.lf, LoopDistributedCentralized.lf.  Looks like TANs are not being sent when they should be.

core/federated/RTI/message_record/message_record.c
```diff
@@ -0,0 +1,170 @@
...
+#include "message_record.h"
+#include <stdlib.h>
+
+/**
+ * @brief Initialize the in-transit message record queue.
+ * 
+ * @return in_transit_message_record_q 
+ */
+in_transit_message_record_q_t initialize_in_transit_message_q() {
+    in_transit_message_record_q_t queue =
```
edwardalee, line 40: I don't think this should be a calloc.  You can just declare an in_transit_message_record_q_t on the stack and then return it. The returned value will be a copy of the stack variable, which is a pair of pointers.  In any case, this looks wrong because calloc always returns a pointer and here you cast it to a non-pointer, which will likely result in overwriting 64 bits of stack data.

core/federated/RTI/message_record/message_record.c
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * @brief Free the memory occupied by the `queue`.
+ * 
+ * @param queue The queue to free.
+ */
+void free_in_transit_message_q(in_transit_message_record_q_t queue) {
+    pqueue_free(queue->main_queue);
+    pqueue_free(queue->transfer_queue);
+    free(queue);
```
edwardalee, line 76: I think this free() is wrong because the argument is not a pointer. This will result in freeing random memory, or more likely in an error of some sort, assuming free() performs some checks.

edwardalee (changes requested):
LGTM, except for one potential error raised in the code.  Also, the PR is misnamed because it does much more than remove TAN messages. Perhaps "Remove TAN messages and record in-transit messages in the RTI"? I did a double take on in_transit_message_record_q_t not being a pointer, but then I realized it is a pair of pointers, so this seems reasonable to me.

core/federated/RTI/message_record/message_record.c
Soroosh129: `in_transit_message_record_q_t` is a pointer:
```C
typedef struct in_transit_message_record_q* in_transit_message_record_q_t;
```

core/federated/RTI/message_record/message_record.c
Soroosh129: Should I perhaps rename `in_transit_message_record_q_t` to `in_transit_message_record_q_t_p` to make the fact that it is a pointer clearer? Or maybe we shouldn't hide pointers at all. This was just an arbitrary stylistic choice by me to abstract away the fact that `in_transit_message_record_q_t` is a pointer because the code outside message_record shouldn't care.

core/federated/RTI/rti.h
```diff
@@ -76,7 +77,11 @@ typedef struct federate_t {
     tag_t last_granted;     // The maximum TAG that has been granted so far (or NEVER if none granted)
     tag_t last_provisionally_granted;      // The maximum PTAG that has been provisionally granted (or NEVER if none granted)
     tag_t next_event;       // Most recent NET received from the federate (or NEVER if none received).
-    instant_t time_advance; // Most recent TAN received from the federate (or NEVER if none received).
+    in_transit_message_record_q_t in_transit_message_tags;  // Record of in-transit messages to this federate that are not 
+                                                            // yet processed. Messages are assumed to be processed once an
+                                                            // LTC arrives that has the same or larger tag. Since messages
+                                                            // are assumed to be delivered in-order, there is no need to
+                                                            // sort this.
```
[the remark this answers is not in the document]
Soroosh129: Ah, nice catch! This was an old comment referring to a vector. I replaced it with pqueue but forgot to update this.

core/federated/RTI/message_record/message_record.c
Soroosh129: @edwardalee did my comments address your concern? This PR is still pending an approval.

core/federated/RTI/message_record/message_record.c
edwardalee: I think it would be clearer to eliminate the following:
```
typedef struct in_transit_message_record_q* in_transit_message_record_q_t;
```
Instead of `in_transit_message_record_q_t`, just use `in_transit_message_record_q*`.

Soroosh129: Okay, done here: f05be58b838141ab2398b1f855fd8dd165a42e89

edwardalee (approved):

core/federated/RTI/message_record/message_record.c
```diff
@@ -0,0 +1,170 @@
...
+ * @param queue The queue to search in (of type `in_transit_message_record_q`).
+ * @return tag_t The minimum tag of all currently recorded in-transit messages. Return `FOREVER_TAG` if the queue is empty.
+ */
+tag_t get_minimum_in_transit_message_tag(in_transit_message_record_q_t queue) {
+    tag_t minimum_tag = FOREVER_TAG;
+
+    in_transit_message_record_t* head_of_in_transit_messages = (in_transit_message_record_t*)pqueue_peek(queue->main_queue);
+    while (head_of_in_transit_messages != NULL) { // Queue is not empty
+        // The message record queue is ordered according to the `time` field, so we need to check
+        // all records with the minimum `time` and find those that have the smallest tag.
```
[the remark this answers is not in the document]
Soroosh129: This has been a long-running interesting conversation. Would you like to create an issue or a discussion for this?

PR #63, opened 2022-04-09
Dereferencing null ptr

### LINKED ISSUE #62: Dereferencing null pointer
In line 540 of `reactor_common.c`, there is a check for `null` token `if (token == NULL || ...)`

If this condition is true, however, line 542 would de-reference the `element_size` field of the `null` token to create a new token. 
https://github.com/lf-lang/reactor-c/blob/ab54ee59662c5ed20831e11b607df0a7b53ecd05/core/reactor_common.c#L540-L542

edwardalee:
This fixes #62

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -531,13 +531,13 @@ lf_token_t* create_token(size_t element_size) {
  *  field pointing to newly allocated memory.
  */
 lf_token_t* _lf_initialize_token_with_value(lf_token_t* token, void* value, size_t length) {
-    // assert(token != NULL);
+    assert(token != NULL);
 
     // If necessary, allocate memory for a new lf_token_t struct.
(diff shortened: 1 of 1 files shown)


housengw (commented):
LGTM

core/reactor_common.c
```diff
@@ -531,13 +531,13 @@ lf_token_t* create_token(size_t element_size) {
  *  field pointing to newly allocated memory.
  */
 lf_token_t* _lf_initialize_token_with_value(lf_token_t* token, void* value, size_t length) {
-    // assert(token != NULL);
+    assert(token != NULL);
```
Soroosh129, line 534: I think the default CMake build type ("Release") disables asserts. Would that be okay here?

edwardalee: I think this is fine. I checked all the callers, and none has the possibility passing a null argument.

Issue #65, opened 2022-04-12
Simplifying the `SET` API and its variants


housengw:
While working on [issue 44](https://github.com/lf-lang/reactor-c/issues/44), I realized it might help with ease of use if we provide an alternate set of `SET` APIs that are defined based on the memory properties of the set data. 

Instead of adding `destructor` variants to existing `SET` APIs, I propose a design that will only have two `SET` variants, one for static data and one for dynamic data:
1. `SET(out, val)`    (implies static data)
2. `SET_DYNAMIC(out, val, len, destructor)`   (implies dynamic data)

Implementation:
`SET(out, val)` is the same as the current `SET`.
`SET_DYNAMIC(out, val, len, destructor)` is the same as `SET_ARRAY` except without the `elem_size` argument (which `SET_ARRAY` doesn't use in its macro body, anyway).

Example use case:
Suppose is a struct called `data`:
1.  if `data` is static (none of its members is malloc'd), just use `SET(out, data)`
2. if `data` has dynamic members, malloc the whole struct, then malloc the members. Then, use `SET_DYNAMIC(out, data, 1, &destructor)`, where `void destructor(void* data)` is a function provided by the user that takes `data` and frees every dynamic member of it and `data` itself. If `destructor` is `NULL`, `free` is used instead.

The point of this API is to simplify memory management and reduce the number of assumptions the user has to make to use the dynamic memory `SET`s.

edwardalee:
As we just discussed in person, the decision is to define `SET_DYNAMIC` with three arguments, a port, a `void*` pointer, and function pointer to a destructor. (No length or element_size argument). If the destructor pointer is null, we will use `free()`.

The `element_size` argument in `SET_ARRAY` is redundant, so we will eliminate that.

Soroosh129:
Great! I like this proposal.

I think @housengw had found a way to use C's `va_arg` mechanism (e.g., `int foo(arg1, ...)`) to overload the `SET` macro for both `SET` and `SET_DYNAMIC`. I think this looks cleaner, but potentially could be confusing to see in C.

What do you think?

housengw:
I think that since the third argument is a function pointer, we should use difference names for the macros to be safe..

lhstrh:
As you are working on this, can I suggest to take the opportunity to switch to lowercase names as brought up in https://github.com/lf-lang/reactor-c/issues/40?

edwardalee:
[...]

PR #66, opened 2022-04-12
Make `SET` work with both token and non-token types

housengw:
Added a destructor variant of `SET` and changed the signature of `SET_ARRAY`. 
Refer to #65 for details.

The LF tests related to this change are added here: https://github.com/lf-lang/lingua-franca/pull/1097


core/reactor.h
```diff
@@ -452,6 +480,8 @@ typedef struct lf_token_t {
     size_t length;
     /** The number of input ports that have not already reacted to the message. */
     int ref_count;
+    /** The  */
```
edwardalee, line 483:
```suggestion
    /** The destructor or NULL to use the default free(). */
```

include/ctarget.h
```diff
@@ -68,6 +68,29 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * Otherwise, free(val) is called.
+ * 
+ * @param out The output port (by name) or input of a contained
+ *  reactor in form input_name.port_name.
+ * @param val The value to insert into the self struct.
+ * @param dtor The function pointer used to free "val" in
+ *             the form of destructor(val).
+ *             If NULL, free(val) is used instead.
+ */
+#define SET_DYNAMIC(out, val, dtor) _LF_SET_DYNAMIC(out, val, dtor)
```
edwardalee, line 92: Update comment and arguments as in .c file.

edwardalee (changes requested):
Let's also change the names to lower case, as suggested, leaving the old names, marked deprecated, for backward compatibility.

core/reactor.h
```diff
@@ -106,6 +106,34 @@ do { \
...
+ * Otherwise, free(val) is called.
+ * 
+ * @param out The output port (by name) or input of a contained
+ *  reactor in form input_name.port_name.
+ * @param val The value to insert into the self struct.
+ * @param dtor The function pointer used to free "val" in
+ *             the form of destructor(val).
+ *             If NULL, free(val) is used instead.
+ */
+#define _LF_SET_DYNAMIC(out, val, dtor) \
```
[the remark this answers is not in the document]
housengw: The reason I used dtor instead of destructor is that C macros does something weird if the parameter name is the same as the field name when I do `token->destructor = destructor;`. It seems to throw an error during compile time.

core/reactor.h
```diff
@@ -120,7 +148,7 @@ do { \
  * @see lf_token_t
  */
 #ifndef __cplusplus
-#define _LF_SET_ARRAY(out, val, element_size, length) \
+#define _LF_SET_ARRAY(out, val, length) \
```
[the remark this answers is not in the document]
housengw: I think we only have to keep the outward facing macros (`SET_ARRAY`) but can change the `_LF...` macros

housengw:
From the error message it seems like the lower case `set` macro is clashing with another function somewhere, probably in included libraries.

```
/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/istream:163:
    In file included from /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/ostream:141:
    /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/bitset:703:17: error: too few arguments provided to function-like macro invocation
        bitset& set() _NOEXCEPT;
                    ^
    /Users/runner/work/reactor-c/reactor-c/test/C/src-gen/target/HelloWorldCCPP/ctarget.h:69:9: note: macro 'set' defined here
    #define set(out, val) _LF_SET(out, val)
```

core/reactor_common.c
```diff
@@ -1697,23 +1720,31 @@ lf_token_t* writable_copy(lf_token_t* token) {
...
         // Count allocations to issue a warning if this is never freed.
-        _lf_count_payload_allocations++;
-        // Create a new, dynamically allocated token.
-        lf_token_t* result = create_token(token->element_size);
-        result->length = token->length;
-        result->value = copy;
-        return result;
+    } else {
+        DEBUG_PRINT("writable_copy: Copy constructor is not NULL. Using copy constructor.");
+        copy = token->copy_constructor(token->value);
```
Soroosh129, line 1738: I wonder if we should also check here if a destructor is set and issue a warning if not.

The potential problem here is that `copy_constructor` could be using something other than `malloc` to allocate memory for `copy`. If the destructor is `NULL`, we will try to call `free` on a memory that was not allocated using `malloc`.

core/tag.c
```diff
@@ -97,6 +98,13 @@ int compare_tags(tag_t tag1, tag_t tag2) {
     }
 }
 
+/**
+ * @deprecated version of 'lf_compare_tags'
+ */
+int compare_tags(tag_t tag1, tag_t tag2) {
+	return lf_compare_tags(tag1, tag2);
```
Soroosh129, line 105: Tabs.

core/tag.c
```diff
@@ -135,32 +143,37 @@ tag_t delay_tag(tag_t tag, interval_t interval) {
 
 /**
  * Return the elapsed logical time in nanoseconds since the start of execution.
+ * @deprecated
  */
 interval_t get_elapsed_logical_time() {
-    return current_tag.time - start_time;
+	return lf_time(LF_ELAPSED_LOGICAL);
```
Soroosh129, line 149: Tabs

core/tag.c
```diff
@@ -135,32 +143,37 @@ tag_t delay_tag(tag_t tag, interval_t interval) {
...
 }
 
 /**
  * Return the current tag, a logical time, microstep pair.
+ * @deprecated
  */
 tag_t get_current_tag() {
-    return current_tag;
+	return lf_tag();
```
Soroosh129, line 157: Tabs (there are more below)

core/tag.h
```diff
@@ -94,32 +95,50 @@ typedef tag_t tag_interval_t;
...
- * Otherwise, add the interval to the time field of the tag and reset
- * the microstep to 0.
- * If the sum overflows, saturate the time value at FOREVER.
- *
- * Note that normally it makes no sense to call this with a negative
- * interval (except NEVER), but this is not checked.
- *
- * @param tag The tag to increment.
- * @param interval The time interval.
+ * An enum for specifying the desired tag when calling "lf_time"
```
Soroosh129, line 102:
```suggestion
 * An enum for specifying the desired time type when calling "lf_time".
```

core/tag.h
```diff
@@ -94,32 +95,50 @@ typedef tag_t tag_interval_t;
...
+ * lf_time(LF_START)
+ * 
+ * - Getting the elapsed physical time:
+ * lf_time(LF_ELAPSED_PHYSICAL)
+ * 
+ * - Getting the logical time
+ * lf_time(LF_LOGICAL)
+ * 
+ * @param type A field in an enum specifying the tag type. 
+ *             See enum "lf_tag_type" above.
```
Soroosh129, line 131:
```suggestion
 *             See enum "lf_time_type" above.
```

include/ctarget_set.h
```diff
@@ -0,0 +1,199 @@
...
+ * If 'lf_set_destructor' is called on 'out', then that destructor
+ * will be used to free 'value'. 
+ * Otherwise, the default void free(void*) function is used.
+ * 
+ * @param out The output port (by name) or input of a contained
+ *  reactor in form input_name.port_name.
+ * @param value The value to insert into the self struct.
+ */
+#define lf_set(out, val) _LF_SET(out, val)
+#define SET(out, val) \
```
Soroosh129, line 76: Is a `do-while` needed here?

See https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html.

Soroosh129 (approved):
Looks great! Thanks for cleaning up the C API.

Superficially, should files like `ctarget_schedule.h` be organized as `ctarget/schedule.h` instead?

lhstrh:
> Looks great! Thanks for cleaning up the C API.
> 
> Superficially, should files like `ctarget_schedule.h` be organized as `ctarget/schedule.h` instead?

I've been critical of the move to a separate `ctarget_schedule.h` because I think it looks weird, but I like your suggestion of having a `ctarget/schedule.h`. I think that might be a satisfying solution for everybody.

edwardalee (approved):
This is too big for me to go through in detail right now, but it looks good. What is the plan for updating the tests? Ideally, they all pass unchanged (backward compatibility), but then we update them so that they serve as proper examples. We also need to update the website code and docs.

housengw:
My plan is to update the tests after merging this PR and https://github.com/lf-lang/lingua-franca/pull/1097.
I will also help with updating the website code

PR #67, opened 2022-04-13
Hashmap

petervdonovan:
This adds another data structure (a simple lookup table) in case it is useful, e.g. for attaching metadata to reactions or to circumstances that a scheduler might encounter at runtime. It is intended to be simple and fast with not many expensive checks. I decided to submit a PR now because the data structure was already implemented, and it would have languished in the branch it was in.

This implementation uses macros to implement generics. This makes it cumbersome and confusing, but I think this is justified by the resulting code reusability.


edwardalee (changes requested):
Wow, this looks like a really clever use of macros! And this looks extremely useful.  The only thing I ask for is some "how to use" documentation in hashmap.h.

lhstrh (approved):

PR #71, opened 2022-04-26
Update time and tag API

housengw:
After the discussion during the LF meeting on April 27 - 2022, a conclusion was reached as to which API to use for time and tag in C target. This PR updates the API accordingly


core/tag.h
```diff
@@ -71,6 +71,32 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // Convenience for converting times
 #define BILLION 1000000000LL
 
+/**
+ * Global physical clock offset.
+ * Initially set according to the RTI's clock in federated
+ * programs.
+ */
+extern interval_t _lf_global_physical_clock_offset;
```
lhstrh, line 79: Would it make sense to call this `_lf_time_physical_offset`? I'm not sure what the `global` part signifies, so I left it out...

core/tag.h
```diff
@@ -71,6 +71,32 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+/**
+ * A test offset that is applied to the clock.
+ * The clock synchronization algorithm must correct for this offset.
+ * This offset is especially useful to test clock synchronization on the
+ * same machine.
+ */
+extern interval_t _lf_global_test_physical_clock_offset;
```
lhstrh, line 87: `_lf_time_physical_test_offset` ?

core/tag.h
```diff
@@ -71,6 +71,32 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+/**
+ * Offset to _LF_CLOCK that would convert it
+ * to epoch time. This is applied to the physical clock
+ * to get a more meaningful and universal time.
+ * 
+ * For CLOCK_REALTIME, this offset is always zero.
+ * For CLOCK_MONOTONIC, it is the difference between those
+ * clocks at the start of the execution.
+ */
+extern interval_t _lf_epoch_offset;
```
lhstrh, line 98: `_lf_time_epoch_offset`?

core/tag.h
```diff
@@ -168,56 +168,60 @@ size_t lf_readable_time(char* buffer, instant_t time);
...
- * 
- * For CLOCK_REALTIME, this offset is always zero.
- * For CLOCK_MONOTONIC, it is the difference between those
- * clocks at the start of the execution.
+ * Return the elapsed logical time in nanoseconds
+ * since the start of execution.
+ * @return A time interval.
  */
-extern interval_t _lf_epoch_offset;
+interval_t lf_time_elapsed_logical(void);
```
lhstrh: I recall from this morning that we proposed `lf_time_logical_elapsed` (and `lf_time_physical_elapsed`). I slightly prefer it this ordering as it makes more sense to me when I think about these concepts in a taxonomy. I think it will also work better with autocomplete.

core/tag.c
```diff
@@ -446,47 +506,4 @@ tag_t _lf_convert_volatile_tag_to_nonvolatile(tag_t volatile const& vtag) {
...
-    case LF_ELAPSED_PHYSICAL:
-        return _lf_physical_time() - physical_start_time;
-    case LF_START:
-        return start_time;
-    default:
-        return NEVER;
-    }
-}
\ No newline at end of file
+#endif
```
lhstrh, line 509: Missing newline

lhstrh (approved):
Looks good to me! I left some suggestions/nitpicks.

core/tag.h
lhstrh: Marking this as resolved since it has already been addressed.

PR #74, opened 2022-05-07
Smarter default number of workers

erlingrj:
Uses the "LF_REACTION_GRAPH_BREADTH" flag proposed here https://github.com/lf-lang/lingua-franca/pull/1139


Soroosh129 (approved):
Looks great!

Issue #75, opened 2022-05-07
More function names needing lf prefix


edwardalee:
I have found a few more functions that probably need the `lf_` prefix:
- request_stop
- print_snapshot
- set_stp_offset
- get_stp_offset

Also, reactor.h seems to have a `get_start_time` that is neither deprecated nor implemented.

lhstrh:
Let's make a fix for this part of the patch release?

PR #77, opened 2022-05-07
`lf_` prefix more functions and reorganize files 

housengw:
`lf_` prefixed functions mentioned in #75 and reorganized user-facing APIs.

APIs changed:
`request_stop` -> `lf_request_stop`
`print_snapshot` -> `lf_print_snapshot`
`set_stp_offset` -> `lf_set_stp_offset`
`get_stp_offset` -> `lf_get_stp_offset`


edwardalee (commented):
Wait, why are all the user facing functions acquiring the prefix '_lf_'?  I thought our convention was that only internal functions, those not intended to be used in reactions, would have that prefix.  User-facing functions, like 'lf_tag', would have the prefix 'lf_'.

housengw:
user-facing functions are acquiring a `lf_` prefix to avoid name collisions. The internal functions will have `_lf_` prefix instead of `lf_`. This would give us more flexibility to change either the user-facing or internal implementation without affecting the other.

housengw:
I changed in all functions in the `reactor-c` core library to use the internal version of the functions `_lf_tag()` instead of `lf_tag()`. The user facing functions (ex. `lf_tag()`) will be defined in `include/ctarget/*.h` instead. Their signatures haven't changed

edwardalee:
Ah, OK, I see. But what is the reason for this? Doesn't this mean that to include some `include/ctarget/*.h` file, you also have to include the `.c` file that defines the `_lf_` function?  The pattern I'm more familiar with is that a `.h` file give the template (not the implementation) for all user-facing functions (the `lf_*` functions), and the `.c` file provides the implementation for these and for any internally used `_lf_` functions, where such functions are not references in any other `.c` file.

housengw:
that's a good point. I will implement the design you mentioned tomorrow..

lhstrh (approved):

PR #80, opened 2022-05-17
Consider the STAA over and above the STA

Soroosh129:


edwardalee (approved):
LGTM! I'm going to go ahead and merge this so that the docs will be in sync with the behavior of the code. Next step is to figure out what to do when STA is violated and there is no handler...

PR #82, opened 2022-05-17
Issue error messages and drop messages when STP violations occur with…

edwardalee:
Issue error messages and drop messages when STP violations occur without a handler

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -924,24 +924,28 @@ bool _lf_worker_handle_STP_violation_for_reaction(int worker_number, reaction_t*
     if (reaction->is_STP_violated == true) {
         reaction_function_t handler = reaction->STP_handler;
         LF_PRINT_LOG("STP violation detected.");
-        // There is a violation
-        violation_occurred = true;
 
         // Invoke the STP handler if there is one.
         if (handler != NULL) {
             LF_PRINT_LOG("Worker %d: Invoking STP violation handler.", worker_number);
+            // There is a violation
+            violation_occurred = true;
             (*handler)(reaction->self);
             
             // If the reaction produced outputs, put the resulting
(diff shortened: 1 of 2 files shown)


core/threaded/reactor_threaded.c
```diff
@@ -937,6 +938,10 @@ bool _lf_worker_handle_STP_violation_for_reaction(int worker_number, reaction_t*
             
             // Reset the is_STP_violated because it has been dealt with
             reaction->is_STP_violated = false;
+        } else {
+        	// The intended tag cannot be respected and there is no handler.
+        	// Print an error message and return true.
+        	lf_print_error("Safe-to-process violation occurred! Discarding message.");
```
Soroosh129, line 944: Should this say something about not having an STP violation handler?

This is not discarding a message entirely, just not invoking this particular reaction.
If another reaction is triggered by the same tardy event but has an STP violation handler, there will not be an error message.

core/threaded/reactor_threaded.c
```diff
@@ -924,19 +924,28 @@ bool _lf_worker_handle_STP_violation_for_reaction(int worker_number, reaction_t*
...
-            
+
             // Reset the is_STP_violated because it has been dealt with
             reaction->is_STP_violated = false;
+        } else {
+        	// The intended tag cannot be respected and there is no handler.
+        	// Print an error message and return true.
+        	// NOTE: STP violations are ignored for control reactions, which need to
+        	// execute anyway.
+        	lf_print_error("STP violation occurred in a trigger to reaction %d, "
```
Soroosh129, line 946: Even though this error may nondeterministically appear, I think this is the best we could do given our current exception handling and annotation capability. There has been interesting advancements on the latter but I think we also need the former.

Soroosh129 (approved):
Looks good to me.

PR #20, opened 2021-11-24
Platform

soyerefsane:
Adding spike platform support for cross code generation.

PR #84, opened 2022-05-31
Baremetal spike support

sberkun:
PR corresponding to https://github.com/lf-lang/lingua-franca/pull/1118. Adds several spike-specific support files in a core/spike/ folder (as well as core/platform/lf_spike_support.c and core/platform/lf_spike_support.c), and tweaks core/platform.h and core/reactor.h to work with riscv-gcc.


core/platform/lf_spike_support.c
```diff
@@ -0,0 +1,108 @@
...
+ *  @author{Efsane Soyer <efsanesoyer@berkeley.edu>}
+ */
+
+#include "lf_spike_support.h"
+#include "../platform.h"
+#include "../tag.h" //needed for BILLION. In flexpret-script under tag.h CLOCK_FREQ is defined, but it's not defined in lf repo 
+
+#ifdef NUMBER_OF_WORKERS
+#if __STDC_VERSION__ < 201112L || defined (__STDC_NO_THREADS__) // (Not C++11 or later) or no threads support
+#include "lf_POSIX_threads_support.c"
```
Soroosh129, line 32: Is POSIX supported on RISC-V? I don't think it is a universal thing.

My guess is that we still don't have a fully fleshed out multi-threading support for RISC-V. If that's the case, here we should throw a compile error (but hopefully we catch this even before compilation).

core/platform/lf_spike_support.c
```diff
@@ -0,0 +1,108 @@
...
+#include "lf_C11_threads_support.c"
+#endif
+#endif
+
+#include <stdio.h> // TODO: remove
+
+// FIXME: This assumption about clock frequency
+// varies by hardware platforms.
+//hardcoded because it wasn't defined in tag.h
+#define CLOCK_FREQ 100000000LL
```
Soroosh129, line 43: It seems like this should be a [target parameter](https://www.lf-lang.org/docs/handbook/target-declaration#summary-of-parameters).

core/platform/lf_spike_support.h
```diff
@@ -0,0 +1,41 @@
...
+ *  
+ *  @author{Efsane Soyer <efsanesoyer@berkeley.edu>}
+ */
+
+#ifndef LF_BAREMETAL_RISCV_SUPPORT_H
+#define LF_BAREMETAL_RISCV_SUPPORT_H
+
+#ifdef NUMBER_OF_WORKERS
+#if __STDC_VERSION__ < 201112L || defined (__STDC_NO_THREADS__) // (Not C++11 or later) or no threads support
+#include "lf_POSIX_threads_support.h"
```
Soroosh129, line 31: Again, I think a compile error here is warranted.

core/spike/common/syscall.c
```diff
@@ -0,0 +1,507 @@
...
+
+#define SYS_write 64
+
+#undef strcmp
+
+extern volatile uint64_t tohost;
+extern volatile uint64_t fromhost;
+
+static uintptr_t syscall(uintptr_t which, uint64_t arg0, uint64_t arg1, uint64_t arg2)
+{
```
Soroosh129, line 19: This code formatting style is different from the existing code in reactor-c. We put `{`s on the same line.

Also, comments are missing for some of the files and the functions within those files.

core/spike/Makefile
```diff
@@ -0,0 +1,47 @@
+# Makefile for compiling LF programs for the risc-v spike platform
```
Soroosh129, line 1: Is the user still supposed to copy the `Makefile` manually? Why not have the code generator do it?
The instructions here seem outdated (what is `demo.riscv`?).

Moreover, it seems like this Makefile could be turned into a CMake-compatible file, which would allow it to be integrated into our existing build system. Conceivably, the CMake-compatible file can automatically be included in the generated CMakeLists.txt by the code generator for the `spike` platform. However, I don't think this is critical at the moment for merging this.

Soroosh129 (changes requested):
This looks great!
I left a few suggestions for changes.

core/spike/Makefile
sberkun: > Is the user still supposed to copy the Makefile manually? Why not have the code generator do it?

Since the spike files are pulled in non-conditionally, it felt strange to have a spike-specific Makefile sitting in the project root when compiling for non-spike platforms.
> what is `demo.riscv`?

default name for the executable; I wasn't sure how to specify a dynamic name so just called it `demo.riscv`.
> Moreover, it seems like this Makefile could be turned into a CMake-compatible file

This would be ideal, but I wasn't sure how to get CMake to invoke riscv64-unknown-elf-gcc. Closest I found was [this stackoverflow answer](https://stackoverflow.com/questions/45933732/how-to-specify-a-compiler-in-cmake), but it says that setting CMAKE_C_COMPILER should be avoided. How would you suggest turning this into a CMake-compatible file?

core/platform.h
```diff
@@ -59,7 +59,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     #include "platform/lf_POSIX_threads_support.h"
 #elif defined(__riscv) || defined(__riscv__) 
     // RISC-V (see https://github.com/riscv/riscv-toolchain-conventions)
-    #error "RISC-V not supported"
+    #include "platform/lf_spike_support.h"
```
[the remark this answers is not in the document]
sberkun: Spike is the simulator, technically the support files are generally for the rv32i architecture. I'm not sure how we would detect rv32i vs another riscv platform like flexpret. @lsk567?

core/spike/Makefile
Soroosh129: > but I wasn't sure how to get CMake to invoke riscv64-unknown-elf-gcc.

If you set the `targetConfig.compiler =  riscv64-unknown-elf-gcc`, then the `CC` environment variable will automatically [be set](https://github.com/lf-lang/lingua-franca/blob/3a3beb070ce5a7a0ff7c1465f770352c34945c64/org.lflang/src/org/lflang/generator/c/CCmakeCompiler.java#L123). According to the StackOverflow answer, this should be enough. 

However, I'm not sure about the extra options that are needed though. Those might require adding additional things to the generated CMakeLists.txt. The best way to do it is probably by using [cmake-include](https://www.lf-lang.org/docs/handbook/target-declaration#cmake-include) to include a (hypothetical) `core/spike/CMakeExtension.txt` (e.g., by adding it to `targetConfig.cmakeIncludes` or `cmakeIncludesWithoutPath`).

sberkun: If we stick with the Makefile, would you suggest conditionally pulling it into the project root based on the "platform" target parameter, always pulling it into project root, or keeping it as-is?

core/platform/lf_spike_support.c
sberkun: Can move this to the Makefile; Should I then have the code generator dynamically edit the Makefile?

core/spike/Makefile
Soroosh129: I don't have a strong preference for this, as long as the instructions are easy enough to find and follow. For example, the code generator can point to a README file or a website link after code generation is finished.

core/platform/lf_spike_support.c
Soroosh129: I think ideally, the code generator should refuse to generate code in this case.

lhstrh:
@sberkun and @lsk567: I noticed this PR has gotten stalled with a pending request for changes. What can we do to move this forward?

sberkun:
iirc I got stuck on translating the Makefile to a equivalent CMake settings, and then summer started. Given that this is a few months old, this will need to be rebased / made to work again in addition to having the changes addressed.

I should some spare time this weekend; I expect rebasing to be doable, but I'm still unsure how to get CMake to invoke `riscv64-unknown-elf-gcc` with the required flags. Do we have any examples of `cmake-includes` being used?

lhstrh:
How about this one: https://github.com/lf-lang/examples-lingua-franca/blob/main/C/src/Rhythm/SensorSimulator.lf

sberkun:
great, will try to set `COMPILE_OPTIONS` within a cmake file like that one. will update with how that goes this weekend

lhstrh:
Awesome. Thanks, @sberkun!

sberkun:
I have a few questions about the current state of the repo:
 - it appears that duplicate generated files appear in `include/core` and `core`. What is the replationship between these two directories?
 - given that `Platform.cmake` can select the platform files based on `${CMAKE_SYSTEM_NAME}`, is there a way to avoid re-selecting platform files in `platform.h`?

lhstrh:
The symlinks are a hack to "fix" the problem we have in our build system. @petervdonovan has been working on addressing this, but his branch runs into segmentation faults in the Python tests on macOS, hence his PR #97 hasn't been merged yet. We are having difficulty understanding where the problems in the Python tests come from and have not been able to reproduce them on a macBook locally...

sberkun:
Interesting, thanks for the info. 

Would you happen to know the base way to pass options to the compiler? I've tried `target_compile_options()`, `add_compile_options()`, and `add_link_options()`, but none of them seem to be taking effect. For now, I've been working around this with `SET(CMAKE_C_FLAGS)`

On a similar note, is there a CMake way to set the linker script? I would like it to use `link.ld` but I can find a way to so.

edwardalee:
If you are targeting a very specific setup and don't need to compile your program on a multiplicity of platforms, a very flexible alternative is to use the `build` target option.  For example, the program at https://github.com/icyphy/lf-buckler/blob/main/src/BuiltInLED.lf uses the script at https://github.com/icyphy/lf-buckler/blob/main/scripts/build_nrf_unix.sh to generate a makefile that compiles the generated C code, links it, and flashes the result onto a USB-connected nRF52 microcontroller board.

lsk567:
Sorry for my delayed response on this PR.

Since there are successful examples with nRF52 and XMOS using the `build` target property, I think it would make sense to get that to work for spike and FlexPRET as well. It seems that many files in this PR can be reused. @sberkun 
 Let's collaborate on this?

sberkun:
Sure, we could do that. I also figured out how to get this to compile with CMake; the main addition would be the following lines in [platform.cmake](https://github.com/lf-lang/reactor-c/blob/main/core/platform/Platform.cmake)
```
elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Spike")
    enable_language(ASM)
    set(LF_PLATFORM_FILE lf_spike_support.c)
    set(CMAKE_C_COMPILER riscv32-unknown-elf-gcc)
    set(CMAKE_ASM_COMPILER riscv32-unknown-elf-gcc)
    SET(CMAKE_C_STANDARD 99)
    add_compile_options(-DPREALLOCATE=1 -mcmodel=medany -static -march=rv32i -mabi=ilp32 -std=gnu99 -Og -ffast-math -fno-common -fno-builtin-printf)
    add_link_options(-T core/spike/common/link.ld -g -static -nostartfiles -lm -lgcc)
```
rebasing is taking a bit longer than I expected, but everything seems to work so far

sberkun:
I think this is almost ready, you can compile a file with the following target parameters:
```
target C {
  threading: false, 
  platform: spike
};
```
and then run `spike --isa=rv32i bin/[name]` to run the resulting executable on spike. 

Some minor issues I'm not sure how to resolve:
-  the clock frequency is set to 100MHz by default (see lf_spike_support.c), but I don't see a way for users to change this. Is there a target parameter for compile definitions?
- the "compiler" target parameter only affects C, and the support files also include assembly. Is there a way to make CMAKE_ASM_COMPILER just track CMAKE_C_COMPILER? 
- the failing CI tests

core/platform.h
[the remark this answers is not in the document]
lhstrh: @lsk567?

lhstrh:
@sberkun: could you hit "resolve" on all the comments that have been addressed? That makes it easier to understand what still needs attention. Thanks!

petervdonovan:
I do not remember whether I removed the `compiler` target parameter, but a quick web search suggests that the `compiler` target parameter can be made compatible with CMake. My goal is to keep `lfc` from directly calling C and assembly compilers so that we can leave that task to CMake.

petervdonovan:
> Some issues I'm not sure how to resolve


1. We pass around a hashmap of compile definitions in LFC ([source](https://github.com/lf-lang/lingua-franca/blob/cebc15c07c9803edd3b3ded279ee97eb61b3b241/org.lflang/src/org/lflang/TargetConfig.java#L119)), but it doesn't seem to be exposed as a target property. We do have a `cmake-include` target property, however. I am not sure if it is a good idea to have that target property (it is maybe too powerful, and it seems to break encapsulation because you have to know where your code is injected into `CMakeLists.txt`), but I think it can do what you want.
2. I've never had to do that, but maybe you can use `cmake-include` to just set those variables equal to each other?
3. As for the failing CI tests... To me, it looks like the `macOS` runner you got was probably just having a bad day. :shrug: I would rerun them.

core/spike/common/syscall.c
sberkun: All of the spike/common files came from https://github.com/riscv-software-src/riscv-tests/tree/master/benchmarks/common  I don't want to mess with them too much, so I've just added a link to the original source + licence at the top of each file.

sberkun:
> Some issues I'm not sure how to resolve
cmake-include is a good way to solve both 1 and 2, the compiler and clock frequency can be changed with the following cmake file:
```
set(CMAKE_C_COMPILER riscv32-unknown-elf-gcc)
set(CMAKE_ASM_COMPILER riscv32-unknown-elf-gcc)
add_compile_definitions(CLOCK_FREQ=5000000LL)
```

lhstrh:
@sberkun @petervdonovan @lsk567 and @erlingrj what is the status of this PR? I saw some discussions in this thread about how the repo was organized, but that has since changed. Is there anything left to be done besides resolve the conflicts? Can this then be merged?

lhstrh:
I think I might have asked this before, but how does this relate to #20, which is really old? Should we close that one?

erlingrj:
I think we should merge in this PR together with FlexPRET support. The timekeeping looks good, but there is no low-power sleep yet, And also no critical sections? Should be possible to achieve both only using RISCV ISA.

lhstrh:
OK, who is taking charge of this effort, @erlingrj? I just want to see that this eventually gets merged...

lsk567:
I can take charge of this effort.

lhstrh:
@lsk567 just reminding you that this needs work...

lsk567:
Yes, indeed. I will get to this issue very soon.

lhstrh:
@lsk567, this PR has been dangling for a long time. What needs to happen to get this merged?

lsk567:
@lhstrh Sorry, this technical debt has been lingering for a while now. I will look into it this weekend and hopefully get it into a mergeable state.

lhstrh:
I'm going to ask one last time @lsk567 and @sberkun and will just close this if it doesn't get addressed soon.

sberkun:
I think it's best to close this; it's too stale and will require major refactors regardless. If this initiative is picked up in the future, it would be more efficient to start from scratch and use this PR as a reference, rather than trying to bring this branch up to speed with the current state of lingua franca.

PR #86, opened 2022-06-09
Treat deadline of zero as always violated

billy-bao:
With this change, if a reactor has a deadline field of zero, the runtime treats it as a deadline with duration zero (so it is always violated). For a description of the issue, see: https://github.com/lf-lang/lingua-franca/issues/1203.


billy-bao:
Yeah, I'm not sure why, unfortunately - it passes locally for me.
Can the system be fast enough so that the physical time value doesn't change between triggering a reaction and checking its deadline? This may explain the inconsistency. If so, maybe I should hard-code a deadline of zero to always fail to ensure deterministic behavior?

edwardalee:
You could force the generator to take at least one jiffy interval before calling lf_set().  The jiffy interval can be as big as 30 msec on some machines, I think.

edwardalee:
Actually, on reflection, I think your suggestion is better.  We should change the logic of deadlines to always trigger the deadline handler if the deadline is zero.  The current design will always be nondeterministic, I think, but I think it's better if the meaning of a zero deadline were deterministic so that it can be used for testing.

lhstrh (approved):
This looks right to me!

Soroosh129 (approved):
LGTM!

PR #89, opened 2022-06-27
Support for Arduino platforms

arengarajan99:


core/tag.h
```diff
@@ -37,6 +37,27 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "platform.h"
 #include "limits.h"
 
+#ifdef GRAN_MICRO
+/* Conversion of time to nanoseconds. */
+#define NSEC(t) (t / 1000L)
```
petervdonovan, line 42: What happens if something gets scheduled 1 nanosecond in the future? I want to make sure that it ends up strictly in the future when it goes into the event queue (and not rounded down to the present).

petervdonovan (approved):
This looks good. Any comments I left are minor compared to other problems we have in the code base.

core/platform/lf_arduino_support.cpp
```diff
@@ -0,0 +1,84 @@
...
+ *  @author{Anirudh Rengarajan <arengarajan@berkeley.edu>}
+ */
+
+
+#include <time.h>
+#include <errno.h>
+
+#include "lf_arduino_support.h"
+#include "../platform.h"
+#include "Arduino.h"
```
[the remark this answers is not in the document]
arengarajan99: Could be mistaken, but Arduino.h is the case sensitive name used by CMake to link said file with the Arduino Core library. Will try testing with lower case to see if we still get working behavior.

core/tag.h
arengarajan99: I believe we set up the workflow such that there is an error when instituting a delay under 3 microseconds due to precision reasons, so we shouldn't run into this problem.

core/platform/Platform.cmake
```diff
@@ -8,6 +8,9 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
     set(LF_PLATFORM_FILE lf_windows_support.c)
     set(CMAKE_SYSTEM_VERSION 10.0)
     message("Using Windows SDK version ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Arduino")
+    set(LF_PLATFORM_FILE lf_arduino_support.cpp)
```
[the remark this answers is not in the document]
arengarajan99: This is a fix required by Arduino-CMake-Toolchain. Without the .cpp extension, the Arduino Build system will not link said library to Arduino's core library, which has CPP Files.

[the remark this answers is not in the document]
arengarajan99: Since it uses a mixture of different fixes that are a part of the C/CCpp targets, I instituted this middle ground solution. That said, it may be better later on to use the CCpp compiler for this purpose.

arengarajan99 (commented):
Looping @edwardalee on this PR since we needed to add 32-Bit and Microsecond Support to Reactor C to support Arduino.

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,84 @@
+/* Arduino Platform API support for the C target of Lingua Franca. */
+
+/*************
+Copyright (c) 2021, The University of California at Berkeley.
```
edwardalee, line 4:
```suggestion
Copyright (c) 2022, The University of California at Berkeley.
```

core/reactor.h
```diff
@@ -512,7 +512,11 @@ struct reaction_t {
     void* self;    // Pointer to a struct with the reactor's state. INSTANCE.
     int number;    // The number of the reaction in the reactor (0 is the first reaction).
     index_t index; // Inverse priority determined by dependency analysis. INSTANCE.
+    #ifdef BIT_32
+    unsigned long chain_id; // Binary encoding of the branches that this reaction has upstream in the dependency graph. INSTANCE.
```
edwardalee, line 516: Is there a reason for changing chain_id to 32 bits? It's not currently used anyway so I think it would be harmless to leave it at 64 bits.

core/reactor_common.c
```diff
@@ -1978,13 +1983,20 @@ void initialize(void) {
     current_tag.time = physical_start_time;
     start_time = current_tag.time;
 
+    #ifdef BIT_32
+    LF_PRINT_DEBUG("Start time: %ldns", start_time);
```
edwardalee, line 1987: Should this be `%ldus`? If the time units are microseconds.

core/tag.h
```diff
@@ -37,6 +37,27 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "platform.h"
 #include "limits.h"
 
+#ifdef GRAN_MICRO
+/* Conversion of time to nanoseconds. */
```
edwardalee, line 41:
```suggestion
/* Conversion of time to microseconds. */
```

core/tag.h
```diff
@@ -37,6 +37,27 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "platform.h"
 #include "limits.h"
 
+#ifdef GRAN_MICRO
```
edwardalee, line 40: Are the flags `GRAN_MICRO` and `BIT_32` any different?  What are they meant to stand for? Perhaps more verbose flags would be better?

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Pause execution for a number of nanoseconds.
+ *
+ * @return 0 for success, or -1 for failure. In case of failure, errno will be
+ *  set appropriately (see `man 2 clock_nanosleep`).
+ */
+int lf_nanosleep(instant_t requested_time) {
+    unsigned int microsec = (unsigned int) requested_time;
+    if(microsec < 3) {
+        Serial.print("Cannot set delay under 3 microseconds due to precision problems");
```
edwardalee, line 49: I think that for nanosleep, there is no reason to have an error here. Just return.

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,84 @@
...
+        return -1;
+    }
+    else if(microsec <= 16383) {
```
edwardalee, line 52: Add a comment here for why this magic threshold?

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Initialize the LF clock.
+ */
+void lf_initialize_clock() {}
+
+/**
+ * Fetch the value of _LF_CLOCK (see lf_linux_support.h) and store it in tp. The
+ * timestamp value in 't' will always be epoch time, which is the number of
+ * nanoseconds since January 1st, 1970.
```
edwardalee, line 69: This comment is no longer correct, right? First, it should be microseconds, not nanoseconds.  Second, I doubt it's epoch time because the number of microseconds since 1970 won't fit in 32 bits.

core/tag.h
edwardalee: The error check for a delay under 3 microseconds is in nanosleep(), I think, which does not get involved when schedule() is called. I think what will happen is that the event will be scheduled one microstep in the future.

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,84 @@
...
+
+#include <time.h>
+#include <errno.h>
+
+#include "lf_arduino_support.h"
+#include "../platform.h"
+#include "Arduino.h"
+
+instant_t ns_to_microsec(instant_t time) {
+    return (time / 1000) - ((time / 1000) % 4);
```
edwardalee, line 37: What is the `((time / 1000) % 4)` for?  Actually, what is this function for? Shouldn't an `instant_t` always be in microseconds?  At a minimum, comments are needed here.

core/reactor_common.c
```diff
@@ -1845,7 +1845,12 @@ int process_args(int argc, char* argv[]) {
             }
             char* time_spec = argv[i++];
             char* units = argv[i++];
+
+            #ifdef BIT_32
+            duration = atol(time_spec);
```
edwardalee, line 1850: There is an issue here that a 32-bit representation of time in microseconds cannot exceed about 2,147 seconds, which is about 35 minutes. If a timeout is given, it cannot exceed 35 minutes.  Also, what will happen to application with no timeout when it runs for more than 35 minutes?

edwardalee (commented):
I think this leaves open the issue of what happens when time overflows.  If a 32-bit long is being used to represent time in microseconds, it will overflow after about 35 minutes.  What happens then?

core/tag.h
arengarajan99: The differences are subtle but necessary. GRAN_MICRO is used to determine what our fundamental unit of time is whereas BIT_32 tells us how large our data structure variable size can be. Arduino has both a fundamental unit of time of microseconds and only supports 32 bit length variable sizes (no long long support). I can clear up the flags a bit better.

arengarajan99: Though we could combine them together as a single flag, there may be use cases down the line where an embedded platform has millisecond fundamental unit of time or supports 32 bit variable size, so it may be fine to keep them split up.

core/platform/lf_arduino_support.c
arengarajan99: Fixed. Had comments copied over from other support files but forgot to change them!

core/platform/lf_arduino_support.c
arengarajan99: Since fundamental unit of time is defined in microseconds, this function has been removed since don't need to run any conversions anymore.

core/platform/lf_arduino_support.c
arengarajan99: Added!

core/reactor_common.c
arengarajan99: Added nested options for what to print.

core/reactor.h
arengarajan99: Arduino AVR Compiler complains that it doesn't understand long long types, so we need to explicitly change all long long values to long regardless of if we use them or not.

core/tag.h
arengarajan99: Changed GRAN_MICRO to MICROSECOND_TIME, kept BIT_32 the same

core/platform/lf_arduino_support.c
arengarajan99: This artifact was from back when we attempted to sync times with the Arduino's internal clock and when we tried to do a nano to micro conversion under the hood, but the solution of changing the time standard turned out to be much cleaner and easier to work with.

core/reactor_common.c
arengarajan99: From Arduino's perspective, the internal timer will overflow back to 0 after around ~70 minutes since its internal clock is tracked as an unsigned 32 bit number. I can do some mathematics for setting an overflow timer to resolve this, but I would need to work out whether we can convert int_32t values to uint_32t values without any issues.

core/reactor.h
edwardalee: Please add a comment to this effect.

core/reactor_common.c
```diff
@@ -1984,13 +1984,21 @@ void initialize(void) {
...
     #else
-    LF_PRINT_DEBUG("Start time: %lldns", start_time);
+        #ifdef MICROSECOND_TIME
+            LF_PRINT_DEBUG("Start time: %ldus", start_time);
+        #else
+            LF_PRINT_DEBUG("Start time: %ldns", start_time);
+        #endif
     #endif
 
     #ifdef ARDUINO
```
edwardalee, line 2000: Why is this not `#ifdef MICROSECOND_TIME`?

core/platform/lf_arduino_support.c
```diff
@@ -33,15 +33,15 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
- * Pause execution for a number of nanoseconds.
+ * Pause execution for a number of microseconds.
+ *
+ * This function works very accurately in the range 3 microseconds and up to 16383.
+ * We cannot assure that delayMicroseconds will perform precisely for smaller delay-times.
+ * Larger delay times may actually delay for an extremely brief time.
  *
  * @return 0 for success, or -1 for failure. In case of failure, errno will be
- *  set appropriately (see `man 2 clock_nanosleep`).
+ *  set appropriately.
```
edwardalee, line 44: The code does not set errno.  Also, I don't understand why an argument under 3 microseconds is an error. Why doesn't this just return?

edwardalee (commented):
A couple more comments embedded.

edwardalee:
Maybe this can be dealt with in a later PR, but I'm concerned about the overflow of the clock, which occurs 35 minutes after booting with a signed integer, 70 minutes with unsigned. I think the LF program will crash rather problematically when this occurs.  At a minimum, we should set a default stop time at the overflow point, exit gracefully (preferably with an informative message about why we are exiting), and document this as a serious current limitation. 

One way to fix this is to assume (and enforce) that we never compare two times that are closer together than 35/2 (or 70/2) minutes.  Then, when two times are compared, if they differ by more than 35/2 (or 70/2), then we assume an overflow occurred between the two times and reverse the comparison.  E.g., time 30 minutes is _less than_ time 2 minutes, but _greater than_ time 20 minutes. This will require changes in the pqueue implementation, which could have performance costs and hence will need to be done carefully.

We should then add validator checks that ensure that timer periods, minimum delays, and stop times never exceed 35/2 (70/2) and a runtime check that schedule() is never called with such a time.

To test the design more conveniently, a good test configuration might be to use nanosecond time scale with a 32-bit clock.  In this case, overflow occurs after 2.1 (4.2) seconds.

core/platform/lf_arduino_support.c
arengarajan99: I set it to error due to the lack of precision for the delay. The alternative would be to give a warning but not error, which I assume is what you'd like be implemented.

arengarajan99:
> Maybe this can be dealt with in a later PR, but I'm concerned about the overflow of the clock, which occurs 35 minutes after booting with a signed integer, 70 minutes with unsigned. I think the LF program will crash rather problematically when this occurs. At a minimum, we should set a default stop time at the overflow point, exit gracefully (preferably with an informative message about why we are exiting), and document this as a serious current limitation.
> 
> One way to fix this is to assume (and enforce) that we never compare two times that are closer together than 35/2 (or 70/2) minutes. Then, when two times are compared, if they differ by more than 35/2 (or 70/2), then we assume an overflow occurred between the two times and reverse the comparison. E.g., time 30 minutes is _less than_ time 2 minutes, but _greater than_ time 20 minutes. This will require changes in the pqueue implementation, which could have performance costs and hence will need to be done carefully.
> 
> We should then add validator checks that ensure that timer periods, minimum delays, and stop times never exceed 35/2 (70/2) and a runtime check that schedule() is never called with such a time.
> 
> To test the design more conveniently, a good test configuration might be to use nanosecond time scale with a 32-bit clock. In this case, overflow occurs after 2.1 (4.2) seconds.

Definitely worth looking into. Arduino's internal clock will automatically overflow back to 0 and count forward. I had a few thoughts on how to address this.

One area we could go down is to basically mimic a 64 bit timer by having 2 32 bit variables: one that tracks the arduino clock and one that increments on every overflow, thereby giving us the ability to treat it like a 64 bit time, but this would need lots of tweaking to make it viable. 

Another area would be to as you said change the pqueue functionality to allow for a wrapping clock, which may be the way to go. 

For the moment, I'll convert the times to unsigned and gracefully end the program once the timer hits 70 minutes.

core/platform/lf_arduino_support.c
edwardalee: The purpose of this function is wait in physical time until physical time closely enough matches logical time. If the precision of the timer is such that it can't wait a small amount of time, then physical time is already closely enough matching logical time. Matching any closer is not possible because of the imprecision of the measurement of time. Hence, I think the function should just return and not issue any warning.

arengarajan99: Makes sense. Will return as normal.

core/reactor.h
arengarajan99: Done!

core/platform/lf_arduino_support.c
```diff
@@ -0,0 +1,82 @@
...
+#include <errno.h>
+
+#include "lf_arduino_support.h"
+#include "../platform.h"
+#include "Arduino.h"
+
+/**
+ * Pause execution for a number of microseconds.
+ *
+ * This function works very accurately in the range 3 microseconds and up to 16383.
```
lhstrh, line 39:
```suggestion
 * This function works very accurately in the range from 3 to 16383 microseconds.
```

core/platform/lf_arduino_support.h
```diff
@@ -0,0 +1,114 @@
...
+typedef int32_t _interval_t;
+
+/**
+ * Microstep instant.
+ */
+typedef uint32_t _microstep_t;
+
+
+
+#endif // LF_ARDUINO_SUPPORT_H
```
lhstrh, line 114:
```suggestion
#endif // LF_ARDUINO_SUPPORT_H

```

core/platform/lf_arduino_support.h
```diff
@@ -0,0 +1,114 @@
+// Arduino does not support threading
+// Assuming single thread; no mutex
```
lhstrh, line 2: These comments are out-of-place.

core/reactor.h
```diff
@@ -512,7 +512,11 @@ struct reaction_t {
     void* self;    // Pointer to a struct with the reactor's state. INSTANCE.
     int number;    // The number of the reaction in the reactor (0 is the first reaction).
     index_t index; // Inverse priority determined by dependency analysis. INSTANCE.
+    #ifdef BIT_32 //Need this wrapper to allow 32 bit systems (i.e. Arduino's AVR Build System) that lack 64 bit support to compile.
```
lhstrh, line 515:
```suggestion
    #ifdef BIT_32 // Use a reduced width for chain IDs on 32-bit systems.
```

core/reactor_common.c
```diff
@@ -1978,13 +1983,28 @@ void initialize(void) {
...
+    #else
+        #ifdef MICROSECOND_TIME
+            LF_PRINT_DEBUG("Start time: %ldus", start_time);
+        #else
+            LF_PRINT_DEBUG("Start time: %ldns", start_time);
+        #endif
+    #endif
 
+    #ifdef ARDUINO
+    printf("---- Start execution at time %ldus\n", physical_start_time);
```
lhstrh, line 2001: This just prints the microseconds. Shouldn't we factor this into seconds + microseconds, like in the `#else` branch?

core/reactor_common.c
lhstrh: +1

core/reactor_common.c
lhstrh: The same issue has been flagged here: https://github.com/icyphy/lf-buckler/issues/3

Should we use an alternative representation that consists of a tuple `(s, m)` where `s` represents the number of seconds and `m` represents the number of microseconds?

lhstrh (approved):
This looks pretty great overall! There are some minor issues/suggestions that I left...

core/reactor_common.c
lhstrh: @arengarajan99: I didn't understand your comment above.

core/reactor_common.c
arengarajan99: Discussed this with @lhstrh at our meeting yesterday. Will probably have a separate pull request for handling overflow since this is an issue for all 32 bit boards. We will most likely need to spread the timer across two variables to remedy the problem at hand. Will work on this in a future PR.

core/platform/lf_arduino_support.h
arengarajan99: Fixed.

core/reactor_common.c
lhstrh: Oh, by "nested options" you must have meant conditionals...

PR #90, opened 2022-06-28
[Draft] Quasi static scheduler runtime

lsk567:
This is a draft PR for the C runtime of the quasi-static (QS) scheduler. The design of the scheduler follows the one in [this report](https://github.com/lf-lang/lingua-franca/files/8882188/Huang.Fang.Lin_Unknown_2022_Static.Parallel.Task.Scheduling.for.Deterministic.Discrete-Event.Simulations.in.Lingua.Franca.pdf) and (mostly) in [this discussion thread](https://github.com/lf-lang/lingua-franca/discussions/1080#discussioncomment-2589055).

This scheduler runtime concurrently must be compiled against a `schedule.h` file, which should be generated by a static schedule generator in `lfc`. A preliminary version of the schedule generator is implemented using [Uclid5](https://github.com/uclid-org/uclid) [on this branch](https://github.com/lf-lang/lingua-franca/tree/quasi-static-scheduler).

Here are the outstanding TODOs in order to make this scheduler production-ready:
[] Support an instruction file specified on the command line. The scheduler then reads the instructions line-by-line from the file during execution. This feature could support scheduler development.
[] Debug the performance issue of this scheduler when running benchmarks, which is significantly slower than the default NP scheduler.
[] Support the `branch` instruction in the instruction set.


core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+_lf_sched_instance_t* _lf_sched_instance;
+
+/////////////////// Scheduler Private API /////////////////////////
+/**
+ * @brief If there is work to be done, notify workers individually.
+ *
+ * This assumes that the caller is not holding any thread mutexes.
+ */
+void _lf_sched_notify_workers() {    
+    // Calculate the number of workers that we need to wake up, which is the
```
edwardalee, line 68: This sentence needs to be completed.

core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+    // Note: All threads are idle. Therefore, there is no need to lock the mutex
+    // while accessing the index for the current level.
```
edwardalee, line 70: It looks like the mutex is held when this is called.

core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+extern const uint32_t* schedule_lengths[];
+
+/////////////////// Scheduler Variables and Structs /////////////////////////
+_lf_sched_instance_t* _lf_sched_instance;
+
+/////////////////// Scheduler Private API /////////////////////////
+/**
+ * @brief If there is work to be done, notify workers individually.
+ *
+ * This assumes that the caller is not holding any thread mutexes.
```
edwardalee, line 65: The caller below is holding the mutex.

core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+
+        
+        _lf_sched_notify_workers();
+    } else {
+        // Not the last thread to become idle. Wait for work to be released.
+        // Wait for the last thread to signal the condition variable.
+        // lf_mutex_lock(&_lf_sched_instance->_lf_sched_semaphore->mutex);
+        // lf_cond_wait(&_lf_sched_instance->_lf_sched_semaphore->cond, &_lf_sched_instance->_lf_sched_semaphore->mutex);
+        // lf_mutex_unlock(&_lf_sched_instance->_lf_sched_semaphore->mutex);
+        lf_cond_wait(&_lf_sched_instance->_lf_sched_array_of_conds[0], &mutex);
```
edwardalee, line 146: Should this index be the worker number rather than 0?

core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+) {
+    LF_PRINT_DEBUG("Scheduler: Initializing with %d workers", number_of_workers);
+
+    // This scheduler is unique in that it requires `num_reactions_per_level` to
+    // work correctly.
+    if (init_sched_instance(&_lf_sched_instance, number_of_workers, params)) {
+        // Scheduler has not been initialized before.
+        if (params == NULL) {
+            lf_print_error_and_exit(
+                "Scheduler: Internal error. The NP scheduler "
```
edwardalee, line 179: Isn't this the QS scheduler, not the NP scheduler? Also, does params really need to be non-null? It is not used below.

core/threaded/scheduler_instance.h
```diff
@@ -165,6 +224,7 @@ bool init_sched_instance(
     }
     lf_mutex_unlock(&mutex);
 
+    // Shaokai: Why are these things in the critical section?
```
edwardalee, line 227: This comment is odd because it's outside the critical section.

core/threaded/scheduler_QS.c
```diff
@@ -0,0 +1,326 @@
...
+                "requires params to be set.");
+        }
+    } else {
+        // Already initialized
+        return;
+    }
+
+    // Initialize the QS-specific fields.
+    _lf_sched_instance->static_schedules = &static_schedules[0];
```
edwardalee, line 188: It looks to me like the global static_schedules does not have any memory associated with it yet.  Where is that memory allocated or defined?

edwardalee (commented):
Looks promising. A few questions included. Also, there is not enough documentation to know how to use it. Ideally, it would explain how to create (manually) a quasi-static schedule somewhere so that one could use these extensions.

core/threaded/scheduler_instance.h
lsk567: Thanks for catching this silly typo of mine... At the time of writing, I meant to say `Why are these things NOT in the critical section?`

core/threaded/scheduler_QS.c
lsk567: Yes, sorry about this. This is copied from the NP scheduler and should be removed.

core/threaded/scheduler_QS.c
lsk567: These are `static const` variables defined in a generated `schedule.h` file, which is included at compile time and is not part of the reactor-c library.

core/threaded/scheduler_QS.c
lsk567: I will see if the mutex can be made unlocked when this is called.

core/threaded/scheduler_QS.c
lsk567: Although the field is called `_lf_sched_array_of_conds`, the current implementation only uses one condition variable to coordinate time advancement. An array is used here so that a future scheduler implementation that needs multiple condition variables can still use this field.

lhstrh: A comment in the code about this would be helpful.

lhstrh:
What is the relationship between this PR and #241?

PR #93, opened 2022-07-17
Update the documentation of vector_at.

### LINKED ISSUE #92: vector implementation does not match documentation
The [vector_at](https://github.com/lf-lang/reactor-c/blob/e1e2bd3b16a3a450c266550ffe745170136fadf0/core/utils/vector.h#L70) function in [vector.h](https://github.com/lf-lang/reactor-c/blob/main/core/utils/vector.h) states "Return a pointer that is contained in the vector at 'idx'."  But this is not what the implementation returns.  The implementation returns a pointer to the pointer that is contained at idx.  Is there a reason for this extra level of indirection in the implementation?

petervdonovan:
Fixes #92.

~~I understand that this is not exactly what was originally suggested, and I'm happy to go back and _actually_ fix the problem.~~

~~However, this procedure was not being used anywhere, and it does not seem strictly necessary because you can always just break the abstraction barrier (which is pretty thin anyway) and reach directly into the `vector_t` struct (although I understand that `vector_at` was originally intended to implement some more complicated logic).~~

### DIFF
--- a/core/utils/vector.c
+++ b/core/utils/vector.c
@@ -90,6 +90,33 @@ void* vector_pop(vector_t* v) {
 }
 
 /**
+ * @brief Return a pointer of the vector element at 'idx'.
+ * 
+ * @param v Any vector.
+ * @param idx The index in the vector.
+ * 
+ * @return NULL on error. A valid pointer to the element at 'idx' otherwise.
+ */
+void** vector_at(vector_t* v, size_t idx) {
+    void** vector_position = v->start + idx;
+    if ((vector_position + 1) > v->next) {
+        v->next = vector_position + 1;
+    }
+    if (v->next >= v->end) {
+        v->votes_required++;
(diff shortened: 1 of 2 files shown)


edwardalee (changes requested):
I suggest instead just changing the documentation for vector_at. The auto-expanding feature is not so trivial.  The following docs could go in both the .h and .c file:
```
/**
 * Return a pointer to where the vector element at 'idx' is stored.
 * This can be used to set the value of the element or to read it.
 * If the index is past the end of the vector, then the vector
 * is automatically expanded and filled with NULL pointers as needed.
 * If no element at `idx` has been previous set, then the value
 * pointed to by the returned pointer will be NULL.
 * 
 * @param v The vector.
 * @param idx The index into the vector.
 * 
 * @return A pointer to the element at 'idx', which is itself a pointer.
 */
```

lhstrh (approved):

PR #97, opened 2022-07-23
Removal of "No CMake" build option

petervdonovan:
This branch deletes 300ish LOC despite incorporating functionality that was factored out of the C generator. (This could be due to deletion of duplicate doc comments.)

These changes only make sense in conjunction with https://github.com/lf-lang/lingua-franca/pull/1299, which removes the "no CMake" option.

I promise to help resolve any conflicts that result from merging this PR.


core/threaded/reactor_threaded.c
```diff
@@ -254,18 +245,18 @@ void _lf_decrement_global_tag_barrier_locked() {
  * been put onto the event queue.
  *
  * If the prposed_tag is greater than the stop tag, then use the stop tag instead.
```
lhstrh, line 247:
```suggestion
 * If the proposed_tag is greater than the stop tag, then use the stop tag instead.
```

core/utils/hashmap.c
```diff
@@ -1,135 +0,0 @@
-/**
- * @author Peter Donovan (peterdonovan@berkeley.edu)
```
lhstrh, line 2: Why are we getting rid of this?

include/core/utils/util.h
```diff
@@ -59,11 +59,33 @@ typedef struct lf_stat_ll {
...
+ * Macro for extracting the level from the index of a reaction.
+ * A reaction that has no upstream reactions has level 0.
+ * Other reactions have a level that is the length of the longest
+ * upstream chain to a reaction with level 0 (inclusive).
+ * This is used, along with the deadline, to sort reactions
+ * in the reaction queue. It ensures that reactions that are
+ * upstream in the dependence graph execute before reactions
+ * that are downstream.
+ */
+#define LEVEL(index) (index & 0xffffLL)
```
lhstrh, line 77: Should this also get the `LF` prefix?

include/core/reactor.h
```diff
@@ -333,313 +308,6 @@ do { \
...
- * Reaction activation record to push onto the reaction queue.
- * Some of the information in this struct is common among all instances
- * of the reactor, and some is specific to each particular instance.
- * These are marked below COMMON or INSTANCE accordingly.
- * The COMMON information is set in the constructor.
- * The fields marked RUNTIME have values that change
- * during execution.
- * Instances of this struct are put onto the reaction queue by the scheduler.
- */
-typedef struct reaction_t reaction_t;
```
lhstrh, line 509: Where did all this stuff go?

lhstrh (approved):
This is a massive amount of work, @petervdonovan! The changes look sensible to me, and I'm excited we're finally getting to merge these :-)

include/core/reactor.h
lhstrh: Answering my own question: `lf_types.h`

core/utils/hashmap.c
petervdonovan: Because it has not been used, and because we can always bring it back if we want. I added it because I thought it might be used imminently, but it turned out not to be necessary.

lhstrh: Why don't we keep it? It's a data structure, so it's of general utility.

petervdonovan: Okay, sure. I'll bring it back.

Issue #88, opened 2022-06-16
Create a LET-aware scheduler


Soroosh129:
Here is the current action plan for a LET scheduler:
- [ ] Determine logical execution time for each reaction during compile time based on immediate downstream logical delays (after delays and actions) and label each reaction with the minimum of all such delays.
- [ ] Create a scheduler where:
    - [ ] When a worker picks up a LET reaction, it withdraws from barrier synchronization on advancement of time (therefore allowing the time to advance around it).
    - [ ] When the withdrawn worker completes executing the LET reaction, the worker rejoins the barrier synchronization on advancement of time.
        - Note: Potentially, we don't need to buffer the outputs because of the AST transformation done on after delays.
- [ ] Augment the scheduler to handle interrupting events (where upstream reactors can produce events while a LET reaction is executing)
    - Possible solution: Force the worker that withdrew from barrier synchronization to rejoin when an interrupting event happens

hokeun:
Thank you, @Soroosh129 ! 

@Wonseo-C, @CloverCho for your information.

edwardalee:
See https://github.com/lf-lang/lingua-franca/issues/1261 for more discussion of LET.

PR #99, opened 2022-08-09
Implementation of LET scheduler

CloverCho:
This PR is aimed to implement the Logical Execution Time (LET) scheduler which is described in #88 and [#1261](https://github.com/lf-lang/lingua-franca/issues/1261)

The scheduler allows time advance if it picks LET reaction


CloverCho:
@Soroosh129 @hokeun 
The scheduler is similar to the GEDF_NP_CI scheduler(I copied its code and modified it a little)
If you want to focus the change, I recommend seeing the following lines in the scheduler_LET.c:
* line 413-415 (init executing_reactions queue)
* line 213-214 (update 'end_timestamp' of the reaction)
* line 467-470 (time advancement)

edwardalee:
> * When the scheduler picks a reaction from 'executing_reactions' queue and give it to a worker, it advance the time to 'end timestamp' of the reaction

I'm confused by this. Does the scheduler pick reactions from `executing_reactions`? I thought `executing_reactions` were reactions that have already been picked up by a worker.  Also, the logical time during the execution of a LET reaction should be *start* time, not the end time.

hokeun:
Predicate to advance time for LET scheduler:

- Triggered reactions queue should be empty.
- Executing reactions queue can be non-empty, provided that all the reactions inside the queue are LET reactions and the minimum of the end time of those LET reactions is larger than the current time.

CloverCho:
@edwardalee Thank you for your detailed feedback!

> Does the scheduler pick reactions from executing_reactions? I thought executing_reactions were reactions that have already been picked up by a worker. 

I designed the new scheduler by referring to the `GEDF_NP_CI` scheduler. The scheduler picks a reaction from `executing_reactions` queue and send it to the worker (It is written at `lf_sched_get_ready_reaction` function in the scheduler, and you can find it [here](https://github.com/lf-lang/reactor-c/blob/6ce1217806ec2421f21e9574c2a1f75b31081e5a/core/threaded/scheduler_GEDF_NP_CI.c#L464)). At this point, the `executing_reactions` is a queue that stores reactions just before being assigned to the worker.

> Also, the logical time during the execution of a LET reaction should be start time, not the end time.

I understand the problem which may occur when the logical time during the execution of a LET reaction is set to the end time, and I fully agree the time should be start time. It is a certainly mistake.

CloverCho:
@Soroosh129 
I remember that you divided the logic of the LET scheduler into two phase:
* Phase 1: Perform time advance when all the reactions in the `executing_reactions` queue are LET reactions
* Phase 2: Handle the situation when the current executing reaction triggers a new LET reaction.

To implement Phase 1, I think you proposed to create three components:
* component 1:  contains executing reactions
* component 2:  handles the time when a reaction started
* component 3:  handles the 'end time' to advance the global time

I think I do not fully understand your suggestion. Could you explain your idea, especially for the three components?

PR #106, opened 2022-09-03
Platform support for Embedded targets

lhstrh:
This PR creates runtime support for (bare-iron) single-threaded targets with interrupts.

- [x] Polish existing NRF52 support
  - Move logic from `wait_until` to `lf_sleep_until`
  - Factor `lf_sleep_until` into smaller functions

- [x] Settle on API changes

  - Move `LF_MIN_SLEEP_DURATION` to platform support
  - Remove `ARDUINO` directive and add `NO_TTY` or something along those lines
  - ...
- [x] Merge changes into `lock-time`
- [x] Merge `main` into `lock-time`
- [x] General cleanup, comments, and removal of spurious `.c`-file includes

Fixes https://github.com/lf-lang/reactor-c/issues/109
Companion of https://github.com/lf-lang/lingua-franca/pull/1348.


core/platform/lf_arduino_support.c
```diff
@@ -33,6 +33,11 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "../platform.h"
 #include "Arduino.h"
 
+/**
+ * Keep track of interrupts being raised.
+ */
+bool _lf_timer_interrupted;
```
edwardalee, line 39: I don't see where this will ever get set to false after once being set to true. Won't the timer always think it has been interrupted after the first time?

core/reactor.c
```diff
@@ -315,24 +272,18 @@ int next(void) {
...
-        // Sleep was interrupted.
-        // FIXME: It is unclear what would cause this to occur in this unthreaded
-        // runtime since lf_schedule() is not thread safe here and should not
-        // be called asynchronously. Perhaps in some runtime such as for a
-        // PRET machine this will be supported, so here we handle this as
-        // if an asynchronous call to schedule has occurred. In that case,
-        // we should return 1 to let the runtime loop around to see what
-        // is on the event queue.
+        // Sleep was interrupted. This could happen when a physical action
+        // gets scheduled from an interrupt service routine.
```
edwardalee, line 278: I think that we have to enter a critical section here and reset _lf_timer_interrupted.

core/reactor.c
```diff
@@ -315,24 +272,18 @@ int next(void) {
...
         return 1;
     }
 
     // At this point, finally, we have an event to process.
     // Advance current time to match that of the first event on the queue.
+    lf_critical_section_enter();
```
edwardalee, line 284: Seems like there is a race condition here. We could get an interrupt between when wait_until returns and here, which means that the next_tag may be obsolete now? Actually, maybe it's OK because if wait_until was not interrupted, then the physical clock will have advanced past next_tag, so even if an interrupt occurred, any events it would have created would have a bigger tag?

edwardalee (commented):
The strategy you have followed here relies on the compiler to detect physical actions and set NUMBER_WORKERS, choosing the threaded runtime.  This is why the functions in [lf_os_single_threaded_support.c](https://github.com/lf-lang/reactor-c/pull/106/files#diff-514d74ecf5160ac667f8bf8a5427a9196451f47f9dbefa2aba764e5ec5cc6b80) can be empty. However, this has a downside: It means the mutex mechanism of the unthreaded runtime is essentially untested. In fact, I have my doubts it is correct.  The mutex is held only for very short times compared to the threaded runtime. Seems like there is real risk of lurking race conditions(?).

Only the Arduino is using this, and that doesn't look correct to me because _lf_timer_interrupted seems to never get reset. Do we have any way to test the Arduino implementation?

What I had in mind was to actually allow physical actions with the unthreaded runtime. This means the mutex acquisition and release should closely mimic (or exactly match) what reactor_threaded.c does.  This also means we could have regression tests that have physical actions with the unthreaded runtime running on all the platforms.

core/reactor.c
lhstrh: All we have to know is that an interruption happened, which is already communicated by the return value of `wait_until`... If an interruption happened, we have to peek again to make sure we got the right tag.

core/reactor.c
lhstrh: I think your concern about the possibility of a race condition has been addressed in a7d20295f7b74a5e4f67beb80494419c38332a30.

core/reactor.c
lhstrh: Oh, you're right about resetting the interrupt flag. This suggests that either we need to make `wait_until` a platform function or we need another function specifically for resetting this flag, which would do nothing for the threaded runtime... We can't just do it in `lf_nanosleep` because the interruption could happen before reaching it (and after leaving the critical section).

core/platform/lf_arduino_support.c
lhstrh: It's happening in `_lf_ack_events`.

core/federated/net_util.c
```diff
@@ -654,8 +655,7 @@ bool extract_match_group(const char* rti_addr, char* dest, regmatch_t group, int
  * Extract match groups from the rti_addr regex.
  * @return true if success, else false.
  */
-bool extract_match_groups(const char* rti_addr, char** rti_addr_strs, bool** rti_addr_flags, regmatch_t* group_array,
-                          int* gids, int* max_lens, int* min_lens, const char** err_msgs) {
+bool extract_match_groups(const char* rti_addr, char** rti_addr_strs, bool** rti_addr_flags, regmatch_t* group_array, int* gids, int* max_lens, int* min_lens, const char** err_msgs) {
```
lhstrh, line 658:
```suggestion
bool extract_match_groups(const char* rti_addr, char** rti_addr_strs, bool** rti_addr_flags, regmatch_t* group_array,
        int* gids, int* max_lens, int* min_lens, const char** err_msgs) {
```

petervdonovan:
The adaptive scheduler just had an STP violation on macOS. Is this something we should be concerned about, or is it just a consequence of the execution time variability that we get on GitHub Actions?

erlingrj:
So the status here is that each single-threaded (bare-metal) target must implement following functions

`int lf_sleep(interval_t sleep_duration);`
Straight-forward. Sleep for requested amount. Return 0 on success and 1 if there was an async event. Care must be taken such that we handle very short sleeps and very long sleeps. For the former its important to make sure that the timeout instant has not passed when we go to sleep (which we cause us to wait for a complete wrap-around. For the latter it must be implemented with a while loop since these embedded platforms typically only has 32bit timers which wraps around maybe after 1 min.

There is one potential problem. What happens if an async event occurs during the call to lf_sleep? It is assumed that it is inside a critical section, so the ISR wont be handled right away. Right before actually going to sleep we leave the critical section. Any pending IRQ will be handled at that point, which will be before the actual sleeping. So it seems as if we could end up in the scenario where we go to sleep waiting for the second element on the pQueue.

We could do a final check of a flag (which can be set by `lf_notify_of_event`) right before the power-down. But I cannot seem to get around the fact that there will always be a time window where if the event occurs within it, we will go to sleep mistakenly.


`int lf_sleep_until(instant_t wakeup_time);`
Same as above.

`int lf_clock_gettime(instant_t* t);`
Here care must be taken if the instant_t has more bits than the HW timer (e.g. 64bit vs 32bit).  Then overflow logic must be added.

`int lf_notify_of_event();`
This function is to be called from an ISR as an async event is added to pQueue through a physical action. If the app was inside a lf_sleep* function it can check whether it was awaken by a Physical Action. If that is NOT the case, then we should just go back to sleep.


`_instant_t type`
`_interval_t type`
`_microstep_t type`

These are OK, basically we can allow resource constrained platforms to use 16 or 32bit to represent time. When looking at the Arduino port I see that this is set to 32bit, which is fine. However I have a few problems with the MICROSECONDS_TIME and 32_BIT macros.

1) In MICROSECOND_TIME the macros use long integers rather than long long. On Arduino and probably most embedded platforms that is 32 bits. The maximum duration you can express is 2,147,483,647 us or ~35 min. If you try MINUTES(36) you will get -35min. So MINUTES, HOURS, DAYS, WEEKS makes no sense when we can only represent 35mins.

2) There is a separate macro for specifying that we want 32bits (BIT_32). But this is convoluted with MICROSECOND_TIME because it already specifies that 32bit. So this macro seems redundant.

3) I don't see where these macros are being defined. I think this should be in the embedded platform port files. Currently it must be inserted by LFC, maybe based on some target property? But the Arduino port is only compatible with microseconds (and NRF port is only compatible with nanoseconds). So either they must be defined in the platform port files or the platform ports must accommodate different time representations and bitwidths.

So to conclude it seems that you want two different capabilities:
1) Flexibly change between lf clock tick (nsec, usec maybe even msec?)
2) Flexible change between 32 and 64 bit representation in key data structures. Which are instant_t, interval_t and apparently the chain_id. 

This might be reasonable, but I would like to see some benchmarks showing that the overhead of using 64bit timestamps on a 32bit arch is unacceptable. Because it does introduce a bunch of

erlingrj:
I have read more on the WFE instruction in the ARM ISA and it seems that we can achieve "safe" sleep where there is no possibility of going to sleep too long if physical action occurs at the wrong time. But our current nrf52_port is vulnerable to this situation. A fix is not possible using the nfr_pwr_mgmt API. 

So the main thing I wanna start a discussion on is how to cleanly handle different time resolutions and representations(bitwidths)

erlingrj:
To further comments.
1. It seems as if Arduino DO support 64 bit. See [here](https://arduino-developer.com/arduino/programming/memory/variables-c). I have not verified as I do not have a arduino board. But if you include <stdint.h> it should work.
2. If we want to use 32bit representation for time, the best would be to use UNSIGNED for instant_t and not SIGNED. The reason is the following. THe interval_t can, and should, be signed. Using unsigned also just makes more sense, what is a negative time instant? Some care has to be taken to catch overflow in interval_t calculation.

core/platform/lf_linux_support.c
```diff
@@ -33,11 +33,14 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "platform.h"
 
 #if defined NUMBER_OF_WORKERS || defined LINGUA_FRANCA_TRACE
-#if __STDC_VERSION__ < 201112L || defined (__STDC_NO_THREADS__)
-#include "lf_POSIX_threads_support.h"  // (Not C++11 or later) or no threads support
+    #if __STDC_VERSION__ < 201112L || defined (__STDC_NO_THREADS__)
+// (Not C++11 or later) or no threads support
+#include "lf_POSIX_threads_support.c"
```
petervdonovan, line 38: I think we decided not to `#include` `.c` files.

include/core/platform/lf_C11_threads_support.h
```diff
@@ -36,6 +36,12 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <stdlib.h>
 #include <stdint.h> // For fixed-width integral types
 
+// The one and only mutex lock.
+_lf_mutex_t mutex;
```
petervdonovan, line 40: I believe it is considered bad practice to define a global variable in a `.h` file. It should just be declared with the `extern` keyword. Same with the function definitions -- if they are in `.h` they should be `static`, but usually they should be in `.c`

core/platform/lf_linux_support.c
lhstrh: Jep. Old code, not adjusted to the new build config yet...

core/reactor.c
```diff
@@ -282,22 +216,22 @@ int _lf_do_step(void) {
...
         // No event in the queue.
-        if (!keepalive_specified) { // FIXME: validator should issue a warning for unthreaded implementation
-                                    // schedule is not thread-safe
-            _lf_set_stop_tag((tag_t){.time=current_tag.time,.microstep=current_tag.microstep+1});
+        if (!keepalive_specified) {
+            _lf_set_stop_tag(
+                (tag_t){.time=current_tag.time, .microstep=current_tag.microstep+1}
+            );
         }
+        next_tag = stop_tag;
```
erlingrj, line 234:
```suggestion
```

core/platform/lf_arduino_support.c
```diff
@@ -42,8 +47,8 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  * @return 0 always.
  */
-int lf_nanosleep(instant_t requested_time) {
-    unsigned int microsec = (unsigned int) requested_time;
+int lf_sleep(interval_t sleep_duration) {
+    unsigned int microsec = (unsigned int) sleep_duration; // FIXME: this cast should not be necessary if the datatype is defined correctly.
```
erlingrj, line 51:
```suggestion
int lf_sleep(interval_t sleep_duration) {
    unsigned int microsec = (unsigned int) (sleep_duration/1000);
```

erlingrj (commented):
Great work Marten! I am little unsure about what is the status of the Arduino work, Anirudh was in the middle of some changes which were gonna move us to 64bit and nanosecond resolution, but don't know where that is

core/reactor.c
lhstrh: Yes, but there's also duplicate code. In addition to removing line 234, lines 243-246 can also be removed, I think.

erlingrj:
I am not able to figure out why some macos tests fails with linker error because `mutex` and `event_q_changed` is undefined. They are defined in a file that is compiled and linked given that the compiler flags are what they are reported to be. I don't see how I can get any further without a Mac to actually see what is happening...

erlingrj:
I am having some trouble understanding the logic behind the platform header files and source files. Maybe some of you can help me understand and maybe we can clean it up a bit in this PR.

1. Inclusions via platform.h
Any file that includes `platform.h` which is a lot. Also includes `lf_linux_support.h` (or similar), which in turn includes `lf_POSIX_threads_support.h` (or similar). I.e. pthreads API is exposed to any file including `platform.h`. Is this necessary? Wouldnt it be better to hide this from user (and runtime) code?

I think `platform.h` should include `lf_linux_support.h` but it should stop there. The translation between lf thread API and pthreads should be handled under the hood in `lf_linux_support.c`.

2. Pthreads but not macos or linux
In `platform.h` we might end up including `lf_POSIC_threads_support.h` directly, not via linux or macos. What types of systems is this for? Currently this will not work since only part of the required API is implemented, `lf_sleep`, `lf_sleep_until` etc is not implemented. My latest change which, fixes the CI, also puts `lf_critical_section_*` into windws/mac/linux support files. 

A. Only include the `lf_POSIX_threads_support.h` and `lf_C11_threads_support.h` from the source files, not from the headers files.
B. Remove support for POSIX without Linux/macos.

If B is a problem we must either:

1) Have POSIX/C11 implemented in source files which can also define the global mutex and cond var. OR
2) Define global mutex/cond var in `reactor_common.c`. Note that the natural thing would be to have it in `reactor_threaded.c` but this wont work because we need it if tracing is enabled. So we must put it in `reactor_common.c` behind macros checking BOTH threading AND tracing.

petervdonovan:
> Also includes lf_linux_support.h (or similar), which in turn includes lf_POSIX_threads_support.h (or similar). I.e. pthreads API is exposed to any file including platform.h.

I'm not sure I understand what you mean by this. There is a preprocessor check to see if we need POSIX or C11 thread support: https://github.com/lf-lang/reactor-c/blob/4c97e960d9a40d60dfb3725678dbd38176023f45/include/core/platform/lf_linux_support.h#L35

erlingrj:
> > Also includes lf_linux_support.h (or similar), which in turn includes lf_POSIX_threads_support.h (or similar). I.e. pthreads API is exposed to any file including platform.h.
> 
> I'm not sure I understand what you mean by this. There is a preprocessor check to see if we need POSIX or C11 thread support:
> 
> https://github.com/lf-lang/reactor-c/blob/4c97e960d9a40d60dfb3725678dbd38176023f45/include/core/platform/lf_linux_support.h#L35

I mean that any file `"#include "platform.h"` will have the whole pthread API in his scope. I think we don't want this for the user-reactions. I also don't think we want that for the run-time in general

edwardalee:
> I am not able to figure out why some macos tests fails with linker error because `mutex` and `event_q_changed` is undefined. They are defined in `lf_macos_support.c` which is compiled and linked given that the compiler flags are what they are reported to be. I don't see how I can get any further without a Mac to actually see what is happening...

Maybe this has something to do with the problem I reported where I almost always have to remove src-gen before compiling to avoid getting linker errors.

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+// Combine 2 32bit works to a 64 bit word
+#define COMBINE_HI_LO(hi,lo) ((((uint64_t) hi) << 32) | ((uint64_t) lo))
+
+// Maximum and minimum sleep possible
+#define LF_MAX_SLEEP_NS USEC(UINT32_MAX)
+#define LF_MIN_SLEEP_NS USEC(5) 
+
+/**
+ * Variable tracking the higher 32bits of the time
+ * Incremented at each timer overflow
```
edwardalee, line 71:
```suggestion
 * Variable tracking the higher 32bits of the time.
 * This is incremented at each timer overflow.
```

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+    error_code = nrfx_timer_init(&g_lf_timer_inst, &timer_conf, &lf_timer_event_handler);
+    APP_ERROR_CHECK(error_code);
+    // Enable an interrupt to occur on channel NRF_TIMER_CC_CHANNEL3
+    // when the timer reaches its maximum value and is about to overflow.
+    nrfx_timer_compare(&g_lf_timer_inst, NRF_TIMER_CC_CHANNEL3, 0x0, true);
+    nrfx_timer_enable(&g_lf_timer_inst);
+}
+
+/**
+ * Fetch the value of _LF_CLOCK (see lf_linux_support.h) and store it in tp. The
```
edwardalee, line 131:
```suggestion
 * Fetch the value of _LF_CLOCK (see lf_linux_support.h) and store it in *t. The
```

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+ * timestamp value in 't' will will be the number of nanoseconds since the board was reset.
+ * The timers on the board have only 32 bits and their resolution is in microseconds, so
+ * the time returned will always be an even number of microseconds. Moreover, after about 71
```
edwardalee, line 134:
```suggestion
 * the time returned will always be an integer number of microseconds. Moreover, after about 71
```

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+ * minutes of operation, the timer overflows. 
+ * 
+ * The function reads out the upper word before and after reading the timer.
+ * If the upper word has changed (i.e. it was an overflow in between),
```
edwardalee, line 138:
```suggestion
 * If the upper word has changed (i.e. there was an overflow in between),
```

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+ * we cannot simply combine them. We read once more to be sure that 
+ * we read after the overflow.
+ *
+ * @return 0 for success, or -1 for failure. In case of failure, errno will be
+ *  set appropriately (see `man 2 clock_gettime`).
+ */
+int lf_clock_gettime(instant_t* t) {
+    if (t == NULL) {
```
edwardalee, line 146: Replace with assert?

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+/**
+ * @brief Return whether the critical section has been entered.
+ * 
+ * @return true if interrupts are currently disabled
+ * @return false if interrupts are currently enabled
+ */
+bool in_critical_section() {
+    // FIXME: if somehow interrupts get disabled directly (not through the NRF API),
```
edwardalee, line 174: See my suggested implementation in the Arduino code for an alternative that will address this FIXME and also support nested critical sections.

core/platform/lf_nrf52_support.c
```diff
@@ -0,0 +1,310 @@
...
+    
+    if (!_lf_async_event) {
+        return 0;
+    } else {
+        LF_PRINT_DEBUG("Sleep got interrupted...\n");
+        return -1;
+    }
+}
+
+int lf_critical_section_enter() {
```
edwardalee, line 290: Comments needed for these functions.

Also, See my suggested implementation in the Arduino code for an alternative that will address this FIXME and also support nested critical sections.

edwardalee (approved):
Looks great!  I made some suggestions that will allow critical sections to be nested plus some cosmetic suggestions.  Throughout, comments should be complete sentences with punctuation. I only marked a few.

erlingrj:
This proves to be a bit hard to debug. Any advice. status is:
HelloWorld.lf and CountTest.lf fails on Windows. It produces no output so it could be a runtime assertion done by Visual Studio. However:
1) I can only reproduce the test failures if I run ALL the C tests using gradlew. `./gradlew test --tests org,lflang.tests.runtime.CTest.*`
2) Running the tests directly with LFC does not reproduce it (neither locally nor in CI)
3) Running the tests directly with gradlew does not reproduce it `./gradlew runSingleTest --args test/C/src/HelloWorld.lf`
4) Running the **generic** test suite which includes these tests also does not reproduce it `./gradlew test --tests org.lflang.tests.runtime.CTest.runGenericTests`

Any suggestions. The whole Gradle system is quite black-box to me. What is happening under the hood that could cause something like this

petervdonovan:
> It produces no output so it could be a runtime assertion done by Visual Studio.

Exit code -1073741819 means access violation. I think this is usually a segfault.

erlingrj:
Did you learn anything @petervdonovan? The problem is that I am only able to reproduce it running all the tests from Gradle. How Gradle runs the test is still a mystery for me, if I could get an executable with debug flags enabled that yielded a seg-fault I would at least be able to understand something.

erlingrj:
How is the single-threaded runtime and tracing supposed to work together? If tracing is defined then we include the threaded support files regardless of whether we have set threading to true or false. 

This could be for two reasons:
1) The tracing mechanism needs threading so it needs access to lf_thread_create, lf_mutex etc.
2) The single-threaded for some reason needs a different support files also in presence of the tracing mechanism.

Can anyone help me clarify this?

edwardalee:
It's been a while since I looked at this code, but I think the new API for entering a critical section will take care of a big part of the problem, which is that tracing calls write to a common data structure.  Alternatively, this could be refactored to have a separate data structure per worker. I think the reason for needing to create a thread is to write to a file when the data structure fills up.  I guess this could be handled differently in the unthreaded runtime with conditional code.

erlingrj:
> It's been a while since I looked at this code, but I think the new API for entering a critical section will take care of a big part of the problem, which is that tracing calls write to a common data structure. Alternatively, this could be refactored to have a separate data structure per worker. I think the reason for needing to create a thread is to write to a file when the data structure fills up. I guess this could be handled differently in the unthreaded runtime with conditional code.

OK, so you are saying that if we are using the **unthreaded runtime** with tracing enabled, `critical_section_enter()` should result in acquiring a global mutex, which is also acquired by the tracing thread? This is not implemented as of now. Actually, in the unthreaded runtime, `critical_section_enter()` does nothing. I guess it should at least disable interrupts. 

I don`t see the place in `trace.c` where we are accessing a data structure shared with the runtime. I.e. where is trace.c calling `critical_section_enter()`

erlingrj:
@lhstrh. After a converation with @edwardalee about tracing today I have concluded that the current implementation is sufficient. The tracing only needs to synchronize access "internally". This is done with a mutex local to `tracing.c`. It seemed to be an edge case when we had unthreaded+tracing. But I don't think it matters. How it looks is like this:

From all the source files including `plaform.h` EXCEPT `tracing.c` will see the unthreaded API and its implementation of critical section etc. For the single source file `tracing.c` it will see the threaded API (and its implementation, the function definitions are in the header files as well). So when we do some tracing the program will call into `tracing.c` and during that period it will get access to the threaded API and can synchronize with the flushing thread.

I will do a pass through the PR and address Edwards comments and fix what I see, then I guess we could do a final review and merge?

core/platform/lf_os_single_threaded_support.c
```diff
@@ -0,0 +1,34 @@
...
+ * to compile. If threads are needed, use a multi-threaded runtime instead.
+ * @copyright BSD 2-Clause License (see LICENSE.md)
+ */
+
+#if defined(_THREADS_H) || defined(_PTHREAD_H)
+    #error Usage of threads in the single-threaded runtime is not safe.
+#endif
+
+int lf_critical_section_enter() {
+    // FIXME: Is this what we want? Dont we want to grab a mutex here?
```
lhstrh, line 18: I suppose on could manually add threading support while still using the single-threaded runtime, yes. But does it make sense to do this, or should we just say that this is not the right way. If we do take this stance, is there a foolproof way of detecting this and telling the programmer not to?

core/platform/lf_windows_support.c
```diff
@@ -67,186 +71,8 @@ int lf_available_cores() {
...
-
-    // Success
-    return 0;
-}
-
-
 #else
-#include "lf_C11_threads_support.h"
-#endif
+#include "lf_os_single_threaded_support.c"
```
lhstrh, line 75: This looks sus. What's going on here?

core/platform/lf_windows_support.c
```diff
@@ -335,3 +161,17 @@ int lf_nanosleep(instant_t requested_time) {
...
+    if (sleep_duration < LF_MIN_SLEEP_NS) {
+        return 0;
+    } else {
+        return lf_sleep(sleep_duration);
+    }
+}
+
+int lf_nanosleep(interval_t sleep_duration) {
+    return lf_sleep(sleep_duration);
+}
```
lhstrh, line 177:
```suggestion
}

```

core/reactor.c
```diff
@@ -433,3 +366,17 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
...
+void _lf_notify_of_event() {
+    lf_notify_of_event();
+}
+
+void _lf_critical_section_enter() {
+    lf_critical_section_enter();
+}
+void _lf_critical_section_exit() {
+    lf_critical_section_exit();
+}
```
lhstrh, line 382:
```suggestion
}

```

core/threaded/reactor_threaded.c
```diff
@@ -1228,3 +1170,28 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
         return -1;
     }
 }
+
+/**
+ * @brief Notification of new event is implemented by broadcasting on a 
+ * condition variable. 
+ * FIXME: Are we sure that it is not sufficient with a `signal`?
```
lhstrh, line 1177: Flagging this as something that needs to be resolved before merging.

core/trace.c
```diff
@@ -29,14 +29,19 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * Include this file instead of trace.h to get tracing.
  * See trace.h file for instructions.
  */
-
+// FIXME: Should be renamed if we use the hack below maybe: LF_INCLUDE_TRACE_FILES
```
lhstrh, line 32: Also this `FIXME` should not be left unaddressed.

include/core/platform.h
```diff
@@ -42,8 +42,20 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef PLATFORM_H
 #define PLATFORM_H
 
-#if defined(ARDUINO)
+// FIXME: We could also use a single flag since this is truly binary
```
lhstrh, line 45: ?

core/trace.c
```diff
@@ -29,14 +29,19 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #ifdef LINGUA_FRANCA_TRACE
```
lhstrh, line 33: This guard seems wrong. Shouldn't it be `LF_TRACING`?

include/core/platform/lf_nrf52_support.h
```diff
@@ -0,0 +1,87 @@
...
+ * Text descriptions and spaces add an additional 55,
+ * for a total of 79. One more allows for a null terminator.
+ */
+#define LF_TIME_BUFFER_LENGTH 80
+
+
+// The underlying physical clock for Linux
+#define _LF_CLOCK CLOCK_MONOTONIC
+
+#endif // LF_nRF52832_SUPPORT_H
```
lhstrh, line 87:
```suggestion
#endif // LF_nRF52832_SUPPORT_H

```

lhstrh (commented):
Left some comments...

edwardalee:
What is the status of this lock-time PR?

lhstrh:
It's almost ready, but there are still a few small things that need @erlingrj's attention (comments :point_up:). Once addressed, we can merge.

core/platform/lf_os_single_threaded_support.c
erlingrj: I think the unthreaded runtime  + Linux is intended to be a very special case where you should not use other threads to schedule physical actions. I think the macro above would catch such a scenario. I think we can close this for now. I will remove the fixme

core/platform/lf_arduino_support.c
```diff
@@ -1,82 +1,167 @@
...
+MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+***************/
+
+/** Arduino API support for the C target of Lingua Franca.
```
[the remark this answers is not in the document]
lhstrh: I'm taking the absence of an answer from @arengarajan99 to be a "no."

include/core/platform.h
lhstrh: Hm, that sounds confusing though. What if both are true?

core/trace.c
lhstrh: That looks fine to me.

PR #113, opened 2022-09-27
Address memory leaks in RTI

erlingrj:
While we were looking at RTI implementation several potential memory leaks were identified


edwardalee (approved):
Good catches!!!

PR #115, opened 2022-10-07
Fix of flaky timedwait on Windows

cmnrd:
The Windows implementation of `lf_timed_wait` calculates the difference between the timepoint to wait until and the current physical time. In some seldom cases, the current physical time could already be larger then the specified timepoint. This will lead to a negative delay passed to `SleepConditionVariableCS`. While the documentation does not specify this precisely, experimentation showed that negative delays are interpreted as infinite delays and hence the program would block forever.

This fix makes two changes. First, it checks for negative delays and returns without waiting if the delay is less or equal to zero. Second, the number of milliseconds to wait for is rounded up in order to prevent unnecessary busy waiting when the call to `SleepConditionVariableCS` returns early.

### DIFF
--- a/core/platform/lf_windows_support.c
+++ b/core/platform/lf_windows_support.c
@@ -223,7 +223,7 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
     }
 
     // convert ns to ms and round up to closest full integer
-    DWORD relative_time_ms = relative_time_ns + 999999LL / 1000000LL;
+    DWORD relative_time_ms = (relative_time_ns + 999999LL) / 1000000LL;
 
     int return_value =
      (int)SleepConditionVariableCS(
--- a/core/platform/lf_windows_support.c
+++ b/core/platform/lf_windows_support.c
@@ -216,7 +216,14 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
     // Convert the absolute time to a relative time
     instant_t current_time_ns;
     lf_clock_gettime(&current_time_ns);
-    DWORD relative_time_ms = (absolute_time_ns - current_time_ns)/1000000LL;
+    DWORD relative_time_ns = (absolute_time_ns - current_time_ns);
+    if (relative_time_ns <= 0){
+      // physical time has already catched up sufficiently and we do not need to wait anymore
+      return 0;
+    }
+
+    // convert ns to ms and round up to closest full integer
+    DWORD relative_time_ms = relative_time_ns + 999999LL / 1000000LL;
 
     int return_value =
      (int)SleepConditionVariableCS(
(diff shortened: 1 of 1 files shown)


core/platform/lf_windows_support.c
```diff
@@ -216,7 +216,14 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
...
     lf_clock_gettime(&current_time_ns);
-    DWORD relative_time_ms = (absolute_time_ns - current_time_ns)/1000000LL;
+    DWORD relative_time_ns = (absolute_time_ns - current_time_ns);
+    if (relative_time_ns <= 0){
+      // physical time has already catched up sufficiently and we do not need to wait anymore
+      return 0;
+    }
+
+    // convert ns to ms and round up to closest full integer
+    DWORD relative_time_ms = relative_time_ns + 999999LL / 1000000LL;
```
petervdonovan, line 226:
```suggestion
    DWORD relative_time_ms = (relative_time_ns + 999999LL) / 1000000LL;
```

petervdonovan:
Thank you Christian!

core/platform/lf_windows_support.c
```diff
@@ -216,7 +216,14 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
     // Convert the absolute time to a relative time
     instant_t current_time_ns;
     lf_clock_gettime(&current_time_ns);
-    DWORD relative_time_ms = (absolute_time_ns - current_time_ns)/1000000LL;
+    DWORD relative_time_ns = (absolute_time_ns - current_time_ns);
+    if (relative_time_ns <= 0){
+      // physical time has already catched up sufficiently and we do not need to wait anymore
```
lhstrh, line 221:
```suggestion
      // physical time has already caught up sufficiently and we do not need to wait anymore
```

core/platform/lf_windows_support.c
```diff
@@ -216,7 +216,14 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
     // Convert the absolute time to a relative time
     instant_t current_time_ns;
     lf_clock_gettime(&current_time_ns);
-    DWORD relative_time_ms = (absolute_time_ns - current_time_ns)/1000000LL;
+    DWORD relative_time_ns = (absolute_time_ns - current_time_ns);
+    if (relative_time_ns <= 0){
```
lhstrh, line 220:
```suggestion
    if (relative_time_ns <= 0) {
```

lhstrh (approved):
Terrific work tracking this down, @cmnrd! I added some nitpicks and noticed some unreachable code on lines `241` and `246` that we might as well get rid of.

edwardalee (approved):
Great catch about the negative time interval! This looks like a good solution to me, but rounding up does mean that time precision below one ms will be impossible on Windows machines.  E.g., any deadline below 1 ms is likely to be violated.  But this probably true anyway on any non-RTOS because of the OS jiffy interval (or whatever it is called in Windows).

PR #118, opened 2022-10-24
Support Modal Reactors with LET

erlingrj:
This is a PR into the Draft PR for LET scheduling. It adds support for handling Modal Reactors. 

The challenge stems from the fact that a LET reaction may execute inside a Modal Reactor while another reaction triggers a mode change. Either in the same reactor or a containing reactor. In that case the LET reaction must stall the mode change. This is achieved by having a "local" mutex per reactor with LET reactions which is locked while the reaction is executing. Before a mode change can occur the mutex of the reactor and all contained reactor must be locked, if they are present.

The challenge stems from the fact that **we have to acquire the local mutex before the global mutex** to avoid deadlocks. Unfortunately the functions for doing mode changes in modes.c are called while holding the global mutex. My approach was to add a prologue and epilogue which is executed from `scheduler_LET.c` that locks and unlocks the required local mutexes **before** the global mutex is locked and the worker tries to advance time.

A companion PR for Lingua-Franca is [here](https://github.com/lf-lang/lingua-franca/pull/1427)

@edwardalee and @a-sr I am tagging both of you here. Would be very nice with some feedback on this approach. I don't have the full picture on Modal Reactors and might have missed something

### DIFF
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -34,8 +34,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  * @author{Alexander Schulz-Rosengarten <als@informatik.uni-kiel.de>}
  * @author{Soroush Bateni <soroush@utdallas.edu}
- * FIXME: Before performing any mode transitions the reactor-local mutex must be acquired.
- *  this will make sure that we dont do any mode transition while executing a LET reaction
+ * @author{Erling R. Jellum <erling.r.jellum@ntnu.no}
  */
 #ifdef MODAL_REACTORS
 
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -591,7 +591,7 @@ int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int s
                 // See if its parent has done any volunteered mode transition
                  for (int i = 0; i< num_transitioning_reactors; i++) {
                     // Start search from the end. Look for a reactor pointer equal to self pointer of parent
-                    if (vector_at(return_vec, num_transitioning_reactors-1-i) == (void *) state->parent_mode->state->self) {
+                    if ( ((self_base_t *) vector_at(return_vec, num_transitioning_reactors-1-i)) == state->parent_mode->state->self) {
                         vector_push(return_vec, state->self);
                         num_transitioning_reactors++;
                         break;
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -370,6 +370,15 @@ void _lf_process_mode_changes(
             // Handle effect of entering next mode
             if (state->next_mode != NULL) {
                 LF_PRINT_DEBUG("Modes: Transition to %s.", state->next_mode->name);
+                // Grab Reactor mutex if it is present. This enables LET scheduling
+                // FIXME: Verify that there is no race condition. Can a worker advance time BEFORE the LET
+                //  worker has grabbed the mutex? 
+                if (state->self->has_mutex) {
+                    LF_PRINT_DEBUG("Modes: Acquire mutex of reactor");
+                    lf_mutex_lock(&state->self->mutex);
+                    LF_PRINT_DEBUG("Modes: Mutex acquired of reactor");
+                }
+
                 transition = true;
 
                 if (state->mode_change == reset_transition) {
@@ -440,6 +449,12 @@ void _lf_process_mode_changes(
                     } else {
                         suspended_event = suspended_event->next;
                     }
+                }
+
+                // Unlock mutex after mode transition has been completed. 
+                if (state->self->has_mutex) {
+                    LF_PRINT_DEBUG("Modes: release mutex of reactor");
+                    lf_mutex_unlock(&state->self->mutex);
                 }
             }
         }
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -557,4 +557,51 @@ void _lf_terminate_modal_reactors() {
     }
     _lf_unsused_suspended_events_head = NULL;
 }
+
+/**
+ * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec 
+ *  for all reactors that will do a transition
+ * 
+ *  A mode does a transition of either:
+ *  1) next_mode is set or
+ *  2) parent has next_mode set
+ * 
+ *  A preliminary implementation is simple:
+ *  1) Check if reactor has next_mode set. If so, add to return_vec 
+ *  2) If not, check if its parent has already been added to return_vec. If so, add it
+ *  
+ *  This assumes that the `states` argument is an array ordered hierarchically. It does a bunch of searching through
+ *  the return_vec, but this might be OK.   
+ * 
+ * @param states 
+ * @param states_size 
+ * @param return_vec 
+ * @return int 
+ */
+int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int states_size, void * _return_vec) {
+    int num_transitioning_reactors = 0;
+    vector_t *return_vec = (vector_t *) _return_vec;
+    for (int i = 0; i < states_size; i++) {
+        reactor_mode_state_t* state = states[i];
+        if (state != NULL) {
+            // Check if reactor is doing a "volunteering" mode transition (scheduled from within itself)
+            if (state->next_mode != NULL) {
+                vector_push(return_vec, state->self);
+                num_transitioning_reactors++;
+            } else {
+                // See if its parent has done any volunteered mode transition
+                 for (int i = 0; i< num_transitioning_reactors; i++) {
+                    // Start search from the end. Look for a reactor pointer equal to self pointer of parent
+                    if (vector_at(return_vec, num_transitioning_reactors-1-i) == (void *) state->parent_mode->state->self) {
+                        vector_push(return_vec, state->self);
+                        num_transitioning_reactors++;
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    return num_transitioning_reactors;
+}
+
 #endif
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -370,15 +370,6 @@ void _lf_process_mode_changes(
             // Handle effect of entering next mode
             if (state->next_mode != NULL) {
                 LF_PRINT_DEBUG("Modes: Transition to %s.", state->next_mode->name);
-                // Grab Reactor mutex if it is present. This enables LET scheduling
-                // FIXME: Verify that there is no race condition. Can a worker advance time BEFORE the LET
-                //  worker has grabbed the mutex? 
-                if (state->self->has_mutex) {
-                    LF_PRINT_DEBUG("Modes: Acquire mutex of reactor");
-                    lf_mutex_lock(&state->self->mutex);
-                    LF_PRINT_DEBUG("Modes: Mutex acquired of reactor");
-                }
-
                 transition = true;
 
                 if (state->mode_change == reset_transition) {
@@ -451,11 +442,6 @@ void _lf_process_mode_changes(
                     }
                 }
 
-                // Unlock mutex after mode transition has been completed. 
-                if (state->self->has_mutex) {
-                    LF_PRINT_DEBUG("Modes: release mutex of reactor");
-                    lf_mutex_unlock(&state->self->mutex);
-                }
             }
         }
     }
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -559,8 +559,8 @@ void _lf_terminate_modal_reactors() {
 }
 
 /**
- * @brief This function accepts an array of ALL the mode_states and adds a reactor ptr into return_vec 
- *  for all reactors that will perform a transition at the current tag.
+ * @brief This function accepts an array of ALL the modes ordered from top-level and ending with leaf modes, 
+ *  and adds a reactor ptr into return_vec for all reactors that will perform a transition at the current tag.
  * 
  *  A mode performs a transition if either:
  *  1) next_mode is set or
@@ -570,8 +570,7 @@ void _lf_terminate_modal_reactors() {
  *  1) Check if reactor has next_mode set. If so, add to return_vec 
  *  2) If not, check if its parent has already been added to return_vec. If so, add it
  *  
- *  This assumes that the `states` argument is an array ordered hierarchically. It does a bunch of searching through
- *  the return_vec, but this might be OK.   
+ *  This assumes that the `states` argument is an array ordered from top-level mode to leaf mode
  * 
  * @param states 
  * @param states_size 
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -559,7 +559,7 @@ void _lf_terminate_modal_reactors() {
 }
 
 /**
- * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec 
+ * @brief This function accepts an array of ALL the mode_states and adds a reactor ptr into return_vec 
  *  for all reactors that will perform a transition at the current tag.
  * 
  *  A mode performs a transition if either:
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -584,9 +584,11 @@ int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int s
         reactor_mode_state_t* state = states[i];
         if (state != NULL) {
             // Check if reactor is doing a "volunteering" mode transition (scheduled from within itself)
+            bool mode_transition = false;
             if (state->next_mode != NULL) {
                 vector_push(return_vec, state->self);
                 num_transitioning_reactors++;
+                mode_transition = true;
             } else {
                 // See if its parent has done any volunteered mode transition
                  for (int i = 0; i< num_transitioning_reactors; i++) {
@@ -594,9 +596,16 @@ int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int s
                     if ( ((self_base_t *) vector_at(return_vec, num_transitioning_reactors-1-i)) == state->parent_mode->state->self) {
                         vector_push(return_vec, state->self);
                         num_transitioning_reactors++;
+                        mode_transition = true;
                         break;
                     }
                 }
+            }
+            
+
+            if (mode_transition) {
+                // This Reactor is performing a mode change. In which case we should grab mutexes of all containing reactors as well
+                _lf_mode_collect_containing_reactors(state->self, return_vec); // FIXME: Implement me
             }
         }
     }
--- a/core/modal_models/modes.c
+++ b/core/modal_models/modes.c
@@ -560,17 +560,8 @@ void _lf_terminate_modal_reactors() {
 
 /**
  * @brief This function accepts an array of ALL the modes ordered from top-level and ending with leaf modes, 
- *  and adds a reactor ptr into return_vec for all reactors that will perform a transition at the current tag.
- * 
- *  A mode performs a transition if either:
- *  1) next_mode is set or
- *  2) any parent has next_mode set.
- * 
- *  A preliminary implementation is simple:
- *  1) Check if reactor has next_mode set. If so, add to return_vec 
- *  2) If not, check if its parent has already been added to return_vec. If so, add it
- *  
- *  This assumes that the `states` argument is an array ordered from top-level mode to leaf mode
+ *  and adds a reactor ptr into return_vec if a mode transition was requested. This will not add reactors having 
+ *  contained modes. Only reactors that have scheduled a transition itself.
  * 
  * @param states 
  * @param states_size 
@@ -583,29 +574,10 @@ int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int s
     for (int i = 0; i < states_size; i++) {
         reactor_mode_state_t* state = states[i];
         if (state != NULL) {
-            // Check if reactor is doing a "volunteering" mode transition (scheduled from within itself)
-            bool mode_transition = false;
             if (state->next_mode != NULL) {
                 vector_push(return_vec, state->self);
                 num_transitioning_reactors++;
-                mode_transition = true;
-            } else {
-                // See if its parent has done any volunteered mode transition
-                 for (int i = 0; i< num_transitioning_reactors; i++) {
-                    // Start search from the end. Look for a reactor pointer equal to self pointer of parent
-                    if ( ((self_base_t *) vector_at(return_vec, num_transitioning_reactors-1-i)) == state->parent_mode->state->self) {
-                        vector_push(return_vec, state->self);
-                        num_transitioning_reactors++;
-                        mode_transition = true;
-                        break;
-                    }
-                }
-            }
-            
-
-            if (mode_transition) {
-                // This Reactor is performing a mode change. In which case we should grab mutexes of all containing reactors as well
-                _lf_mode_collect_containing_reactors(state->self, return_vec); // FIXME: Implement me
+                
             }
         }
     }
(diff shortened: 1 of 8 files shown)


edwardalee (approved):
This looks good, module some questions inserted in the code.  One big question: if you have a mode that contains within two reactors, each of which, at the same tag, start a LET reaction, then I'm guessing that the second of these will block upon attempting to acquire the modal reactor's mutex. Right?  Is that what we want?  It will effectively sequentialities the two LET reactions.  This seems like a pretty esoteric corner case, but at a minimum we should add a test case to make sure we don't deadlock.

core/threaded/scheduler_LET.c
```diff
@@ -516,4 +549,54 @@ void lf_sched_reaction_epilogue(reaction_t * reaction, int worker_number) {
...
+    // FIXME: In this loop we are actually overwriting the transitions of containing modes
+    //  this is also done later inside the `next` function. I dont think it has any effect.
+    for (int i = 0; i < states_size; i++) {
+        reactor_mode_state_t* state = states[i];
+        if (state != NULL) {
+            // Hierarchical reset: if this mode has parent that is entered in
+            // this step with a reset this reactor has to enter its initial mode
+            if (state->parent_mode != NULL
+                    && state->parent_mode->state != NULL
+                    && state->parent_mode->state->next_mode == state->parent_mode
```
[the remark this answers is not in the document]
a-sr: No, it checks for each modal reactor (`state`) if the enclosing modal reactor (`state->parent_mode->state`) is transitioning (`next_mode`) into the mode that contains this modal reactor (`state->parent_mode`).
The test is independent of the current mode and thus not limited to self transitions.

a-sr (changes requested):
I have concerns regarding the way `_lf_sched_mode_change_prologue` duplicates parts of the transition processing. See comment.

core/threaded/scheduler_LET.c
```diff
@@ -516,4 +549,54 @@ void lf_sched_reaction_epilogue(reaction_t * reaction, int worker_number) {
...
\ No newline at end of file
+}
+
+#ifdef MODAL_REACTORS
+static void _lf_sched_mode_change_prologue() {
+    reactor_mode_state_t **states;
+    int states_size = _lf_mode_get_reactor_mode_states(&states);
+
+    // FIXME: In this loop we are actually overwriting the transitions of containing modes
+    //  this is also done later inside the `next` function. I dont think it has any effect.
```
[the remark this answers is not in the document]
erlingrj: Thanks for your insight. You are completely right, it is an ugly hack. I do like your idea of
`_lf_mode_get_transitiong_reactors` I will try to implement this soon.

erlingrj:
EAL:
If I understand your example correct we have mode which has 2 (or more) reactors contained within it. I dont think this would be a problem. Before executing the LET reaction (and any other reaction). The worker threads will acquire the local mutex of the parent. It is not until a worker tries to advance time that they will start locking the local mutexes of any reactor which should do a transition. 

I will make a test case to verify this

core/threaded/scheduler_LET.c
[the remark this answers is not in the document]
erlingrj: Actually Edward is right. For LET purposes we want to stop **any** mode transition in any parent reactor. I dont think we should be allowed to **leave** a mode where we have an executing LET reaction

erlingrj:
Update: I followed @a-sr suggestion and added `_lf_mode_collect_transitioning_reactors` which fills all the reactors which are supposed to have a mode transition into a vector. I opt'ed for the simplest algorithm which can be improved later if necessary. I also added a test case for EAL concern and it shows that having multiple reactors contained in a mode is not a problem

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
     }
     _lf_unsused_suspended_events_head = NULL;
 }
+
+/**
+ * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec 
+ *  for all reactors that will do a transition
+ * 
+ *  A mode does a transition of either:
```
edwardalee, line 565:
```suggestion
 *  A mode performs a transition if either:
```

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
     }
     _lf_unsused_suspended_events_head = NULL;
 }
+
+/**
+ * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec
```
edwardalee, line 562:
```suggestion
 * @brief This function accepts an array of ALL the mode_states and adds a reactor ptr into return_vec 
```
Also, from the implementation, it looks like the array passed in should not be ALL mode states, but rather a list of nested mode states starting with the top-level mode and ending with a leaf mode?

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
     }
     _lf_unsused_suspended_events_head = NULL;
 }
+
+/**
+ * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec 
+ *  for all reactors that will do a transition
```
edwardalee, line 563:
```suggestion
 *  for all reactors that will perform a transition at the current tag.
```

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
...
+ * 
+ *  A mode does a transition of either:
+ *  1) next_mode is set or
+ *  2) parent has next_mode set
+ * 
+ *  A preliminary implementation is simple:
+ *  1) Check if reactor has next_mode set. If so, add to return_vec 
+ *  2) If not, check if its parent has already been added to return_vec. If so, add it
+ *  
+ *  This assumes that the `states` argument is an array ordered hierarchically. It does a bunch of searching through
```
edwardalee, line 573: Ordered in which direction? Shallow first or deep first?

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
...
     _lf_unsused_suspended_events_head = NULL;
 }
+
+/**
+ * @brief This function accepts an array of ALL the mode_states and should add a reactor ptr into return_vec 
+ *  for all reactors that will do a transition
+ * 
+ *  A mode does a transition of either:
+ *  1) next_mode is set or
+ *  2) parent has next_mode set
```
edwardalee, line 567:
```suggestion
 *  2) any parent has next_mode set.
```

core/modal_models/modes.h
```diff
@@ -107,9 +110,9 @@ void _lf_handle_mode_changes(void);
 void _lf_handle_mode_triggered_reactions(void);
 
 /**
- * Function (to be code generated) to get the reactor_mode_state array and its length
+ * Function (to be code generated) to fill up a vector with the reactors who has mode transitions
```
edwardalee, line 113:
```suggestion
 * Function (to be code generated) to fill up a vector with the reactors which have mode transitions.
```
Also, is there no restriction to nested reactors?

core/threaded/scheduler_LET.c
```diff
@@ -336,20 +337,16 @@ void lf_sched_init(
     }
 
     // Allocate array to hold information about what workers are in the workforce
-    worker_is_in_workforce = malloc(_lf_number_of_workers * sizeof(bool));
+    _lf_sched_worker_is_in_workforce = malloc(_lf_number_of_workers * sizeof(bool));
```
edwardalee, line 340: The return value from malloc needs to be cast to bool* or this will fail with the CCpp target.

core/threaded/scheduler_LET.c
```diff
@@ -552,51 +549,30 @@ void lf_sched_reaction_epilogue(reaction_t * reaction, int worker_number) {
 }
 
 #ifdef MODAL_REACTORS
+/**
+ * @brief
```
edwardalee, line 553: Description needed?

core/modal_models/modes.c
```diff
@@ -558,4 +557,51 @@ void _lf_terminate_modal_reactors() {
...
+ *  
+ *  This assumes that the `states` argument is an array ordered hierarchically. It does a bunch of searching through
+ *  the return_vec, but this might be OK.   
+ * 
+ * @param states 
+ * @param states_size 
+ * @param return_vec 
+ * @return int 
+ */
+int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int states_size, void * _return_vec) {
```
edwardalee, line 581: If the return type (for _return_vec) is self_base_t*, then it won't need to be cast when used, maybe? Also, is the leading underscore in _return_vec meaningful?

edwardalee (changes requested):
Looks good! Some suggested changes, mostly in documentation.

core/modal_models/modes.c
erlingrj: Thanks for that info. I am gonna look into that. I thought there was a single array with all mode_states.

core/modal_models/modes.c
erlingrj: Shallow first. Starting with "top-level" mode

core/modal_models/modes.c
erlingrj: Yes you are right. I havent fully grapsed the modal design. But yes. There is not 1 entry per **mode_state,** but rather 1 entry per **mode**

core/modal_models/modes.h
erlingrj: I am not entirely sure what you mean? I believe that reactors contained within the mode also should be added to this vector. I will make a test for it. Will clarify the comment

core/modal_models/modes.c
erlingrj: `_return_vec` because it is void*, when casting it to `vector_t` i name it `return_vec`. Made sense in my head, but open for another way of expressing it

core/modal_models/modes.h
erlingrj: Good catch. I had misunderstood the implementation. It is indeed right that we dont add reactors **contained** in the mode. I added a failing test, will address it.

erlingrj: Is there a way of seeing if a Reactor is local to a specific mode? Worst-case is that we have to lock ALL contained reactors in any reactor which is gonna do a mode change.  Ideally we only lock those that are contained in the mode we are transitioning away from

erlingrj: Actually. Is it even a way of getting the containment hierarchy in the run-time? Might have to code-generate a function for this?

core/threaded/scheduler_LET.c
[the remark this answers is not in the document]
a-sr: Oh yes, in that regard it is correct. I thought it was about the way the transitions where processed.

core/modal_models/modes.h
a-sr: No, at the moment you cannot see if a Reactor is local to a mode. I was a minimalist in introducing pointers for these structures and designed them to be unidirectional and upward in containment, to circumvent having to allocate collections of references.
Reactions are associated with their directly enclosing mode, because these need to suppressed upon inactivity. Reactors only have a reference to an enclosing mode if they are modal reactors, in order to establish a mode hierarchy.
Hence, you can get the enclosing mode via any contained reaction or the mode state; or you need to introduce another pointer if that is not sufficient.

As far as I know the regular containment relations between reactors are not available at runtime, otherwise I would have utilized them for modes, I guess.

erlingrj: Had a brief discussion with @edwardalee yesterday where he proposed that we add the containment hierarchy to the self_struct by adding a `self_base_t *parent` pointer. We then do the following:

1. Before reaction invocation the worker will acquire the local mutex of the reactor but also traverse up the containment hierarchy and acquire the local mutex of any modal reactor up its containment chain. 
2. Before advancing time the worker will acquire the local mutex of any modal reactor which is about to perform a mode change. If this modal reactor has a contained reactor with a LET reaction it will have to wait on that mutex.

There are three features of this design
1. We don't have to "propagate" down the modal hierarchy and lock the contained modal reactors as well **before advancing time**. It will be sufficient to lock the modal reactors which have requested mode change.
2. It will be conservative as a reactor OUTSIDE any mode will also block mode transition. 
3. We can no longer execute LET reactions in independent reactors contained in a modal reactor in parallel. If e.g. a mode has 2 contained reactors, both with LET reactions, both will try to acquire the local mutex of the parent modal reactor. Marten suggested that a semaphore might solve this

erlingrj:
Update:

1. I implemented the locking strategy proposed by Edward. In the reaction prologue, the worker thread will traverse up the containment hierarchy and lock any parent reactor which is modal. There is a FIXME regarding whether I am really checking that condition. In the reaction epilogue the worker thread traverses the hierarchy again and unlocks any modal parent.
2. Before advancing time, the worker thread now only needs to lock any modal reactor which has requested a mode change. I.e. we don't need to go down the nested mode hierarchy and lock the nested Reactors also. The LET worker will have locked the uppermost Reactor.
3. Deadline violation handling is now moved to AFTER this locking of modal parent reactors. There is now a real chance of blocking when acquiring those through `_lf_sched_lock_modal_parents` and thus the deadline violation must be handled AFTER this. The STP violations are still checked before. I am not 100% of the details of STP, but those should not be affected by additional delaying?
4. There are corresponding changes to LFC to code-generate pointer to parent reactor etc.

edwardalee:
> The LET worker will have locked the uppermost Reactor.

Locking the uppermost reactor will have an unfortunate cost for federated execution, which is that no messages will be received from the network during the execution of the LET reactor.  Why is this necessary (assuming the uppermost reactor is not modal)?

erlingrj:
> > The LET worker will have locked the uppermost Reactor.
> 
> Locking the uppermost reactor will have an unfortunate cost for federated execution, which is that no messages will be received from the network during the execution of the LET reactor. Why is this necessary (assuming the uppermost reactor is not modal)?

I edited the original post to be more clear. We only lock parent reactors that are modal. So it is the uppermost modal reactor.

core/threaded/scheduler_LET.c
```diff
@@ -456,29 +487,92 @@ void lf_sched_trigger_reaction(reaction_t* reaction, int worker_number) {
...
+ *  up the containment hierarchy and locks any modal reactor
+ * 
+ * @param reactor 
+ */
+static void _lf_sched_lock_modal_parents(self_base_t* reactor, int worker_number) {
+    self_base_t *parent = reactor->parent;
+    // If has a parent/containing reactor
+    if (parent) {
+        // Check if it is modal
+        if (parent->_lf__mode_state.current_mode) { // FIXME: Is this a reliable way to check if reactor is modal?
```
erlingrj, line 503: @a-sr Is there a clearer way of figuring out if a Reactor is modal or not?

edwardalee: This looks like a reasonable way to me. An alternative is to check whether ASTUtils.allModes(reactor) returns an empty list, but I don't think you need to change this.

core/reactor_common.c
```diff
@@ -1571,20 +1571,39 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
+            // Deadline violation has occurred.
+            violation = true;
+            // Invoke the local handler, if there is one.
+            reaction_function_t handler = reaction->deadline_violation_handler;
+            if (handler != NULL) {
+                (*handler)(reaction->self);
+            }
+        }
+    }
+
```
erlingrj, line 1593: @edwardalee This is rather big change that was necessary. We have to check for deadline violation in the very end. Because we might stall in `_lf_sched_reaction_prologue` waiting for a mutex.

core/threaded/reactor_threaded.c
```diff
@@ -1023,7 +1023,9 @@ void _lf_worker_do_work(int worker_number) {
                 current_reaction_to_execute->chain_id,
                 current_reaction_to_execute->deadline);
 
-        bool violation = _lf_worker_handle_violations(
+        // Check for STP violations. Deadline violations checked after acquiring all the
+        //  mutexes.
```
erlingrj, line 1027: Here I remove the check for deadline violation and only check for STP. I believe STP violations are unaffected by the prospect of stalling for an arbitrary period of time between the check and reaction invokation

core/threaded/scheduler_LET.c
```diff
@@ -29,6 +29,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define NUMBER_OF_WORKERS 1
 #endif  // NUMBER_OF_WORKERS
 
+#ifndef LF_MULTI_THREADED
+#error LET-scheduler was chosen without LF_MULTI_THREADED flag defined
+#endif
```
erlingrj, line 34: LET scheduler without multi-threading is pointless.

erlingrj:
Note that this is a PR into another draft PR, so it is not too big deal to merge it. But we of course wanna be more or less happy with the design

core/threaded/scheduler_LET.c
a-sr: I agree, this is a good way to check for a modal reactor.

core/reactor_common.c
```diff
@@ -1571,20 +1571,39 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
-    ((self_base_t*) reaction->self)->current_tag = current_tag;
-    ((self_base_t*) reaction->self)->executing_reaction = reaction;
-    reaction->function(reaction->self);
-    ((self_base_t*) reaction->self)->executing_reaction = NULL;
-    tracepoint_reaction_ends(reaction, worker); 
-    LF_PRINT_DEBUG("Worker %d Finished Reaction", worker);
+    #endif        
+
+    bool violation = false;
+    if (reaction->deadline >= 0LL) {
```
edwardalee, line 1579: Should this be:
```suggestion
    if (reaction->deadline >= 0LL && reaction->deadline < FOREVER) {
```

core/threaded/scheduler_LET.c
```diff
@@ -342,7 +370,8 @@ void lf_sched_free() {
     free(_lf_sched_instance->_lf_sched_triggered_reactions);
     free(_lf_sched_instance->_lf_sched_executing_reactions);
     lf_semaphore_destroy(_lf_sched_instance->_lf_sched_semaphore);
-    free(worker_is_in_workforce);
+    free(_lf_sched_worker_is_in_workforce);
+    vector_free(&_lf_sched_transitioning_reactors);
```
edwardalee, line 374: Shouldn't this free be guarded by `ifdef MODAL_REACTORS`?

core/threaded/scheduler_LET.c
```diff
@@ -515,5 +609,42 @@ void lf_sched_reaction_epilogue(reaction_t * reaction, int worker_number) {
...
+#ifdef MODAL_REACTORS
+/**
+ * @brief This function should be invoked by the worker thread about to advance time BEFORE it acquires the global mutex.
+ *  It is only relevant if we have modal reactors. It gathers all the Reactors which are about to perform mode changes, and reactors contained within them,
+ *  and acquires there local mutex to make sure there are no LET reactions currently executing in any of them.
+ *  In the case that there are LET reactions, this will block the advancement of time until the LET reaction completes.
+ * 
+ */
+static void _lf_sched_mode_change_prologue() {
+    int n_reactors = _lf_mode_get_transitioning_reactors((void *) &_lf_sched_transitioning_reactors);
```
edwardalee, line 629: Does this need to called every time the prologue is invoked? Does this change during execution? If so, why is it not dependent on the hierarchy? Also, won't this just keep adding to the vector argument?

core/threaded/scheduler_LET.c
```diff
@@ -515,5 +609,42 @@ void lf_sched_reaction_epilogue(reaction_t * reaction, int worker_number) {
...
\ No newline at end of file
+}
+
+#ifdef MODAL_REACTORS
+/**
+ * @brief This function should be invoked by the worker thread about to advance time BEFORE it acquires the global mutex.
+ *  It is only relevant if we have modal reactors. It gathers all the Reactors which are about to perform mode changes, and reactors contained within them,
+ *  and acquires there local mutex to make sure there are no LET reactions currently executing in any of them.
+ *  In the case that there are LET reactions, this will block the advancement of time until the LET reaction completes.
+ *
```
edwardalee, line 626:
```suggestion
 * @brief Acquire the local mutex for any reactor that has pending mode changes.
 * This function should be invoked by the worker thread about to advance time BEFORE it acquires the global mutex.
 *  It is only relevant if we have modal reactors. 
 * It gathers all the Reactors which are about to perform mode changes, and reactors contained within them,
 *  and acquires their local mutex to make sure there are no LET reactions currently executing in any of them.
 *  In the case that there are LET reactions, this will block the advancement of time until the LET reaction completes.
```

core/modal_models/modes.h
```diff
@@ -90,18 +109,11 @@ void _lf_handle_mode_changes(void);
...
-//// Type definitions for modal infrastructure.
-
-/** Typedef for reactor_mode_t struct, used for representing a mode. */
-typedef struct reactor_mode_t reactor_mode_t;
-/** Typedef for reactor_mode_state_t struct, used for storing modal state of reactor and/or its relation to enclosing modes. */
-typedef struct reactor_mode_state_t reactor_mode_state_t;
-/** Typedef for mode_state_variable_reset_data_t struct, used for storing data for resetting state variables nested in modes. */
-typedef struct mode_state_variable_reset_data_t mode_state_variable_reset_data_t;
+/**
+ * Function (to be code generated) to fill up a vector with the reactors which have mode transitions.
```
edwardalee, line 113: More information is needed here.  From the usage of this function, I infer that a new vector will be allocated (which never gets freed?) and that it only adds reactors which have pending mode transitions, not all reactors with modes. Is this correct? Please clarify and explain how the argument should be used (if a new vector is allocated, it must later be freed).

edwardalee (approved):
LGTM, modulo comments.

core/reactor_common.c
erlingrj: You are right!

core/modal_models/modes.c
```diff
@@ -557,4 +557,31 @@ void _lf_terminate_modal_reactors() {
...
+ *  contained modes. Only reactors that have scheduled a transition itself.
+ * 
+ * @param states 
+ * @param states_size 
+ * @param return_vec 
+ * @return int 
+ */
+int _lf_mode_collect_transitioning_reactors(reactor_mode_state_t **states, int states_size, void * _return_vec) {
+    int num_transitioning_reactors = 0;
+    vector_t *return_vec = (vector_t *) _return_vec;
```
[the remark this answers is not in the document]
erlingrj: This function is called from code-generated functions. My thinking was to avoid including vector.h in the code-generated file to keep it out of the namespace/scope of the user reactions.

core/modal_models/modes.h
erlingrj: Have clarified. Vector is allocated and freed inside `scheduler_LET.c`.

core/modal_models/modes.h
```diff
@@ -90,18 +109,11 @@ void _lf_handle_mode_changes(void);
...
+ */
+int _lf_mode_get_transitioning_reactors(void * return_vec);
```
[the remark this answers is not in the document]
erlingrj: Might be. See my comment above about keeping `vector.h` out of the codegenerated namespace in which the user-code also lives

core/threaded/scheduler_LET.c
erlingrj: Yes, you are right. Thanks!

core/threaded/scheduler_LET.c
erlingrj: This has not related to the reaction invocation prologue. This is the prologue to time advancement. So it is called each time time is about to be advanced. In the epilogue all the transitioning reactors are popped of the vector and unlocked.

I have renamed it to `_lf_sched_mode_time_advance_prologue` Which is verbose but perhaps clearer?

erlingrj (commented):
Only need to clear up whether we should use void* to pass the vector_t* through the code-generated code, or expose vector.h to user-code =

edwardalee:
> Only need to clear up whether we should use void* to pass the vector_t* through the code-generated code, or expose vector.h to user-code =

OK, I see the reason for the hesitation.  At a minimum, the data type needs to be documented clearly.

PR #124, opened 2022-11-03
Token payload freed with after delay

edwardalee:
This PR fixes a logic error where the payload of a token was not being freed when the token itself was not freed. A token is not freed if it is the "template" token for a port or action because it contains type information for that port or action and it will be reused in future events for that port or action. However, the payload, the value field, should be freed when it is no longer needed. This was not being done. This bug was probably introduced with the Python target, where payloads never get freed because the payload is a Python target and Python handles garbage collection.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -55,7 +55,11 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 /**
  * The flag OK_TO_FREE is used to indicate whether
- * the void* in token_t should be freed or not.
+ * the void* value in token_t should be freed or not.
+ * In particular, in the Python target, the value should
+ * not be freed because Python handles garbage collection.
+ * But the token will still need to be freed (unless it's a
+ * template token).
  */
 #ifdef _LF_GARBAGE_COLLECTED
 #define OK_TO_FREE token_only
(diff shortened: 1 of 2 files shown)


erlingrj:
Valgrind still says that 64B was definitely lost:
image

Compared to when running without the after-delay:
image


I am really a novice with valgrind so it might not be a problem

erlingrj:
```
target C {
    build-type: debug
};

reactor Source {
    output out:int
    timer t(0,10 msec)
    reaction(t) -> out {=
        lf_set(out,1);
    =}
}

reactor Sink {
    input in:int
    reaction(in) {==}
}

main reactor {
    a = new Source()
    b = new Sink()
    a.out->b.in after 10 msec
}
```

edwardalee:
I think this means there is still a problem. Do the numbers change with the length of the run?

erlingrj:
No. it seems static. I will try to poke a little more at it and update here

erlingrj:
image

More info, it seems as the token created never gets freed again. If it keeps reusing the token, then somehow it lacks a final cleanup upon termination. I added a timeout to the program.

core/reactor_common.c
```diff
@@ -358,10 +362,18 @@ static token_freed _lf_free_token(lf_token_t* token) {
...
         _lf_count_payload_allocations--;
-        if (OK_TO_FREE != token_only && token->ok_to_free == token_and_value) {
-            LF_PRINT_DEBUG("_lf_free_token: Freeing allocated memory for payload (token value): %p", token->value);
+        // Free the value field (the payload).
+        // First check whether the value field is garbage collected (e.g. in the
+        // Python target), in which case the payload should not be freed.
+        if (OK_TO_FREE != token_only
+                && token->value != NULL
+                && token->ok_to_free != token_only
+            ) {
```
erlingrj, line 374: Is there a difference between `token->ok_to_free == token_and_value` and `!= token_only`?

erlingrj (commented):
As far as i could tell the only actual change is in the `_lf_free_token` where the condition for freeing the value is changed. I don't fully understand how it changes the behaviour though.

edwardalee:
The difference is that the payload gets freed when token->ok_to_free == no.

erlingrj (approved):
Great!

edwardalee:
There is still more work to do to get a clean valgrind report. Specifically, there is memory that is allocated at initialization time that never gets freed. I will keep working in this branch to attempt to address these issues, even though these are less critical and will only become a real issue when mutations are supported.

PR #126, opened 2022-11-12
Fixed hashmap and use it for recycling bin and final cleanup

edwardalee:
This PR replaces the token recycling bin with a hashmap rather than a priority queue as a guard against a token being inserted more than once. It also frees memory at termination on the recycling bin and memory used as token templates in ports and actions, which should eliminate some of valgrind's reported memory loss.

PR #127, opened 2022-11-15
avoid duplicate tokens in token recycle bin

OmerMajNition:
This issue was reproduced when using pointers to struct for input and output of reactions. After a short run (when size of recycle bin overflows by duplicate entries into recycle bin) occurs a double delete.

Trigger still holds the pointer to token so the actual issue of double delete would still be there. Complete fix of this problem needs a thorough review of token usage, and I would try to create a PR with changes in the future.

Currently this fix would do some good for shorter programs that don't overflow recycle bin, or we could increase the recycle bin limit as a workaround in that case.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -389,10 +389,19 @@ static token_freed _lf_free_token(lf_token_t* token) {
     if (token->ok_to_free) {
         // Need to free the lf_token_t struct also.
         if (_lf_token_recycling_bin_size < _LF_TOKEN_RECYCLING_BIN_SIZE_LIMIT) {
-            // Recycle instead of freeing.
-            token->next_free = _lf_token_recycling_bin;
-            _lf_token_recycling_bin = token;
-            _lf_token_recycling_bin_size++;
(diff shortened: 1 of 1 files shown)


edwardalee (commented):
Thanks for pointing out this bug. However, PR #827 has a more efficient solution to this problem, which is to use a hashset for the recycle bin. It also includes a number of other improvements to the token handling.  Can we wait for that PR to be completed and merged? There is still some tuning going on to get a clean memory report from valgrind.

OmerMajNition:
> Thanks for pointing out this bug. However, PR #827 has a more efficient solution to this problem, which is to use a hashset for the recycle bin. It also includes a number of other improvements to the token handling. Can we wait for that PR to be completed and merged? There is still some tuning going on to get a clean memory report from valgrind.

OmerMajNition:
Totally makes sense. I've closed this PR, and I've put some comments on original PR [https://github.com/lf-lang/reactor-c/pull/126](url) targeting the fix.
Please review the comments.

PR #131, opened 2022-12-08
Continuations

petervdonovan:
Requirements:
- [x] Continue a reaction where it left off when it is re-triggered in a later tag.
- [ ] Add a mechanism to send multiple requests in the same tag, and only (spiritually) "block" on the response ports when a response is needed (like futures)
- [ ] Create a test in which a caller reaction processes a response and produces a new request in the same tag.
- [ ] Create a test in which a caller reaction processes multiple different responses in the same tag
- [ ] Create a test in which a request is distributed by an intermediate handler to several workers, which each send back partial responses that are merged and send back to the caller
- [ ] Create a test in which a caller reaction receives and successfully processes several responses at the same tag.
- [ ] Create a test in which a caller reaction sends multiple requests and receives them out of order. (This is only possible with the "futures" construct.)
- [ ] Implement a few examples using continuations to show that they make code more readable (and more like programming in a conventional GPL)
- [ ] Do not have a memory leak when no request comes back. (**What should be the mechanism for this?**)
- [ ] Provide a way to introspect into pending requests, e.g. so that decisions can be made according to their number. (**What other data needs to be exposed?**)
- [ ] Provide a way to modify pending requests? (**If you wish to have this amount of power, why not use LF's lower-level primitives? Is it a matter of convenient programming (which this proposal can solve), or of reliable program analysis (which this proposal does not solve at all)?**)

Low-level control over pending requests will probably require another level of indirection. Probably we will need to attach a hashmap to the self struct of the requestor reactor from request IDs to request metadata, or something like that.

Questions:
- Should we allow a reaction to receive a response in the same tag as when it produces the request? I doubt that we can do this in general without introducing a mechanism that is equivalent to microsteps. The handling of a request can, in general, require an iteration through the entire reaction graph. Why would we implement a mechanism that allows arbitrarily many iterations through the reaction graph when we already have a mechanism to do that in zero time (microsteps)? The underlying issue I have that causes me to raise this point is that I do not understand what actual "consistency" is lost by using multiple microsteps versus multiple levels; this lack of understanding has come up before in our discussion of what exactly would be lost if we forbade federated execution with zero-delay cycles between federates, and it's important because we claim that it distinguishes us from actor-based models that totally order events that are processed and impose rules wrt how each event can be processed.

### DIFF
--- a/core/utils/CMakeLists.txt
+++ b/core/utils/CMakeLists.txt
@@ -1,3 +1,3 @@
-set(GENERAL_SOURCES vector.c pqueue.c util.c)
+set(GENERAL_SOURCES vector.c pqueue.c util.c context_switch.c)
 set(MULTITHREADED_SOURCES semaphore.c)
 add_sources_to_parent(GENERAL_SOURCES MULTITHREADED_SOURCES "")
--- a/core/utils/context_switch.c
+++ b/core/utils/context_switch.c
@@ -1,0 +1,44 @@
+#include <stddef.h>
+#include <stdlib.h>
+#include <setjmp.h>
+
+typedef struct context_t {
+    jmp_buf registers;
+    int n;
+    long *stack;
+} context_t;
+
+static void save_stack(context_t *c, long *top_of_stack, long *bottom_of_stack) {
+    int n = top_of_stack-bottom_of_stack;
+    c->stack = (long *) malloc(n*sizeof(long));
+    c->n = n;
+    for (int i = 0; i < n; i++) {
+        c->stack[i] = top_of_stack[-i];
+    }
+}
+
+context_t* context_save(long* top_of_stack) {
+    context_t* ret = (context_t*) malloc(sizeof(context_t));
+    long bottom_of_stack;
+    if (!setjmp(ret->registers)) {
+        save_stack(ret, top_of_stack, &bottom_of_stack);
+        return ret;
+    } else {
+        return NULL; // setjmp was returned to by a longjmp call
+    }
+}
+
+void context_switch(context_t* c, long* top_of_stack) {
+    long mem_allocated_on_stack[12];
+    long current_bottom_of_stack;
+    if (top_of_stack - &current_bottom_of_stack < c->n) context_switch(c, top_of_stack);
+    for (int i = 0; i < c->n; i++) {
+        top_of_stack[-i] = c->stack[i];
+    }
+    longjmp(c->registers, 1);
+}
+
+void context_free(context_t* ctx) {
+    free(ctx->stack);
+    free(ctx);
+}
--- a/include/api/set.h
+++ b/include/api/set.h
@@ -29,10 +29,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * Target-specific runtime functions for the C target language.
  * This API layer can be used in conjunction with:
  *     target C;
- * 
- * Note for target language developers. This is one way of developing a target language where 
- * the C core runtime is adopted. This file is a translation layer that implements Lingua Franca 
- * APIs which interact with the internal _lf_SET and _lf_schedule APIs. This file can act as a 
+ *
+ * Note for target language developers. This is one way of developing a target language where
+ * the C core runtime is adopted. This file is a translation layer that implements Lingua Franca
+ * APIs which interact with the internal _lf_SET and _lf_schedule APIs. This file can act as a
  * template for future runtime developement for target languages.
  * For source generation, see xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CCppGenerator.xtend.
  */
@@ -54,20 +54,20 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * to the specified value.
  *
  * If the value argument is a primitive type such as int,
- * double, etc. as well as the built-in types bool and string, 
+ * double, etc. as well as the built-in types bool and string,
  * the value is copied and therefore the variable carrying the
  * value can be subsequently modified without changing the output.
  * This also applies to structs with a type defined by a typedef
  * so that the type designating string does not end in '*'.
- * 
+ *
  * If the value argument is a pointer
  * to memory that the calling reaction has dynamically allocated,
  * the memory will be automatically freed once all downstream
  * reactions no longer need the value.
  * If 'lf_set_destructor' is called on 'out', then that destructor
- * will be used to free 'value'. 
+ * will be used to free 'value'.
  * Otherwise, the default void free(void*) function is used.
- * 
+ *
  * @param out The output port (by name) or input of a contained
  *  reactor in form input_name.port_name.
  * @param value The value to insert into the self struct.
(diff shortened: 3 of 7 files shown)


petervdonovan:
More notes from talking to Omer:
- User-level mechanisms for preventing the memory leak could include:
    - Attaching a timeout to a request. If the request is not handled in time.
    - Setting a maximum allowable number of pending requests.
- If a response comes back, but the memory where the context was stored has been freed, the response can just be ignored in the use cases we have currently.
- Information about pending requests that should be accessible to the user can include
    - The number of pending requests. This might be sufficient for most use cases that we have currently.
- The lack of convenient syntax is an issue. Not knowing whether a request will come back might be an issue later, but it will involve analyzing C code. We can probably take this step by step, and deal with that issue later?
- A good implementation might be to keep some simple data structure, like a linked list or a hash map, on the heap to keep track of the pending requests.

petervdonovan:
There's another problem... Silly of me not to think of it before.

The stack pointer `%esp` gets restored when we `longjmp` to the saved context, but the top of the stack is not necessarily in the same place as it was before because since `%esp` was saved to the stack, the stack shrank back a few frames before re-growing, and depending on control flow it might not have re-grown to the exact same length, even when calling the same function as before with basically the same backtrace. I have confirmed by viewing the value of `%esp` in GDB that this is what is happening, and I think it's probably the reason for the stack smashing and segfault errors that I have been encountering. I'm still working on a fix.

petervdonovan:
# Notes from conversation today with @lhstrh

Please feel free to edit or reply with corrections -- this is my current understanding of what we discussed.

## Topics
* The problem with a single microstep dimension
  - motivation
  - mixing old and new data
  - composability
* Semantics of sync and async request/response in TypeScript and in C
* Implementation of sync and async request/response in TypeScript
* Implementation of async request/response calls in C

## The problem with a single microstep dimension

### Motivation

The premise (which we have debated) is that the implementation of synchronous request/response cannot merely be a special case of the implementation of the asynchronous case, and that it will be more work to support both than to just support the asynchronous case.

If that premise is true, then if we support the synchronous case, we must have a reason to do so.

Here is one possible reason: It is problematic to have a single microstep dimension that is used by everything. The outside environment of the subprogram that uses the request/response pattern will evolve in the microstep dimension, but the subprogram that uses the pattern will also evolve in the same microstep dimension. The synchronous request/response pattern would have no benefit over the asynchronous version if it were possible to abstract away the fact that the asynchronous subprogram takes at least a microstep to complete, during which time the surrounding reactors may also evolve without pause.

### Mixing old and new data

One reason to avoid using microsteps is to avoid mixing old data with new. In particular, if a reaction sends an async request, then when the same reaction resumes (at a later logical time), it is undesirable for it to have access to updated state. But if we do not do copying, then it must have access to updated state, both on its input ports and in its state variables.

I argued that this was related to the problem we faced in federated execution, where we wanted to allow zero-delay federate-level cycles because we wanted to get all federates to, at a given logical time, all agree on the value of some variable. Marten pointed out that this is a separate problem, but in any case everyone agrees that it can sometimes be undesirable to allow old data to be accidentally combined with new data.

### Composability

In a combinational subprogram written in Lingua Franca, it is possible to abstract away how many reactions are involved. If a user instantiates a purely combinational reactor, all they will see is that once its inputs are provided, its outputs will appear after precisely zero time (or not appear at all).

This is not the case for sequential subprograms that use time or even microsteps. The number of microsteps that a computation takes is not an implementation detail; it affects precedence and simultaneity.

### Three solutions

#### Solution 1: Reset to zero

As Andres mentioned today, if one uses the "reset microsteps to zero" semantics for logical actions, it is possible to circumvent this problem by delaying the output of a computation that takes uncertain microsteps by a nonzero logical time. This will erase the uncertainty in the microstep dimension by resetting the microstep to zero.

Although this helps, it only pushes the problem one step backward. The resulting reactor now takes time in the logical time dimension, not just microsteps, and so if it, in turn, is also used in a composition, then we will not be able to use the same trick again.

#### Solution 2: Alignment with after delays

The other solution is not to abstract away the lack of simultaneity at all. If one knows that a computation will always take 37 (logical) nanoseconds, and one wishes its output to be simultaneous with another result that is available immediately, then simply delay the other result by 37 nanoseconds so that they line up.

The appeal of this approach is that in some cases, the user has other reasons to want to know exactly how program execution will unfold across logical time, in which case they might as well use that knowledge. Furthermore, it should be possible to automatically analyze some programs to determine how long they take to produce outputs.

#### Solution 3: ABO reactors

This solution is the simplest and the most general. Create a reactor that accumulates its inputs and then, when they all are present, produces them all as simultaneous outputs.

#### Discussion

It is a little troubling to have three ways of solving essentially the same problem (there is a reason why the Zen of Python forbids such things).

It is tempting to argue that solution 3 should always be used while the other two are discouraged, since solution 3 is the only one that can cover all cases. However, it has the following problems:
* It reflects the least knowledge about the program. Therefore, it might be easier to write programs that usually work using approach 3 but not always. What if one output appears twice before the other output appears once? By contrast, using the second approach forces the programmer to determine exactly what their program will do.
* It is not efficient because it involves copying. This may not turn out to be a valid reason because after delays also involve copying, so approach 3 is probably not less efficient than approach 2 or perhaps even approach 1.

In any case, everyone agrees that the downsides of all three of these "solutions" would be at least somewhat mitigated by making things combinational whenever possible. This justifies at least some interest in implementing a syncronous request/response pattern.

## Semantics of sync and async request/response in TypeScript and in C

In my discussion with Marten, it seemed that the semantics of the asynchronous request/response pattern that I prototyped in C using stack-copying probably had very few open questions. This is because the asynchronous version is a fairly thin layer over the existing message-passing that we already implement.

Currently, the semantics of a synchronous version are also possible to explain in a fairly simple way. The caller reaction invokes a callee reaction in another reactor, and the callee reaction is not a regular reaction. It is executed basically just like a part of the caller reaction, except that it has access to state in another reactor.

However, its access to state in another reactor means that additional dependency relationships are implied on the caller reaction so that two caller reactions do not arbitrarily interleave accesses to the same reactor. It is this extra set of dependency relationships that makes the synchronous request/response pattern impossible to implement without a nontrivial (and obviously user-visible) extension of LF semantics. Marten has the technical details of these dependency relationships written down somewhere, but even if they are already figured out, they will take work to implement. Furthermore, they would have to be implemented in Java, not just TypeScript, to hook them into the diagrams and the language server.

## Implementation of async and sync request/response in TypeScript

### Async calls

We discussed whether async calls should be implemented using callbacks or using async/await. We decided that they should use async/await for the following reason:

Suppose that the async calls are made inside a loop using callbacks so that many of them are forked off. The semantics here would have to be that only the inside of the loop is somehow wrapped inside the callback, because it would not work to have part of the loop outside the callback, but the later iterations of the loop inside the callback. This is possible to imagine, but without async/await, JavaScript syntax does not allow it. So if one chooses this implementation and is thus forced into this semantics (where many callbacks may be forked off), then one must determine the order in which the callbacks are executed (since they have access to the same closure). They could just be executed in the order in which they were created, so this question is answerable. However, for some reason we did not like this idea of using callbacks, perhaps because async/await semantics are so convenient in JavaScript.

Async/await and the semantics that it is friendly toward are in line with the path that I have taken in the prototype C implementation. It does have the downside that it is unfriendly toward the spawning of several requests in parallel, but one can work around this by making one request and fanning it out externally.

### Sync calls

The current TypeScript implementation uses regular function calls as synchronous calls. However, we decided that if synchronous calls are implemented in the TypeScript target, they should also use async/await. This is because the same reactor class should be able to make calls, regardless of whether the connections outside it cause the calls to be synchronous or asynchronous.

## Implementation of async request/response calls in C

In principle, both synchronous and asynchronous request/response calls should be possible in C, but because it seems like implementing both would be more work than implementing just asynchronous calls, it seems prudent to worry about asynchonous calls first.

There are two options on the table for implementation of asynchronous request/response calls in C:
1. Save the registers and top few stack frames on the heap and restore them later. In the matrix multiplication example, the amount of stack that must be saved and restored is on the order of 28 quadwords I believe.
2. Save all variables in scope on the heap and restore them later. This requires parsing the reaction bodies using an existing C parser to find out what variables are in scope at the point where the call occurs.

I like to describe option 1 as the hack and option 2 as "the right way," but that does not mean I think option 2 is better. Let me explain: Option 1 relies on details that are not documented in the formal specification of the C language. Furthermore, there is no guarantee that the stack will be copied into exactly the same region of memory, which means that when the registers are restored, the stack pointer and stack base pointer %rsp and %rbp must be manually updated. GCC explicitly forbids inline assembly from modifying those registers without restoring them to their original state. However, if option 1 turns out to be easier to implement, and if all implementations of C that we wish to target are compatible with option 1, then who knows, maybe violating the spec and relying on undefined behavior is a good idea.

Benefits of option 2, besides not involving hacky low-level magic, include:
* Shaokai already parses reaction bodies anyway, so it might not actually add much extra work.
* If we parse reaction bodies, then we can allow them to be a slight extension of C that might be more ergonomic than the macro-based API that is currently in use. So again, it is something that we might like to do anyway.
* The amount of copying might be reduced since the other approach copies register contents that may just be garbage.

petervdonovan:
Closing this for now. It does not seem to solve the most pressing problem that the cache simulator is facing, and the current implementation probably cannot be made to work properly for general programs.

PR #134, opened 2023-01-05
Zephyr support

erlingrj:
- Adds support for threaded and unthreaded LF programs with Zephyr as platform. Only unthreaded is properly tested yet. I could throw an error/warning if threaded is used until I have tested it properly
- Aso adds RUNTIME_CHECKS macro to hide some runtime checks that arent strictly necessary it is enabled by default.
- The program returns with exit(0) instead of return 0. If not the Zephyr runtime doesn't jump to the termination function. It might be several places in the code where the program could terminate where we should replace `return 0` with `exit(0)`


core/platform/Platform.cmake
```diff
@@ -1,4 +1,5 @@
 # Check which system we are running on to select the correct platform support
+
```
erlingrj, line 2:
```suggestion
```

core/platform/lf_zephyr_support.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <stdint.h> // For fixed-width integral types
+#include <time.h>   // For CLOCK_MONOTONIC
+#include <stdbool.h>
+#include <stdlib.h> //malloc, calloc, free, realloc
+
+#include <zephyr/kernel.h>
+
+// FIXME: This flag should also be set by the compiler
+#define TARGET_EMBEDDED
```
erlingrj, line 12:
```suggestion
#define NO_TTY
```

core/platform/lf_zephyr_support.h
```diff
@@ -0,0 +1,53 @@
...
+ * Interval of time.
+ */
+typedef int64_t _interval_t;
+
+/**
+ * Microstep instant.
+ */
+typedef uint32_t _microstep_t;
+
+#ifdef NUMBER_OF_WORKERS
```
erlingrj, line 34:
```suggestion
#ifdef LF_THREADED
```

core/platform/lf_zephyr_support.h
```diff
@@ -0,0 +1,53 @@
...
+extern _lf_mutex_t mutex;
+extern _lf_cond_t event_q_changed;
+
+// Atomics
+int _zephyr_atomic_fetch_add(int *ptr, int value);
+int _zephyr_atomic_add_fetch(int *ptr, int value);
+bool _zephyr_bool_compare_and_swap(bool *ptr, bool value, bool newval);
+bool _zephyr_val_compare_and_swap(int *ptr, int value, int newval);
+
+#endif
```
erlingrj, line 48:
```suggestion
#endif // LF_THREADED
```

core/reactor.c
```diff
@@ -321,9 +325,11 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
             && process_args(argc, argv)) {
         LF_PRINT_DEBUG("Processed command line arguments.");
         LF_PRINT_DEBUG("Registering the termination function.");
+        #ifndef TARGET_EMBEDDED
```
erlingrj, line 328:
```suggestion
```

core/reactor.c
```diff
@@ -321,9 +325,11 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
...
         if (atexit(termination) != 0) {
             lf_print_warning("Failed to register termination function!");
         }
+        #endif
```
erlingrj, line 332:
```suggestion
```

core/reactor.c
```diff
@@ -131,10 +133,11 @@ void _lf_trigger_reaction(reaction_t* reaction, int worker_number) {
  *  should stop.
  */
 int _lf_do_step(void) {
+    reaction_t* reaction;
+
     // Invoke reactions.
-    while(pqueue_size(reaction_q) > 0) {
+    while(reaction = (reaction_t*)pqueue_pop(reaction_q)) {
```
erlingrj, line 139:
```suggestion
    while(pqueue_size(reaction_q) > 0) {
```

core/reactor.c
```diff
@@ -131,10 +133,11 @@ void _lf_trigger_reaction(reaction_t* reaction, int worker_number) {
...
         // lf_print_snapshot();
```
erlingrj, line 140:
```suggestion
        // lf_print_snapshot();
        reaction_t* reaction = (reaction_t*)pqueue_pop(reaction_q);
```

erlingrj:
Note to self:
1. Merge cmake-improvements to main
2. merge main to this
3. remove hard-coded compile-defs from LF
4. Run regression tests on qemu
5. Look into threaded (or disable the option)
6. Look into to `prj.conf` for multiple boards such that counters work as expected

core/reactor.c
```diff
@@ -266,6 +274,7 @@ int next(void) {
     // Advance current time to match that of the first event on the queue.
     // We can now leave the critical section. Any events that will be added
     // to the queue asynchronously will have a later tag than the current one.
+    // FIXME: use tag or instant here?
```
erlingrj, line 277:
```suggestion
```

core/reactor.c
```diff
@@ -33,14 +33,20 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * @author{Soroush Bateni <soroush@utdallas.edu>}
  * @author{Erling Jellum <erlingrj@berkeley.edu>}
  */
-#include <signal.h> // To trap ctrl-c and invoke termination().
+
+
```
erlingrj, line 37:
```suggestion
```

core/reactor_common.c
```diff
@@ -1893,7 +1909,9 @@ const char** default_argv = NULL;
  * Process the command-line arguments. If the command line arguments are not
  * understood, then print a usage message and return 0. Otherwise, return 1.
  * @return 1 if the arguments processed successfully, 0 otherwise.
+ * TODO: Not necessary for NO_TTY
  */
+
```
erlingrj, line 1914:
```suggestion
```

arengarajan99 (approved):

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,527 @@
...
+    bool res = false;
+    if (*ptr == value) {
+        *ptr = newval;
+        res = true;
+    }
+    lf_critical_section_exit();
+    return res;
+}
+
+// FIXME: Are you sure that it should return bool?
```
petervdonovan, line 515: When these two functions were added to our platform files, they were based directly on [GCC's atomic builtins](https://gcc.gnu.org/onlinedocs/gcc-4.1.0/gcc/Atomic-Builtins.html), in particular `__sync_bool_compare_and_swap` and `__sync_val_compare_and_swap`. Unlike `__sync_bool_compare_and_swap`, `__sync_val_compare_and_swap` does not necessarily return a `bool`, so yes, it does seem like something is likely not right here.

core/reactor.c
```diff
@@ -194,7 +200,7 @@ int _lf_do_step(void) {
 #endif
 
     if (lf_tag_compare(current_tag, stop_tag) >= 0) {
-        return 0;
+        exit(0);
```
petervdonovan, line 203: What is the reason for this change? It looks like the code should do the same thing either way, but the doc comments do not say anything about returning with 0 exit code directly out of this function.

core/reactor_common.c
```diff
@@ -1656,13 +1666,18 @@ void _lf_invoke_reaction(reaction_t* reaction, int worker) {
...
+#ifdef FEDERATED
     if (reaction->is_a_control_reaction) {
         // Control reactions will not produce an output but can have
         // effects in order to have certain precedence requirements.
         // No need to execute this function if the reaction is a control
         // reaction.
         return;
     }
+#endif
+    if (reaction->num_outputs == 0) {
```
petervdonovan, line 1678: This special case looks like it should be unnecessary. A reaction with no outputs a) has no outputs, so the for loop does nothing and b) does not set "downstream to execute now," so the `if` clause at the end should also do nothing -- right?

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        _lf_alarm_cfg.ticks = sleep_duration_ticks;
+        int err = counter_set_channel_alarm(_lf_counter_dev, LF_TIMER_ALARM_CHANNEL,  &_lf_alarm_cfg);
+     
+        if (err != 0) {
+            lf_print_error_and_exit("Could not setup alarm for sleeping. Errno %i", err);
+        }
+        
+        lf_critical_section_exit();
```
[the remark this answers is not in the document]
erlingrj: Good feedback. For now I will document it. I think we need to decide also what part of the platform.h API should be available to the user. It seems it is a convention to prepend functions that are internal with _. We might wanna do that for this function. ALso we might want a `lf_sleep_until` that doesn't require a critical section and the users can call from reaction bodies

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,527 @@
...
+ * @brief Enter critical section by disabling interrupts.
+ * Support nested critical sections by only disabling
+ * interrupts on the first call 
+ * 
+ * @return int 
+ */
+int lf_critical_section_enter() {
+    if (_lf_num_nested_critical_sections++ == 0) {
+        // First nested entry into a critical section.
+        // If interrupts are not initially enabled, then increment again to prevent
```
erlingrj, line 282:
```suggestion
```

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * Make calling thread wait for termination of the thread.  The
+ * exit status of the thread is stored in thread_return, if thread_return
+ * is not NULL.
+ * 
+ * @return 0 on success, platform-specific error number otherwise.
+ */
+int lf_thread_join(lf_thread_t thread, void** thread_return) {
+    k_thread_join(thread, K_FOREVER);
```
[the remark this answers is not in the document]
erlingrj: THanks yes. Of course we should return the return value from `k_thread_join`

core/platform/lf_zephyr_support.c
erlingrj: Thanks. It seems as if the `val` version should return the `type.` Which in this implementation would be an `int` and the value should be the original value of `ptr`. Thanks, nice catch. The threaded Zephyr runtime isn't properly tested. I think I will add a Preprocessor warning about this. Alternatively we could completely disallow it by rejecting LF programs with platform=Zephyr and threading: true

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,527 @@
...
+
+        _lf_alarm_cfg.ticks = sleep_duration_ticks;
+        int err = counter_set_channel_alarm(_lf_counter_dev, LF_TIMER_ALARM_CHANNEL,  &_lf_alarm_cfg);
+     
+        if (err != 0) {
+            lf_print_error_and_exit("Could not setup alarm for sleeping. Errno %i", err);
+        }
+        
+        lf_critical_section_exit();
+        k_sem_take(&_lf_sem, K_FOREVER);
```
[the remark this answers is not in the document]
petervdonovan: Okay, sounds reasonable

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,525 @@
...
+    if (res == 0) {
+        return 0;
+    } else {
+        return LF_TIMEOUT;
+    }
+}
+
+// Atomics
+//  Implemented by just entering a critical section and doing the arithmetic
+//  FIXME: We are now restricted to atomic integer operations
```
edwardalee, line 484: What does this FIXME mean? Do we need anything more than atomic integer operations?

Also, these functions need documentation.

core/platform/lf_zephyr_support.c
```diff
@@ -0,0 +1,525 @@
...
+
+int _zephyr_atomic_add_fetch(int *ptr, int value) {
+    lf_critical_section_enter();
+    int res = *ptr + value;
+    *ptr = res;
+    lf_critical_section_exit();
+    return res;
+}
+
+// FIXME: Are you sure that it should return bool?
```
edwardalee, line 501: Answering this FIXME would be easier if the function were documented.

edwardalee (approved):
LGTM.

core/platform/lf_zephyr_support.c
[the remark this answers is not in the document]
erlingrj: I opt`ed for a rename as you suggested

core/platform/lf_zephyr_support.c
erlingrj: Removed it and added documentation

core/platform/lf_zephyr_support.c
erlingrj: Yes. Fixed

erlingrj:
@lhstrh just to notify you. After suggestion from Peter, I changed `lf_sleep_until` to `lf_sleep_until_locked` since we assume that it is always called from `next` when we are in a critical section. We might also just "remove" it as a user-api by calling it `_lf_sleep_until_locked`. Instead we have `lf_sleep` and `lf_nanosleep` which are user-facing and does not require a critical section

core/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ endmacro()
 # parameters to Cmake.
 target_compile_definitions(core PRIVATE INITIAL_EVENT_QUEUE_SIZE=10)
 target_compile_definitions(core PRIVATE INITIAL_REACT_QUEUE_SIZE=10)
+target_compile_definitions(core PUBLIC RUNTIME_CHECKS)
```
erlingrj, line 68:
```suggestion
```

PR #105, opened 2022-08-31
Add basic authentication using HMAC for federates joining a federation

hokeun:
This PR is a submodule PR of the PR lf-lang/lingua-franca#1432 and is the first step towards resolving lf-lang/lingua-franca#1146. Both PRs should be merged to work.

Overall information is also in lf-lang/lingua-franca#1432. This PR details the HMAC authentication implementation.

This PR is about the RTI.c and federate.c authentication implementation.

The main idea of security is not to send the federation ID in plain text, by using it as an HMAC key.

### 1. RTI_HELLO (RTI -> Federate) ###
The RTI sends a RTI_HELLO message to the federate when TCP connection is established. The message is in this order:
- One byte equal to MSG_TYPE_RTI_RESPONSE.
- Eight bytes for randomly created RTI's nonce.
### 2. FED_RESPONSE  (Federate -> RTI) ###
The federate sends a response message to the RTI. It creates it's own nonce, and makes a HMAC tag using the federate ID as the key. The message is in the following order:
- One byte equal to MSG_TYPE_FED_RESPONSE.
- Eight bytes for federate's nonce.
- Two bytes for the federate ID.
- 32 bytes for HMAC tag based on SHA256. 

 The HMAC tag is composed of the following order:
- One byte equal to MSG_TYPE_FED_RESPONSE.
- Two bytes for the federate ID.
- Eight bytes for received RTI's nonce.

### 3. RTI_RESPONSE  (RTI -> Federate) ###
The RTI sends a response message to the federate. It checks the received HMAC created by the federate, and also makes a HMAC tag. The message is in the following order:
- One byte equal to MSG_TYPE_RTI_RESPONSE.
- 32 bytes for HMAC tag based on SHA256.

The HMAC tag is composed of the following order:
- One byte equal to MSG_TYPE_RTI_RESPONSE.
- Eight bytes for received federate's nonce.

### Federate Check ###
The federate finally checks the received HMAC created by the RTI.

## Discussion ##
The 3-way handshake protocol does not change the existing federation id check process. ```receive_and_check_fed_id_message()``` still always happen after the 3-way handshake. So, the ```auth:true``` option only enables the 3-way handshake. Changing the protocol will be discussed at the future.

Also, the ```auth:true``` option will be changed from boolean to string in future.

PR #137, opened 2023-01-07
Make improvements on CMake

erlingrj:
Mainly changing how reactor-c corelib is built. Originally it used `add_subdirectory` to add platform, modal, threaded and utils sources. This is a bit hacky since it creates new node and compile definitions and sources can not easily be added to the parent node. Now I use `include` instead. Needed some additional changes for this to work.


erlingrj:
Btw, @petervdonovan can you immediately see why the CI is not running on my PRs?

petervdonovan:
No, I'm not sure why not. I think that sometimes one has to merge master into the branch in order to trigger CI.

petervdonovan (approved):
This looks good to merge. Thanks for the cleanup!

erlingrj:
OK. Just have to fix the CI.  @lhstrh. There is something wrong with the CI see here: https://github.com/lf-lang/reactor-c/actions/runs/3864134424/workflow

image

Do you understand anything of this?

petervdonovan:
Oh, I get it now. Dongha just forgot to change from `auth` back to `master` when merging #105, and the `auth` branch in `lingua-franca` has since been deleted. So if, in this PR, we just change `auth` back to `master` in `reactor-c/.github/workflows/ci.yml`, it should be fine again.

Issue #133, opened 2022-12-21
Federated test programs sometimes fails in CI


erlingrj:
Adding this issue as a reminder to check this out. It has occurred on lock-time branch, LET branch and also currently on main. This problem quite consistently goes away if you rerun the failed tests. It is obviously a race condition, but is it in our runtime or in the test framework?

PR #138, opened 2023-01-09
Update Reactor-C with changes to facilitate arduino-cli support

arengarajan99:
Minor changes are needed in reactor-c to faciliate arduino-cli support. Compiles successfully locally with changes in LFC


core/platform/lf_arduino_support.c
```diff
@@ -57,6 +58,10 @@ static volatile int _lf_num_nested_critical_sections = 0;
 static volatile uint32_t _lf_time_us_high = 0;
 static volatile uint32_t _lf_time_us_low_last = 0;
 
+int lf_ack_events() {
```
petervdonovan, line 61: This is only tangentially related to this PR, but I searched this repo and could not find `lf_ack_events` doing anything useful. The only thing I know about it is from [this comment](https://github.com/lf-lang/reactor-c/pull/106#discussion_r962182789), and I'm not sure whether we still need it.

petervdonovan (approved):
LGTM!

core/platform/lf_arduino_support.c
erlingrj: My opinion is that we just remove it. We don't need to hide setting/clearing a single flag behind a function

erlingrj: The calls to `lf_ack_events()` should be replaced by `_lf_async_event = false`

include/core/platform/lf_arduino_support.h
```diff
@@ -90,6 +90,19 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
+
+#ifndef _SYS__TIMESPEC_H_
+#define	_SYS__TIMESPEC_H_
+
+struct timespec {
+	long long	tv_sec;		/* seconds */
+	long	tv_nsec;	/* and nanoseconds */
+};
```
erlingrj, line 102: What is this used for?

erlingrj:
Good work. The test failure is related to #133. I don't know when this started or what the cause is. We should look into it at some point. But just re-run the failing tests and it always just works

include/core/platform/lf_arduino_support.h
arengarajan99: AVR doesn't have timespec to hold the time when you first load the program. MBED does, so I just made the custom timespec struct to hold it.

core/platform/lf_arduino_support.c
lhstrh: I'm OK with removing it, although I think that hiding the clearing of a flag behind a function is cleaner. I do think that the name problematic (my bad). Maybe `_lf_clear_event_flag` or something like that would be better.

lhstrh (commented):
Left a remark about the CI config.

include/core/platform/lf_arduino_support.h
erlingrj: Hm. I don't understand what you mean. Is there some external dependency we are linking with that needs this struct defined?

arengarajan99:
One comment I had for @erlingrj is the reason behind removing using Arduino delay functionality? I must have missed when this was taken out.

include/core/platform/lf_arduino_support.h
lhstrh: I'm also a bit puzzled by this code. Do we actually need it?

arengarajan99: Inside of reactor_common.c, when we initialize we have the following set of code:
```
LF_PRINT_DEBUG("Start time: " PRINTF_TIME "ns", start_time);
struct timespec physical_time_timespec = {physical_start_time / BILLION, physical_start_time % BILLION};
printf("---- Start execution at time %s---- plus %ld nanoseconds.\n",
            ctime(&physical_time_timespec.tv_sec), physical_time_timespec.tv_nsec);
```

This is the only place the timespec is referenced for the Arduino lifecycle, but I myself don't know what purpose this has.

arengarajan99: For MBED systems like the BLE Sense Boards, timespec is already defined, but AVR doesn't have it formally defined. To prevent the system from breaking, I simply hardcoded the timespec struct. That said, I still don't know the underlying purpose of the struct if we simply read to and from the struct and have it get discarded later?

arengarajan99: It looks like the federated reactor still needs timespec and I'm currently working on federated support for Arduino so I think it's best to keep it.

PR #140, opened 2023-01-11
Removed Interrupts Functionality from Arduino to Solve Unknown Blocking Problem

arengarajan99:
Currently, there exists bug(s) that cause Arduino to hang when we use interrupts. The problem area was found to be caused by calloc, which should still work when interrupts are disabled but does not under TBD circumstances. To unblock developers, we won't support interrupts for now (no physical actions) but can add them later.

### DIFF
--- a/core/platform/lf_arduino_support.c
+++ b/core/platform/lf_arduino_support.c
@@ -133,18 +133,23 @@ int lf_clock_gettime(instant_t* t) {
  * nested critical sections.
 */
 int lf_critical_section_enter() { 
-    if (_lf_num_nested_critical_sections++ == 0) {
-        // First nested entry into a critical section.
-        // If interrupts are not initially enabled, then increment again to prevent
-        // TODO: Do we need to check whether the interrupts were enabled to
-        //  begin with? AFAIK there is no Arduino API for that 
-        noInterrupts();
-    }
+    _lf_num_nested_critical_sections++;
+    // if (_lf_num_nested_critical_sections++ == 0) {
+    //     // First nested entry into a critical section.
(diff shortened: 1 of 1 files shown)


petervdonovan:
Are interrupts disabled to begin with? Or does the code generator do something to prevent the programmer from using them?

Just checking, because if interrupts are not completely disabled, then these changes mean that `lf_critical_section_enter` will fail to meet its specification.

arengarajan99:
Interrupts in arduino are default enabled when you start the program. Interrupts are very brittle in the sense that if you disable interrupts, you are limited in what you can do (timer overflow problems, Serial buffer crashing). Our scheme of doing entering and exiting critical sections in different functional locations is counter to the expectation of how Arduino was designed (small, one-to-few line critical sections), which is causing lots of crashing and hanging. We do need to implement this functionality, but we need to unblock developers for the meantime.

petervdonovan (approved):
This seems reasonable to me. Thanks for the explanation Anirudh

lhstrh:
What is the prospect for a long-term solution?

arengarajan99:
Long term we would need to support sending Lingua Franca messages over Serial by ensuring all lf_prints can be interrupt safe. I can get more progress on a patch once I have access to the Arduino Zero board which allows me to do GDB style step through for debugging.

erlingrj:
We should add a check in `lingua-franca` that throws an error if there is a physical action in a program with Arduino as target platform. I am really bewildered by this issue and look forward to learning what are the underlying mechanisms of the Arduino runtime that makes disabling interrupts dangerous. 

Certainly gives a new meaning to  "Interrupt-safe"

erlingrj (approved):

PR #141, opened 2023-01-13
Preliminary multi-board support for Arduino

arengarajan99:
Different Arduino boards have different infrastructures, which requires us to determine what board type we are flashing to (AVR, SAM, SAMD, megaAVR, or MBED). Currently, the only change we need to make on reactor-c is ensuring MBED boards have a different call for disabling and enabling interrupts. This pull request will allow us to re-enable physical actions on Arduino boards.


arengarajan99:
@erlingrj It may be good to see if this way of using compile-time params in the board type is sufficient. In cases where we need the specific board rather than its family, we can simply pull those up via referencing its specific compile-time macro.

include/core/platform/lf_arduino_support.h
```diff
@@ -36,57 +36,65 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-#elif defined(ARDUINO_AVR_ROBOT_CONTROL)       
-    #define BOARD "Robot Ctrl"
-#elif defined(ARDUINO_AVR_ROBOT_MOTOR)       
-    #define BOARD "Robot Motor"
-#elif defined(ARDUINO_AVR_UNO) || defined(__AVR_ATmega4809__)       
-    #define BOARD "Uno"
-#elif defined(ARDUINO_AVR_YUN)       
-    #define BOARD "Yun"
+    #define BOARD AVR
+#elif defined(ARDUINO_AVR_BT)    // Bluetooth
```
lhstrh, line 48: There appears to be a lot of weird white space on these lines that we should get rid of.

core/platform/lf_arduino_support.c
```diff
@@ -134,13 +134,17 @@ int lf_clock_gettime(instant_t* t) {
 */
 int lf_critical_section_enter() { 
     _lf_num_nested_critical_sections++;
```
lhstrh, line 136: Why are we incrementing twice? Once on this line and once in the if? Also tagging @erlingrj to get more :eyes: on this.

lhstrh (approved):

core/platform/lf_arduino_support.c
erlingrj: That doesnt look right? I dont think `_lf_num_nested_critical_sections++ == 0` will ever be `true` with this design. There are two decrements in the `leave` function also.

include/core/platform/lf_arduino_support.h
arengarajan99: Fixed.

core/platform/lf_arduino_support.c
arengarajan99: There were left over decrement and increment statements from when we needed to disable interrupt support, so I reverted it back to what it was before.

lhstrh (approved):

PR #142, opened 2023-01-15
Make critical section functions part of the public API

edwardalee:
This PR removes the distinction between `_lf_critical_section_enter()` and `lf_critical_section_enter()`, using only the latter, and makes the function always available, not just for unthreaded targets. Similarly for `lf_critical_section_exit()` and `lf_notify_of_event()`. This enables programmers to use these functions. For example, in the MQTT reactors, this is useful when trying to deterministically assign a timestamp to an incoming message.

### DIFF
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -218,7 +218,7 @@ int _lf_do_step(void) {
 int next(void) {
     // Enter the critical section and do not leave until we have
     // determined which tag to commit to and start invoking reactions for.
-    if (_lf_critical_section_enter() != 0) {
+    if (lf_critical_section_enter() != 0) {
         lf_print_error_and_exit("Could not enter critical section");
     }
     event_t* event = (event_t*)pqueue_peek(event_q);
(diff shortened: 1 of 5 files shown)


erlingrj (approved):
Thanks for addressing that. This looks better. I think it should work. The extra level of indirection was added to address the scenario where we have threading:false but tracing:true. In that scenario we want the C-runtime to use normal unthreaded API (e.g. disable interrupts) while the tracing files need the threaded API.

erlingrj:
Currently there is a mix of threaded and unthreaded in that case. It will work with linux/mac/win since the platform support implementing threaded/unthreaded is header-only. So `trace.c` can include the threaded version and `reactor.c` can include the unthreaded. If you look in `trace.c` you will see that it defines `_LF_TRACE` before including `platform.h` and then undefs it. This is to achieve this. The reason it is needed is that tracing creates threads and uses mutexes. We had some discussion of whether tracing could be implemented without any thread-support.

edwardalee:
The test failure has nothing to do with this PR and has been showing up intermittently. I think it is due to excessively large debug output and I've separately pushed a change that should reduce the size. Merging this PR.

PR #144, opened 2023-01-20
Token payload free

edwardalee:
This PR refactors the token handling code considerably and delivers clean memory reports from valgrind (or the Mac equivalent, leaks). It still includes many deprecated APIs for working with tokens, but the plan is to eliminate those in a separate PR.

This branch will only work with the token-payload-free branch of lingua-franca. After merging, I guess lingua-franca-ref.txt will need to be updated.


core/lf_token.c
```diff
@@ -0,0 +1,358 @@
...
+
+/**
+ * Set of token templates (trigger_t or port_base_t objects) that
+ * have been initialized. This is used to free their tokens at
+ * the end of program execution. 
+ */
+static hashset_t _lf_token_templates = NULL;
+
+////////////////////////////////////////////////////////////////////
+//// Functions that user's may call.
```
lhstrh, line 67:
```suggestion
//// Functions that users may call.
```

core/utils/hashset/README.md
```diff
@@ -0,0 +1,25 @@
+# hashset.c
+
+A hash set implementation in C from:
+https://github.com/avsej/hashset.c
```
lhstrh, line 4: Were fixes necessary in this code? If not, then why not include it as a submodule?

include/core/utils/hashset/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
lhstrh, line 1: Do we really need to include this file twice and have it in two different locations? Note that it doesn't even list a copyright holder...

core/utils/hashset/test.c
```diff
@@ -0,0 +1,223 @@
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+#include "hashset/hashset.h"
+#include "hashset/hashset_itr.h"
+
+static void trivial(void)
```
lhstrh, line 7: Is this code run in CI?

include/core/lf_token.h
```diff
@@ -0,0 +1,362 @@
...
+ * struct that cannot be freed by a simple call to free() or copied by a call
+ * to memcpy().
+ * 
+ * An instance of a port struct and trigger_t struct (an action or an input port)
+ * can be cast to token_template_t, which has a token_type_t field called type
+ * and a pointer to a token (which may be NULL).  The same instance can also be
+ * cast to token_type_t, which has an element_size field and (possibly) function
+ * pointers to a destructor and a copy constructor.
+ * 
+ * A "template token" is one pointed to by a token_template_t (an action or a port).
```
lhstrh, line 50: It's still not really clear to me what a template token _is_ (or why we have it) after reading this text.

include/core/lf_token.h
```diff
@@ -0,0 +1,362 @@
...
+
+/**
+ * Token type for dynamically allocated arrays and structs sent as messages.
+ *
+ * This struct is the wrapper around the dynamically allocated memory
+ * that carries the message.  The message can be an array of values,
+ * where the size of each value is element_size (in bytes). If it is
+ * not an array, or is not to be treated as an array, the length == 1.
+ *
+ * In the C LF target, a type for an output that ends in '*' or '[]' is
```
lhstrh, line 101: It would be great if we implement @oowekyala's suggested syntax for token types in LF because this, I feel, is too hacky.

core/lf_token.c
```diff
@@ -0,0 +1,358 @@
...
+}
+
+lf_token_t* lf_writable_copy(lf_port_base_t* port) {
+    assert(port != NULL);
+
+    lf_token_t* token = port->tmplt.token;
+    if (token == NULL) return NULL;
+    LF_PRINT_DEBUG("lf_writable_copy: Requesting writable copy of token %p with reference count %zu.",
+            token, token->ref_count);
+    // FIXME: token_template_t needs to be augmented with a field
```
lhstrh, line 80: Please document this in an issue and reference it in the code.

core/lf_token.c
```diff
@@ -0,0 +1,358 @@
...
+    return _lf_free_token(token);
+}
+
+void _lf_free_token_copies() {
+    while (_lf_tokens_allocated_in_reactions != NULL) {
+        lf_token_t* next = _lf_tokens_allocated_in_reactions->next;
+        _lf_done_using(_lf_tokens_allocated_in_reactions);
+        _lf_tokens_allocated_in_reactions = next;
+    }
+}
```
lhstrh, line 358:
```suggestion
}

```

lhstrh (approved):
This looks great! How is the clean Valgrind report that you mentioned obtained? Is this something we could do in CI? At the very least we might want to document which commands to run to assure that any changes to this rather critical code don't introduce memory leaks.

include/core/lf_token.h
petervdonovan: It looks like it supports the persistence of port values until they are overwritten, as Edward has been advocating for (?)

include/core/lf_token.h
```diff
@@ -0,0 +1,362 @@
...
+ * @brief Base type for ports.
+ * Port structs are customized types because their payloads are type
+ * specific. This struct represents their common features. Given any
+ * pointer to a port struct, it can be cast to lf_port_base_t and then
+ * these common fields can be accessed.
+ * IMPORTANT: If this is changed, it must also be changed in
+ * CPortGenerator.java generateAuxiliaryStruct().
+ */
+typedef struct lf_port_base_t {
+	token_template_t tmplt;               // Type and token information (template is a C++ keyword).
```
petervdonovan, line 155: I have heard from @erlingrj that ports consume a substantial amount of memory in the C target. Will this addition to `lf_port_base_t` cause problems on resource-constrained platforms? Or could it cause more general performance problems on platforms with caches?

edwardalee: I don't think `lf_port_base_t` has added any fields to the port structs.  It just rearranges them so that the common fields across ports always come first, so any port struct can safely be cast to `lf_port_base_t`.

petervdonovan: Okay, maybe I misunderstood.

I was assuming that because `port_base_t` now includes this `token_template_t tmplt` field which it did not have before, that would mean that all ports have the field, even if they do not use tokens. `token_template_t tmplt` alone looks like it should consume 5 quadwords, whereas the part of `port_base_t` that existed before this PR is 3 quadwords and the total memory usage that was reported in the [NGRES23 paper](https://github.com/icyphy/lf-pubs/blob/master/ngres23/ngres23.tex) was 5 quadwords I think.

erlingrj: I observed that ports need memory. I think we were limited to 10 Reactors in the Big benchmark from Savina on the nrf52. This is because there was an all-to-all connection topology. So we had, in fact a total of a 100 ports in that program.

edwardalee: @petervdonovan is correct here and the new design does use more memory for each port. I would prefer to get this merged and deal with this in a separate PR. I've opened issue #151 for this.

edwardalee:
> This looks great! How is the clean Valgrind report that you mentioned obtained? Is this something we could do in CI? At the very least we might want to document which commands to run to assure that any changes to this rather critical code don't introduce memory leaks.

On a Mac, the test looks like this (for one of the tests):
```
cd test/C
lfc src/token/TokenContainedPrint.lf
leaks -atExit -- bin/TokenContainedPrint
```
The output of the `leaks` program is:
```
---- Start execution at time Sun Jan 29 10:07:25 2023
...
leaks Report Version: 4.0
Process 84454: 205 nodes malloced for 35 KB
Process 84454: 0 leaks for 0 total leaked bytes.
```
A valgrind test, I guess, would be similar. I don't know how to include in the CI. Likely it will expose more leaks.

core/utils/hashset/README.md
edwardalee: The original hashset code had serious bugs, poor documentation, and poor generalizability. This is a pretty substantial rewrite. I suppose we could issue a PR in the original repo, but I don't think we want to depend on that repo.

include/core/utils/hashset/LICENSE
edwardalee: Actually, I don't think we need to include either LICENSE file. The source files themselves point to the license terms online.

core/utils/hashset/test.c
edwardalee: This code is not run in CI. I'm not sure how to do that.

include/core/lf_token.h
edwardalee: I have improved this documentation and pushed to the examples repository an [implementation of the ROSACE case study](https://github.com/lf-lang/examples-lingua-franca/blob/main/C/src/periodic/README.md) (an aircraft controller) that requires persistence of inputs. This example would be much more complicated, I think, in the Cpp target, which does not support persistent inputs.

core/lf_token.c
edwardalee: Actually, this FIXME is fixed.  Removed the comment.

PR #147, opened 2023-01-20
Relative paths lf_types.h

gundralaa:
changed the modes.h header path to avoid conflict


lhstrh (approved):
Looks good to me! Could you leave a brief summary in the description that describes the need for this change? It's good to have that documented. Thanks!

PR #148, opened 2023-01-21
Delete util.c

lhstrh:
This PR removes `util.c` because it has been deprecated and causes conflicts in build configurations.


edwardalee (approved):
I'm all for removing lib/util.c. However, this will require updating quite a few tests in the lingua-franca repo and examples in the examples repo.  This should be straightforward, but a bit tedious.

erlingrj:
I have made a PR with the required changes in the examples repo. here https://github.com/lf-lang/examples-lingua-franca/pull/19

Will look at the tests next

erlingrj:
As far as I can see, the deprecated API is only used in some Python tests. How would this affect the Python target? Is there a Python wrapper around the C-API, or can you actually call the C API directly from Python. In the case of the latter we should just update those Python tests. I can do it

erlingrj:
This PR removes use of the deprecated API in the Python tests: https://github.com/lf-lang/lingua-franca/pull/1566

PR #150, opened 2023-01-27
Disable MBED Support since it is a multithreaded enabled board.

arengarajan99:
MBED is more complicated than other Arduino boards since it has an onboard RTOS, which forces us to have the multithreaded reactor-c support present due to issues with hanging threads. We disable MBED boards to kickstart Arduino Support and will integrate it back later.


lhstrh (approved):
LGTM!

PR #153, opened 2023-02-02
Threading support for Arduino MBED/coupling of mutexes and condition variables

arengarajan99:
Arduino MBED boards use an onboard RTOS that prevent us from using standard Arduino API calls for disabling interrupts that we need to support physical actions. 

To take advantage of the RTOS, I've added threading support for Arduino MBED Boards that uses locks to control critical sections and spawns threads as needed for jobs.


lhstrh:
@arengarajan99 could you have a look at the conflict with `main`?

arengarajan99:
@lhstrh Looks like a pending pull request needs to go through in lingua-franca here before I can merge these pull requests through. https://github.com/lf-lang/lingua-franca/pull/1586/files#diff-553dd41b4a914c16b1eab867893f90250d263953fd4891331f9078d6931a453c

arengarajan99:
This pull request also couples conditionvariables and mutexes together as well as clean up variable naming

core/platform/lf_arduino_support.c
```diff
@@ -179,3 +174,138 @@ int lf_notify_of_event() {
...
+    bool res = condition_wait_for(*cond, sleep_duration_ns);
+    if (!res) {
+        return 0;
+    } else {
+        return LF_TIMEOUT;
+    }
+}
+
+#endif
+#endif
```
lhstrh, line 311:
```suggestion
#endif

```

core/platform/lf_linux_support.c
```diff
@@ -72,3 +73,4 @@ int lf_sleep_until_locked(instant_t wakeup_time) {
 int lf_nanosleep(interval_t sleep_duration) {
     return lf_sleep(sleep_duration);
 }
+#endif
```
lhstrh, line 76:
```suggestion
#endif

```

core/platform/lf_linux_support.c
```diff
@@ -1,3 +1,4 @@
+#if !defined(PLATFORM_ARDUINO)
```
lhstrh, line 1: It seems a little strange that this would be necessary. Clearly, if the platform is Linux, it isn't Arduino. Same issue in other support file as well. I bet this can be refactored so that this isn't needed.

lhstrh (dismissed):
Bundling the condition variable with the mutex makes for much cleaner code and I like it a lot! The appearance of the `#if !defined(PLATFORM_ARDUINO)` macro all over the place, however, doesn't look right to me. I suspect the build configuration can be altered to where this isn't necessary.

include/core/platform.h
```diff
@@ -208,7 +202,7 @@ extern int lf_cond_signal(lf_cond_t* cond);
  *
  * @return 0 on success, platform-specific error number otherwise.
  */
-extern int lf_cond_wait(lf_cond_t* cond, lf_mutex_t* mutex);
+extern int lf_cond_wait(lf_cond_t* cond);
```
edwardalee, line 205: Removing the mutex argument from all these API functions seems dangerous to me.  It seems to force a design that has only a single global mutex, precluding any design with multiple mutexes. This could be a serious limitation. Is this really necessary?  The argument could be ignored for platforms that really do force just a single mutex.

edwardalee (changes requested):
I'm concerned about removing the mutex argument from all the threading API arguments.  Doesn't this box us in to designs that use only a single global mutex?

include/core/platform.h
lhstrh: I don't think we force the use of a single mutex with this design; AFAIK the `lf_cond_t` type simply bundles a condition variable with a mutex (any mutex).

include/core/platform/lf_C11_threads_support.h
```diff
@@ -36,9 +36,12 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <stdlib.h>
 #include <stdint.h> // For fixed-width integral types
 
-typedef mtx_t _lf_mutex_t;
-typedef cnd_t _lf_cond_t;
-typedef thrd_t _lf_thread_t;
+typedef mtx_t lf_mutex_t;
+typedef struct {
+    lf_mutex_t* mutex;
```
lhstrh, line 41: @edwardalee have a look here

edwardalee: Ah, OK, I missed this.

arengarajan99 (commented):
Resolved

lhstrh (approved):

PR #154, opened 2023-02-02
Fix scheduling preprocessing

erlingrj:
The conditional inclusion of code based on which scheduler was chosen didnt work.
```
#if SCHEDULER == LET
```
Does not work if you do `-DSCHEDULER=LET`. A fix is to define macros for `LET` etc. But this means that the header file with these `#define`s must be included before any use of `#if SCHEDULER == ...`


petervdonovan (approved):
LGTM

PR #159, opened 2023-02-15
Fix problem with threaded Zephyr and overflow handling in QEMU emulation

erlingrj:
This PR addresses various bugs that I discovered when debugging the QEMU emulation of the threaded runtime.
1. When number of workers were not specified the system could not create any workers through `lf_thread_create`. Fixed by turning off threading support until we know whether `lf_thread_create` should be a user-facing API.
2. In the threaded runtime, the adaptive scheduler was still used because it was missing a `#inlude "lf_types.h"` at the top.
3. Some left-over debug-code was using the low-res timer also in the `lf_clockgettime` implementetion for the hi-res timer.
4. The implementation of `lf_clock_gettime` for the low-res timer (which is what is used by QEMU), did not handle overflows which caused the emulation to stop after some

### DIFF
--- a/core/platform/lf_zephyr_support.c
+++ b/core/platform/lf_zephyr_support.c
@@ -144,7 +144,9 @@ void lf_initialize_clock() {
 // Zephyrs Counter API
 
 /**
- * Return the current time in nanoseconds
+ * Return the current time in nanoseconds. It gets the current value
+ * of the hi-res counter device and also keeps track of overflows
+ * to deliver a monotonically increasing clock.
  */
 int lf_clock_gettime(instant_t* t) {
     uint32_t now_cycles;
@@ -154,8 +156,6 @@ int lf_clock_gettime(instant_t* t) {
     res = counter_get_value(_lf_counter_dev, &now_cycles);
     now_nsec = counter_ticks_to_us(_lf_counter_dev, now_cycles)*1000ULL;
     *t = now_nsec + _lf_timer_last_epoch_nsec;
-    now_cycles = k_cycle_get_32();
-    *t = (SECOND(1)/CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC)*now_cycles;
     return 0;
 }
 
--- a/core/platform/lf_zephyr_support.c
+++ b/core/platform/lf_zephyr_support.c
@@ -350,7 +350,7 @@ int lf_notify_of_event() {
 // FIXME: What is an appropriate thread prio?
 #define _LF_THREAD_PRIORITY 5
 
-// If NUMBER_OF_WORKERS is not specified, or specified to 0. Then we default to 1
+// If NUMBER_OF_WORKERS is not specified, or set to 0, then we default to 1.
 #if !defined(NUMBER_OF_WORKERS) || NUMBER_OF_WORKERS==0
 #undef NUMBER_OF_WORKERS
 #define NUMBER_OF_WORKERS 1
--- a/core/platform/lf_zephyr_support.c
+++ b/core/platform/lf_zephyr_support.c
@@ -45,6 +45,9 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // Combine 2 32-bit words to a 64-bit word
 #define COMBINE_HI_LO(hi,lo) ((((uint64_t) hi) << 32) | ((uint64_t) lo))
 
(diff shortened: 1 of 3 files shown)


core/platform/lf_zephyr_support.c
```diff
@@ -337,6 +337,13 @@ int lf_notify_of_event() {
 #define _LF_STACK_SIZE 1024
 // FIXME: What is an appropriate thread prio?
 #define _LF_THREAD_PRIORITY 5
+
+// If NUMBER_OF_WORKERS is not specified, or specified to 0. Then we default to 1
```
lhstrh, line 341:
```suggestion
// If NUMBER_OF_WORKERS is not specified, or set to 0, then we default to 1.
```

core/platform/lf_zephyr_support.c
```diff
@@ -367,7 +374,8 @@ int lf_available_cores() {
 int lf_thread_create(lf_thread_t* thread, void *(*lf_thread) (void *), void* arguments) {
     // Use static id to map each created thread to a 
     static int tid = 0;
-    
+
+    // Make sure we dont try to create too many threads
```
lhstrh, line 378: If `lf_thread_create` is called in user code, then the number of workers isn't really relevant, is it? Does this really belong here?

core/threaded/reactor_threaded.c
```diff
@@ -1031,7 +1031,9 @@ void start_threads() {
     LF_PRINT_LOG("Starting %u worker threads.", _lf_number_of_workers);
     _lf_thread_ids = (lf_thread_t*)malloc(_lf_number_of_workers * sizeof(lf_thread_t));
     for (unsigned int i = 0; i < _lf_number_of_workers; i++) {
-        lf_thread_create(&_lf_thread_ids[i], worker, NULL);
+        if (lf_thread_create(&_lf_thread_ids[i], worker, NULL) != 0) {
```
lhstrh, line 1034: It seems that the check implemented in `lf_thread_create` should be done here, at the call site, before even invoking `lf_thread_create`...

lhstrh (commented):
Left some questions...

core/platform/lf_zephyr_support.c
erlingrj: I think it is bad practice to do too much dynamic spawning of threads in a microcontroller setting. You want to avoid dynamic memory allocation if you can. Therefore the stacks and thread control blocks are statically allocated. This puts an upper limit to how many times you can call lf_thread_t. The user can use ordinary calls to zephyrs API to create other threads.

core/threaded/reactor_threaded.c
erlingrj: Since I have statically allocated the stacks for the threads I have to check it inside `lf_thread_create` to make sure that it is not called from somewhere else

core/platform/lf_zephyr_support.c
lhstrh: While I don't disagree with you on this, AFAIK, the Zephyr implementation is the only one that does this. My point of criticism has to do with the fact that the function is `lf_thread_create` not `lf_worker_create`. Threads and workers are two different things, and you're now looking at the number of workers in the thread creation function, which I think is odd...

erlingrj: Hm, I would argue that reactor-c is not a threading library and that user-code should not use the LF threading API. That would make the runtime more attractive for use in these embedded systems where people are wary of using dynamic memory allocation. There is already a lot of that in the runtime though, but theoretically, most of it can be replaced by static allocations.

lhstrh: The function does not have a leading underscore, so it is part of the public API...

erlingrj: Zephyr has a POSIX implementation that we could use temporarily, or we can just turn of threaded support for Zephyr, that is fine with me. With the bodiless reactions we have to make a decision about what is actually going to be part of the user-facing API since those includes must be code-generated. So let's make the right decision there and revisit this later

lhstrh: I'm not sure I understand your proposal. Here are some possible solutions that crossed my mind:
 - have a define `MAX_THREAD_COUNT` and check against that
 - update the docs and have all implementations check against `NUMBER_OF_WORKERS`
 - rename to `_lf_thread_create`
 - introduce a `lf_worker_create` that does the check and then just invokes `lf_thread_create`
 - ...?

lhstrh (approved):

PR #160, opened 2023-02-16
Intended tag printed instead of last-seen tag in debug message

byeong-gil:
The log message in [rti.c/#L891](https://github.com/lf-lang/reactor-c/blob/2ec5deff12daf66ce106892083443fc85385d035/core/federated/RTI/rti.c#L891) says that it prints the received next event tag but actually it prints the current next event tag of its federate.


lhstrh (approved):
LGTM!

PR #163, opened 2023-02-22
Migrate logic from CMake to preprocessor

petervdonovan:
I realized that in the amount of time I spend talking about doing this, I could probably just do it... So here we go.

Motivated by https://github.com/lf-lang/reactor-c/pull/153#pullrequestreview-1306818160. If this passes checks, I propose to merge it into #153 and then merge #153 into main.


lhstrh (approved):
This looks so much cleaner! Thanks for taking the lead on this!

PR #164, opened 2023-02-22
Federated Support without C include reliance

arengarajan99:


include/core/federated/net_common.h
```diff
@@ -196,7 +196,9 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef NET_COMMON_H
 #define NET_COMMON_H
 
+#ifndef PLATFORM_ARDUINO
```
lhstrh, line 199: This doesn't seem the be the right macro.

include/core/federated/net_util.h
```diff
@@ -38,9 +38,12 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef NET_UTIL_H
 #define NET_UTIL_H
 
+#ifndef PLATFORM_ARDUINO
```
lhstrh, line 41: Idem

core/federated/RTI/CMakeLists.txt
```diff
@@ -50,11 +50,11 @@ else()
 endif()
 
 # The following should not be done
-include_directories(${CoreLib}/..)
-include_directories(${CoreLib})
-include_directories(${CoreLib}/federated)
-include_directories(${CoreLib}/platform)
-include_directories(${CoreLib}/utils)
+# include_directories(${CoreLib}/..)
```
lhstrh, line 53: Should we just delete these lines then?

core/federated/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(FEDERATED_SOURCES clock-sync.c federate.c net_util.c)
+list(APPEND INFO_SOURCES ${FEDERATED_SOURCES})
+
+list(TRANSFORM FEDERATED_SOURCES PREPEND federated/)
+target_sources(core PRIVATE ${FEDERATED_SOURCES})
```
lhstrh, line 5:
```suggestion
target_sources(core PRIVATE ${FEDERATED_SOURCES})

```

core/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ list(APPEND INFO_SOURCES ${GENERAL_SOURCES})
 # Create the core library
 add_library(core ${GENERAL_SOURCES})
 
+# Add sources for either threaded or unthreaded runtime
+if(DEFINED FEDERATED)
```
lhstrh, line 19:
```suggestion
if (DEFINED FEDERATED)
```

lhstrh (changes requested):
Generally looks very good, but the `ifndef ARDUINO` looks out of place.

core/federated/RTI/CMakeLists.txt
arengarajan99: Sure

include/core/federated/net_common.h
arengarajan99: Arduino doesn't have pthread.h support.

include/core/federated/net_util.h
arengarajan99: Same as above.

include/core/federated/net_common.h
lhstrh: That I understand, but this could become a very long list eventually. Don't we have a `THREADED` macro or something along those lines? Maybe @petervdonovan as suggestions?

edwardalee: LF_THREADED is set if the threaded runtime is being used.

arengarajan99: The issue is that Arduino boards have threading but don't use the pthread library

lhstrh: Looks like this particular include isn't needed, actually.

include/core/federated/net_util.h
lhstrh: Here it would make sense to have it because Arduino would need to have its own implementations...

lhstrh (approved):

petervdonovan (approved):

Issue #168, opened 2023-02-25
Deadline misses are not traced


erlingrj:
I was looking at the reaction invokation lags and noticed that if there is a deadline that is missed, we do not trace it. How should a deadline miss be traced? A proposal in pseudo-code

```
tracepoint_reaction_start
tracepont_deadline_miss
reaction->deadline_func()
tracepoint_reaction_stop
```

I.e. we trace the deadline as a "normal" reaction invokation, but add a deadline-miss point?

erlingrj:
After some more consideration, I think it would suffice with a new trace point called e.g.
`tracepoint_reaction_deadline_miss(reaction, worker)`

It would then store the the logical and physical time (and thus the lag). I am not sure whehter we need to trace the beginning and start of the deadline handler

edwardalee:
Yes, I suggest that simpler is better. Just tracing the deadline miss seems sufficient to me. We can always add more tracepoints later when we need them, so let's not prematurely clutter the traces.

PR #169, opened 2023-02-26
Add new tracepoint for deadline misses

erlingrj:
Solves #168 By doing the following:
1. Introducing a new tracepoint called `reaction_deadline_missed`.
2. Adding a tracepoint in reactor.c reactor_common.c and reactor_threaded.c when a deadline violation is detected.

We do NOT trace it in the `lf_check_deadline` API function, even though it can trigger the execution of the deadline handler. This is because, semantically, it is not a deadline miss if it expires during reaction execution.


lhstrh (approved):
LGTM!

Issue #170, opened 2023-02-28
Splitting rti.c to lib_rti.c for reuse of rti.c functions.


Jakio815:
I'm planning to make a `secure_rti` executable, using `rti.c`'s functions. But I see the current code is putting all functions in `rti.c`, which has a `main()`, so the functions in the code is not reusable. What do you think about splitting up the code to a different `lib_rti.c` with a header `lib_rti.h`, and rti.c does `#include "lib_rti.h"`? @lhstrh @hokeun

lhstrh:
Yes, that is a good idea.

edwardalee:
I suggest naming things rti_secure and rti_lib (I.e., use "rti" as a prefix not a suffix).

Jakio815:
Thanks for the feedback @edwardalee. I fixed it. It's working on #172.

PR #143, opened 2023-01-18
Draft: Add SST options for security in federated executions.

Jakio815:
This PR is related to lf-lang/lingua-franca#1553. Must be merged together.

PR #172, opened 2023-03-01
Split up functions of rti.c to rti_lib.c for reuse of functions.

Jakio815:
This PR is about #170.

The original `rti.c` and `rti.h` #include each other files. I made a common `lib_rti.c` which includes `rti_lib.h`. `rti.c` with a `main()` function now only includes `rti_lib.h`.

The `rti_lib.c` is an exact copy of `rti.c` only excluding `main()`. `rti_lib.h` is also a copy of `rti.h`, and added declarations of functions in `rti_lib.c`. This is for reuse of the functions in `rti_lib.c` to make a `rti_secure` which is working on #143.


lhstrh (approved):
Looks good!

PR #174, opened 2023-03-06
Code in `rti.c` made available as library

Jakio815:


Jakio815:
I think it will work if the main branch reverts to commit `4f78b42a` which was the commit just before I started pushing. I also merged #169 together what you merged right after my PR. Would you please have a look? @lhstrh.

edwardalee (changes requested):
This PR also seems to loose the history. Note that rti_lib.c is entirely authored by @Jakio815 even though it is mostly copied from rti.c.

lhstrh:
The trick is to use `git mv` for the file that will carry most of the history and create new ones for the smaller ones... Whatever gets factored out, however, will not of course loose history, there is no way around that.

lhstrh:
@Jakio815 did you see the link that @erlingrj shared about doing this?

edwardalee:
You can preserve the history even on the smaller file, although it may not be worth the trouble if it is small enough. See:

https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892

Jakio815:
> 

I'm not sure what you are mentioning. Are you mentioning [this](https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892)? 

And actually when I checkout to this branch, I see the file history. I followed the link above which professor @edwardalee mentioned. I see that the Git's Files Changed tap seems to change everything, but in my VSCode GitLens extension, it has the original history. I'm not sure if this is the right way. Would you please check? @lhstrh

edwardalee:
Oh, interesting. Yes, it could be just that the history doesn't show up in Files Changed. I hadn't realized that, but it makes sense.

Jakio815:
I also see that the contributors for rti_lib.c is only me. I'm not sure how I can check this on the Github web.

Jakio815:
I found out that if you tap ['Blame'](https://github.com/lf-lang/reactor-c/blame/0e60d4b72811c011b5486f6d16278c8f9553bcb7/core/federated/RTI/rti_lib.c), you could see the history of every lines. Also, if you tap the ['History'](https://github.com/lf-lang/reactor-c/commits/84845c5391dc3f9c1f5d44cd832654d9bed3b99d/core/federated/RTI/rti_lib.c) you could check the history below saying that it was renamed.

Sorry for bothering everyone with this problem.

lhstrh:
I reverted my merge commit of the previous PR, so once approved, you can just go ahead and hit the merge button. Please mark it as "ready for review" so that @edwardalee  can approve it. Thanks!

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       cmake-args: '-DNUMBER_OF_WORKERS=4 -DLF_THREADED=1'
 
   build-rti:
-    uses: lf-lang/reactor-c/.github/workflows/build-rti.yml@main
+    uses: lf-lang/reactor-c/.github/workflows/build-rti.yml@redo-rti
```
edwardalee, line 23: I guess this should change back to 
```
uses: lf-lang/reactor-c/.github/workflows/build-rti.yml@main
```
as part of the merge into main?

edwardalee (approved):
LGTM!  Thanks for fixing this.

.github/workflows/ci.yml
Jakio815: Marten usually does it after merged to main.

Jakio815: I just done it as the tests all passed.

PR #175, opened 2023-03-07
Do not send TAG message when a federate do not have to advance its tag

byeong-gil:
This draft PR is for optimizing message exchanges in centralized federated execution. 

For more information, please look at [lingua-franca/discussions/1626](https://github.com/lf-lang/lingua-franca/discussions/1626).


core/federated/RTI/rti_lib.c
```diff
@@ -438,7 +438,9 @@ bool send_advance_grant_if_safe(federate_t* fed) {
             "(adjusted by after delay).",
             fed->id,
             min_upstream_completed.time - start_time, min_upstream_completed.microstep);
-    if (lf_tag_compare(min_upstream_completed, fed->last_granted) > 0) {
+    if (lf_tag_compare(min_upstream_completed, fed->last_granted) > 0
+        && lf_tag_compare(min_upstream_completed, fed->next_event) > 0 // The federate doesn't have to advance its tag
```
edwardalee, line 442: Shouldn't this be >= rather than > ?
Also, this only works `fed->next_event` is updated not only when a NET is received but also when a message is forwarded to the federate. I believe this does in fact happen, though the next_event field docs give no indication.

core/federated/RTI/rti_lib.c
```diff
@@ -566,7 +568,10 @@ void update_federate_next_event_tag_locked(uint16_t federate_id, tag_t next_even
         next_event_tag = min_in_transit_tag;
     }
 
-    _RTI.federates[federate_id].next_event = next_event_tag;
+    if (lf_tag_compare(_RTI.federates[federate_id].next_event, next_event_tag) != 0) {
+        _RTI.federates[federate_id].next_event = next_event_tag;
+        //FIXME: notify upstream federates
```
edwardalee, line 573: Maybe separate this change into another PR rather than putting a FIXME here in this PR?

edwardalee (changes requested):
Modulo the two questions I raised, this looks good to me. Let's make sure all the tests pass.

core/federated/RTI/rti_lib.c
byeong-gil: I moved this change into another draft PR ([reactor-c/pull/176](https://github.com/lf-lang/reactor-c/pull/176)).
Also, removed this change at this commit (67f9d53de775c85665eb6e3da01da9921016a2ea).

core/federated/RTI/rti_lib.c
byeong-gil: I change '>' to '>=' at this commit (89320fd6227cdeb600899ec07ab3544f270747a4).

Also, as I understood, your concern is that this does not work if the RTI does not update `fed->next_event` when a message is forwarded to the federate, right?
If my understanding is right, I think it will work well because the function [`update_federate_next_event_tag_locked`](https://github.com/lf-lang/reactor-c/blob/a267815b540cd397be87ab1e294f0420a6296311/core/federated/RTI/rti_lib.c#L559) updates `next_event` as the minimum of `next_event_tag` and the `min_in_transit_tag` (the minimum tag of in-transit messages to the federate).

edwardalee (approved):
LGTM!

PR #176, opened 2023-03-08
Draft: Remove unnecessary LTC messages to optimize the RTI

byeong-gil:
This draft PR tries to remove unnecessary LTC messages from federates to the RTI.
To do this, the RTI has to forward NET messages to indicate when the LTC message is needed. 

For more information, please see this discussion, [lingua-franca/discussions/1626](https://github.com/lf-lang/lingua-franca/discussions/1626).

### DIFF
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -566,7 +566,10 @@ void update_federate_next_event_tag_locked(uint16_t federate_id, tag_t next_even
         next_event_tag = min_in_transit_tag;
     }
 
-    _RTI.federates[federate_id].next_event = next_event_tag;
+    if (lf_tag_compare(_RTI.federates[federate_id].next_event, next_event_tag) != 0) {
+        _RTI.federates[federate_id].next_event = next_event_tag;
+        //FIXME: notify upstream federates
+    }
 
     LF_PRINT_DEBUG(
        "RTI: Updated the recorded next event tag for federate %d to (%ld, %u).",


byeong-gil:
I added a logic for detecting the cycle. Also, the RTI doesn't send NDTs to federates that are in a cycle. This includes not only zero-delay cycles but also every cycle. Tests are passing now.

Also, ABS messages are skipped by NDTs again in this commit [ce6ba0f](https://github.com/lf-lang/reactor-c/pull/176/commits/ce6ba0f320eefdb7f4fcbbe023bc12ecfa6c9bd8). This can be done safely as we don't apply this optimization to a cycle.

However, basically port absent messages are not needed except for zero-delay cycles. So we have to remove those messages.

Lastly, disabling NDT messages for physical actions is still remaining.

core/utils/pqueue.c
```diff
@@ -473,3 +473,63 @@ void print_event(void *event) {
...
+*/
+int tag_in_reverse_order(pqueue_pri_t thiz, pqueue_pri_t that) {
+    return lf_tag_compare(((ndt_node*) thiz)->tag, ((ndt_node*) that)->tag) == 1;
+}
+
+/**
+ * Return whether or not the tags contained by given pointers are identical.
+ */
+int tag_matches(void* next, void* curr) {
+    return lf_tag_compare(((ndt_node*) next)->tag, ((ndt_node*) curr)->tag);
```
edwardalee, line 491: This line does not look right.  It will return 0 (false) if the tags are identical.

byeong-gil: Oh, I missed it. It's fixed in the commit [3faa188](https://github.com/lf-lang/reactor-c/pull/176/commits/3faa18840cce38d89d8d833ba8ce03ba097b6606). Thank you for letting me know, @edwardalee!

byeong-gil:
This PR is substituted by https://github.com/lf-lang/reactor-c/pull/337.

PR #177, opened 2023-03-09
Runtime support for watchdogs

Benichiwa:
This PR implements necessary Watchdog code.


lhstrh:
Before we start reviewing this, @Benichiwa, could you please 1) remove all the `FIXME`s and 2) provide a summary in the description of this PR that gives an overview of the implications of your changes? Particularly a description of the changes wrt the locking behavior would be helpful for reviewers. Thanks!

erlingrj:
Would be great with a summary yes :)

Benichiwa:
The changes to reactor-c accomplish the following (summary):

- define watchdog_t struct
- add mutex field for watchdogs to the self_base_t struct
- define watchdog_function_t
- calls lf_initialize_watchdog_mutexex inside lf_reactor_c_main to initialize mutexes for reactors containing watchdogs
- implements lf_watchdog_start
- In _lf_invoke_reaction: if reaction's reactor has a watchdog mutex, lock the mutex before invoking reaction body and unlock after invocation.

core/reactor_common.c
```diff
@@ -1316,18 +1316,70 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
+        interval_t T = watchdog->expiration - lf_time_physical();
+        lf_mutex_unlock(&(base->watchdog_mutex));
+        lf_nanosleep(T);
+        lf_mutex_lock(&(base->watchdog_mutex));
+    }
+
+
+    watchdog_function_t watchdog_func = watchdog->watchdog_function;
+    (*watchdog_func)(watchdog);
+    lf_mutex_unlock(&(base->watchdog_mutex));
```
erlingrj, line 1335: When you return here, the thread_active field is still true. So there wont be another thread created. Also, shouldnt the thread keep looping inside this function and wait for a watchdog reset?

core/reactor_common.c
```diff
@@ -1316,18 +1316,70 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
 
+#ifdef LF_THREADED
+
+void* run_watchdog(watchdog_t* watchdog) {
+    watchdog->thread_active = true;
+    self_base_t* base = watchdog->base;
+
+    while (lf_time_physical() < watchdog->expiration) {
+        interval_t T = watchdog->expiration - lf_time_physical();
+        lf_mutex_unlock(&(base->watchdog_mutex));
```
erlingrj, line 1327: This doesnt seem right. The first iteration of this while loop you are not holding the mutex and can thus not unlock it. Please document how the interaction between this thread running a single watchdog and the worker threads executing reactions should look like.

erlingrj (changes requested):
Sorry for the delay, this is a good start. Well done. 

I am not sure I understand your design entirely though, it seems like you are spawning new threads each time lf_watchdog_start is called. Are you sure you want that? Also I believe there is a race condition and some mismatch of lock and unlock. It would be nice with some documentation of how you intend this to work so we can understand your design and maybe iterate on it.

I guess this is paired with a PR in the Lingua Franca repo also? Can you give a link to that as well?

core/reactor_common.c
```diff
@@ -1316,18 +1316,70 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
 /**
  * Invoke the given reaction
  *
  * @param reaction The reaction that has just executed.
  * @param worker The thread number of the worker thread or 0 for unthreaded execution (for tracing).
  */
 void _lf_invoke_reaction(reaction_t* reaction, int worker) {
+
+#ifdef LF_THREADED
+    if (&(((self_base_t*) reaction->self)->watchdog_mutex) != NULL) {
```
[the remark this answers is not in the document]
Benichiwa: Yes, I think that was actually causing an issue once I fixed the race condition associated with 'thread_active' field race condition mentioned below. I now use a boolean field added to the self_base_t struct indicating whether or not the mutex has been initialized.

core/reactor_common.c
```diff
@@ -1316,18 +1316,70 @@ bool _lf_check_deadline(self_base_t* self, bool invoke_deadline_handler) {
...
+    self_base_t* base = watchdog->base;
+
+    watchdog->expiration = lf_time_logical() + watchdog->min_expiration + additional_timeout;
+
+
+    if (!watchdog->thread_active) {
+        lf_thread_create(&(watchdog->thread_id), run_watchdog, watchdog);
+    } 
+
+    watchdog->thread_active = false;
```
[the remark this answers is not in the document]
Benichiwa: Should be fixed now.

core/reactor_common.c
Benichiwa: > This doesnt seem right. The first iteration of this while loop you are not holding the mutex and can thus not unlock it. 

Yep, should be fixed now.

> Please document how the interaction between this thread running a single watchdog and the worker threads executing reactions should look like.

Not sure if there is a better way of drawing this out:

- `_lf_invoke_reaction` locks the mutex if the reaction is part of a reactor that contains a watchdog
- `_lf_invoke_reaction` unlocks the mutex after invoking the reaction
- `_lf_watchdog_start` unlocks the mutex only after creating the thread for the watchdog
- `run_watchdog` locks the mutex that was just unlocked by `_lf_watchdog_start`
- `run_watchdog` unlocks->locks the mutex inside the while loop so that other reactions in the reactor can run while the thread is asleep.
- `run_watchdog` unlocks the mutex after executing the watchdog function

Benichiwa:
This PR is paired with the following from the "lingua-franca" repo: https://github.com/lf-lang/lingua-franca/pull/1605

core/reactor_common.c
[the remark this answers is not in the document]
erlingrj: I think have a `lf_mutex_t *`, i.e. a pointer to a mutex, would be better. Then its either set to NULL, or initialized with `lf_mutex_create` during init

core/reactor_common.c
erlingrj: I am pretty sure we wouldnt want that. Creating a thread has a significant overhead. It would be better to have a single thread per watchdog. But you do want it to do an actual sleep while it is waiting, not a busy-sleep

core/reactor_common.c
erlingrj: Would be nice to have it represented as a sequence. In what order to you expect this to happen. I think there is some flaw in your logic here. The watchdog mutex should be held whenever a reaction within that reactor is executed. I don't think it makes sense to unlock that mutex when you call `_lf_watchdog_start` from inside a reaction. Remember that a mutex has to unlocked by the same thread that locks it.

Benichiwa: Sorry for the delay. My thought process was the following:
Thread 1: If reaction is in a reactor that contains a watchdog, lock mutex - `lf_invoke_reaction`.
Thread 1: Execute reaction.
Thread 1: If reaction body calls `_lf_watchdog_start` , start thread 2.
Thread 1: Inside `_lf_watchdog_start`, unlock mutex.
Thread 2: acquire mutex, set expiration, and unlock mutex
Thread 2: sleep
Thread 1: Inside `_lf_watchdog_start`- if the mutex has been unlocked, lock mutex
Thread 1: Unlock mutex inside `lf_invoke_reaction`.
Thread 2: Once done sleeping, lock mutex, and either reset expiration or execute watchdog function.

In hindsight, I'm now realizing that I need to reset the sleep each time, or I will get undesired behavior past the first watchdog function call.

PR #178, opened 2023-03-09
Tracing federate interactions

edwardalee:
This PR, co-authored by @ChadliaJerad, adds trace points to the interactions between federates and the RTI.

It also removes the tracing mechanism's dependency on the ability to create threads and uses the platform-independent critical section API instead.


core/trace.c
```diff
@@ -454,7 +435,8 @@ void tracepoint_schedule(trigger_t* trigger, interval_t extra_delay) {
             && trigger->reactions[0] != NULL) {
         reactor = trigger->reactions[0]->self;
     }
-    tracepoint(schedule_called, reactor, 0, 0, NULL, trigger, extra_delay);
+    // FIXME: First 0 should be the worker.
```
petervdonovan, line 438: Maybe use -1 so that it does not look like a valid worker number? It looks pretty difficult to provide the correct worker number here, but maybe we can document it as an issue.

core/trace.c
```diff
@@ -521,19 +505,15 @@ void tracepoint_scheduler_advancing_time_ends() {
...
-    tracepoint(reaction_deadline_missed, reaction->self, reaction->number, worker, NULL, NULL, 0);
+    tracepoint(reaction_deadline_missed, reaction->self, NULL, worker, worker, reaction->number, NULL, NULL, 0);
 }
 
-/**
- * Flush any buffered trace records to the trace file and
- * close the file.
- */
 void stop_trace() {
+    lf_mutex_lock(&_lf_trace_mutex);
```
petervdonovan, line 512: It doesn't seem right to possibly return (if `_lf_trace_stop`) without unlocking the mutex.

core/trace.c
```diff
@@ -561,4 +541,140 @@ void stop_trace() {
...
+        partner_id,     // int dst_id,
+        NULL,   // instant_t* physical_time (will be generated)
+        NULL,   // trigger_t* trigger,
+        0       // interval_t extra_delay
+    );
+}
+#endif // FEDERATED
+
+////////////////////////////////////////////////////////////
+//// For RTI execution
```
petervdonovan, line 634: This will make it harder to eventually disentangle the RTI from `reactor-c` (IIRC, a long-standing to-do is to put the RTI in a separate repository). Can these function implementations be factored out into a file in `core/federated/RTI` that is only used in the building of the RTI?

The same comment applies to the RTI-specific changes to `trace.h`. If we disentangle these things then that might allow us to avoid having to use both `LF_TRACE` and `RTI_TRACE` since the build system would take care of linking in the right code for the given executable.

petervdonovan (changes requested):
This PR has a number of helpful cleanups such as deleting duplicated documentation -- thanks! I left a few comments on things I was not sure about.

core/trace.c
```diff
@@ -37,12 +40,16 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // FIXME: This is a hack to allow trace.c to get the threaded support API
 //  even though we are using a unthreaded runtime
-#define _LF_TRACE
+// #define _LF_TRACE
```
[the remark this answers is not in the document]
erlingrj: This is causing some of the CI failures with unthreaded programs with tracing enabled. That commented out define is to enable tracing with the unthreaded runtime. A solution would be to make unthreaded version of the tracing,  I think this is the most natural approach, it would work with e.g. the Arduino target as well. I think it will be surprising for our users that even if you specify `threading: false` we will link with a threading library if you specify `tracing: true`.

core/federated/RTI/rti.c
```diff
@@ -47,14 +47,32 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ */
+char *rti_trace_file_name = "rti.lft";
+
+int main(int argc, const char* argv[]) {
     if (!process_args(argc, argv)) {
         // Processing command-line arguments failed.
         return -1;
     }
+    if (_RTI.tracing_enabled) {
+#ifdef RTI_TRACE
```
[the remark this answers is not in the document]
ChadliaJerad: Indeed, it seems redundant. 
@edwardalee and I discuss this feature. The reason behind adding preprocessing using `RTI_TRACE` is to have a smaller memory footprint of the RTI. The issue is whether it is user-friendly to have 2 separate executables, let's say RTI and RTI-T.
This is open to discussion.

ChadliaJerad:
The list of reported tracepoints is probably consistent now.
There are two main questions though:
- Should both: the compile-time and run-time tracing checks be kept?
- Will threaded support be enabled when tracing, even though threaded target property is set to false?

edwardalee:
> The list of reported tracepoints is probably consistent now. There are two main questions though:
> 
> * Should both: the compile-time and run-time tracing checks be kept?

With the federates, the trace functions are defined as empty macros when tracing is turned off. There is no need for any further check.

With the RTI, I think we should have just one RTI executable and that the -t command-line option should specify whether tracing is on or not. So I don't think compile-time checks make sense for the RTI.

> * Will threaded support be enabled when tracing, even though threaded target property is set to false?

I think that in an unthreaded runtime, tracing can be implemented without threads. I think we can change the tracing code to use only the critical section features of platform.h, in which case, it will automatically work with both the threaded and unthreaded runtimes.  But I have not tried to do this yet, so I could be underestimating the effort.

ChadliaJerad:
> With the federates, the trace functions are defined as empty macros when tracing is turned off. There is no need for any further check. 
> 
> With the RTI, I think we should have just one RTI executable and that the -t command-line option should specify whether tracing is on or not. So I don't think compile-time checks make sense for the RTI.

Both are addressed in the two previous commits.
Still, for reducing the memory footprint, currently, RTI_TRACE is kept and used in trace.c.
Any suggestions @petervdonovan?

> I think that in an unthreaded runtime, tracing can be implemented without threads. I think we can change the tracing code to use only the critical section features of platform.h, in which case, it will automatically work with both the threaded and unthreaded runtimes. But I have not tried to do this yet, so I could be underestimating the effort.

This PR #182 makes compiling pass. The generated trace file, however, is empty.
The reason is that ` _lf_flush_trace_thread` is the thread who writes to the file. Since in an unthreaded execution (case of bare-metal), `lf_thread_create` will not do anything, then no thread is created. 

I think that writing to the trace file needs to be part of the sequential execution. 
What do you think @edwardalee and @erlingrj?

petervdonovan:
> for reducing the memory footprint, currently, RTI_TRACE is kept and used in trace.c.
Any suggestions @petervdonovan?

I still stand by my suggestion [here](https://github.com/lf-lang/reactor-c/pull/178#discussion_r1133340389) unless there is some reason why that suggestion (moving RTI-specific code to the `federated` directory instead of having it in `trace.c`) would not work.

ChadliaJerad:
> This PR #182 makes compiling pass. The generated trace file, however, is empty. The reason is that ` _lf_flush_trace_thread` is the thread who writes to the file. Since in an unthreaded execution (case of bare-metal), `lf_thread_create` will not do anything, then no thread is created.

This breaks the CI though...

ChadliaJerad:
Refactoring the tracing mechanism this way is much cleaner than in this [PR](https://github.com/lf-lang/reactor-c/pull/182). Thanks, Edward!
There is a longer overhead that appears from time to time though, due to `flush_trace()` call whenever the buffer is full.
Does it make sense to account for the execution time of  `flush_trace()` and subtract it?
The aim is to make the maximum execution time of reactions agnostic to that overhead.

core/trace.c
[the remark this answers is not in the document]
edwardalee: I have simplified the tracing code so it doesn't need threads.  There will be a performance hit when trace buffers have to be flushed to files, but that should occur rarely, and for most useful trace files, not until shutdown.

core/trace.c
```diff
@@ -360,44 +348,32 @@ void start_trace(char* filename) {
     _lf_trace_stop = 0;
 
     // In case the user forgets to stop to the trace in wrapup.
-    if (atexit(stop_trace) != 0) {
-        lf_print_warning("Failed to register stop_trace function for execution upon termination.");
-    }
+    // NOTE: Removed because this results in premature stopping of the trace,
+    // before threads have finished writing to the trace buffer.
```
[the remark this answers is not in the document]
edwardalee: The `stop_trace()` function is now invoked in `termination()` in `reactor_common.c`, which is already being invoked using at_exit(), so I think we are good here.

ChadliaJerad:
> > for reducing the memory footprint, currently, RTI_TRACE is kept and used in trace.c.
> > Any suggestions @petervdonovan?
> 
> I still stand by my suggestion [here](https://github.com/lf-lang/reactor-c/pull/178#discussion_r1133340389) unless there is some reason why that suggestion (moving RTI-specific code to the `federated` directory instead of having it in `trace.c`) would not work.

What do you think we should do here?
I think that having tracing in different files (each specific to a particular endpoint) may lead to inconsistencies when updating. 
Otherwise, we can use `FEDERATED` only.

petervdonovan (approved):

core/federated/RTI/rti.c
```diff
@@ -47,14 +47,25 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include "rti_lib.h"
+unsigned int _lf_number_of_workers = 0u;
 
 extern RTI_instance_t _RTI;
 
-int main(int argc, char* argv[]) {
+/**
+ * RTI trace file name
+ */
+char *rti_trace_file_name = "rti.lft";
```
lhstrh, line 57:
```suggestion
const char *rti_trace_file_name = "rti.lft";
```

core/federated/RTI/rti_lib.c
```diff
@@ -1488,32 +1304,19 @@ void* federate_thread_TCP(void* fed) {
...
- */
 void send_reject(int socket_id, unsigned char error_code) {
     LF_PRINT_DEBUG("RTI sending MSG_TYPE_REJECT.");
     unsigned char response[2];
     response[0] = MSG_TYPE_REJECT;
     response[1] = error_code;
     // FIXME: Ignore errors on this response.
     write_to_socket_errexit(socket_id, 2, response, "RTI failed to write MSG_TYPE_REJECT message on the socket.");
+    if (_RTI.tracing_enabled) {
+    }
```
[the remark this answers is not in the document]
lhstrh: :point_up:

core/trace.c
lhstrh: Let's resolve this before merging and not leave the FIXME.

core/trace.c
```diff
@@ -465,7 +358,8 @@ void tracepoint_schedule(trigger_t* trigger, interval_t extra_delay) {
  */
 void tracepoint_user_event(char* description) {
     // -1s indicate unknown reaction number and worker thread.
-    tracepoint(user_event, description,  -1, -1, NULL, NULL, 0);
+    // FIXME: Worker thread should be given?
```
lhstrh, line 361: Please address.

core/trace.c
```diff
@@ -480,39 +374,40 @@ void tracepoint_user_event(char* description) {
  */
 void tracepoint_user_value(char* description, long long value) {
     // -1s indicate unknown reaction number and worker thread.
-    tracepoint(user_value, description,  -1, -1, NULL, NULL, value);
+    // FIXME: Worker thread should be given?
```
lhstrh, line 377: Idem.

core/trace.c
```diff
@@ -480,39 +374,40 @@ void tracepoint_user_event(char* description) {
...
+    tracepoint(user_value, description,  NULL, -1, -1, -1, NULL, NULL, value);
 }
 
 /**
  * Trace the start of a worker waiting for something to change on the event or reaction queue.
  * @param worker The thread number of the worker thread or 0 for unthreaded execution.
  */
 void tracepoint_worker_wait_starts(int worker) {
-    tracepoint(worker_wait_starts, NULL, -1, worker, NULL, NULL, 0);
+    tracepoint(worker_wait_starts, NULL, NULL, worker, worker, -1, NULL, NULL, 0);
```
lhstrh, line 386: This looks confusing. Why are we passing `worker` along with `worker-1`?

core/trace.c
lhstrh: I agree with Peter. Let's handle this now, before merging.

lhstrh (changes requested):
This looks good overall, but there are some minor unaddressed issues. Once fixed, let's merge.

edwardalee:
> > I think that having tracing in different files (each specific to a particular endpoint) may lead to inconsistencies when updating.
> 
> This is true. There are pros and cons.
> 
> I think its fine to consider this a matter of opinion and keep it as-is for now. We can always change this later.

As a practical matter, separating out the code for tracing of federated events is difficult because we would have to somehow split the enemy’s that define event types.  I’m sure it could be done, but I think it will complicate maintenance of the code and introduce opportunities for errors.  So i vote for keeping the code together.

edwardalee:
> Refactoring the tracing mechanism this way is much cleaner than in this [PR](https://github.com/lf-lang/reactor-c/pull/182). Thanks, Edward!
> There is a longer overhead that appears from time to time though, due to `flush_trace()` call whenever the buffer is full.
> Does it make sense to account for the execution time of `flush_trace()` and subtract it?
> The aim is to make the maximum execution time of reactions agnostic to that overhead.

Oh yes, good point. It may not be necessary to subtract it, however.  What if instead we ensure that the physical time of a tracepoint is always recorded after the flush for start events (reaction start, etc) and before the flush for end events?  This suggests one more argument to the tracepoint function.

lhstrh:
@ChadliaJerad, when a concrete code suggestion is given, you can also just click on "Commit suggestion" and save the time of copying the code, committing, and pushing. Also, once a comment has been addressed, please hit the "Resolve conversation" button (unless, of course, there is still more conversation to be had about the issue at hand). Thanks!

petervdonovan:
The comment

> I agree with Peter. Let's handle this now, before merging.

relates to [this discussion](https://github.com/lf-lang/reactor-c/pull/178#issuecomment-1480660930) on RTI-specific code in `trace.c`.

I read your comment and can kind of see why the change I requested might be difficult. I would like to avoid having a long discussion about this. If it turns out to be a problem later, then someone else can submit another PR.

[Also, I think this minor comment might still be unresolved.](https://github.com/lf-lang/reactor-c/pull/178/files/edee7f18d026ee2f289516f48d666c79c6a2aefb#r1133338232)

core/trace.c
edwardalee: This has been resolved with commit #684e354

lhstrh (approved):
FYI: the failing test is due to the runner not being able to download Gradle. This problem comes up with Windows runners because its caching is broken (see https://github.com/lf-lang/lingua-franca/issues/1664)

PR #179, opened 2023-03-10
Add compile definitions for federated.

petervdonovan:
Companion PR: https://github.com/lf-lang/lingua-franca/pull/1622


petervdonovan:
Do you have any idea what is going on here @erlingrj ? The Zephyr test is the only one that this branch has not been able to pass. I did not make changes related to Zephyr tests, and I failed to reproduce the problem locally using Zephyr version 3.2.99 (after changing the generated CMakeLists to let me try 3.2.99). I am also re-running CI on `main` to check whether the changes from the last 4 days have anything to do with the test failure.

erlingrj:
Should be fixed in master now. The problem was that I was using a Zephyr docker image in CI and I used the `latest` version. Yesterday some new changes were pushed there which were not compatible with our CI. Lesson learned, use explicit versions...

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
   lf-default-zephyr:
     needs: fetch-lf
-    uses: lf-lang/lingua-franca/.github/workflows/c-zephyr-tests.yml@master
+    uses: lf-lang/lingua-franca/.github/workflows/c-zephyr-tests.yml@c826449228978bd8ba12630f0a4124d4cebe98e7
```
petervdonovan, line 46:
```suggestion
    uses: lf-lang/lingua-franca/.github/workflows/c-zephyr-tests.yml@master
```

PR #182, opened 2023-03-16
Fix compile error when unthreded execution and tracing are both enabl…

ChadliaJerad:
This PR fixes the compile error due to tracing in an unthreaded execution. 
Currently, the .lft file is generated, but w/o contents.
Writing to the trace file is WIP.

### DIFF
--- a/core/trace.c
+++ b/core/trace.c
@@ -356,6 +356,14 @@ void start_trace(char* filename) {
 
     lf_thread_create(&_lf_flush_trace_thread, flush_trace, NULL);
 
+#ifdef LF_UNTHREADED 
+    // FIXME: If LF_UNTHREADED is defined, then no thread will be create to 
+    //        flush the file (instruction above).
+    //        A sequential mechanism is needed for emptying the buffer into the
+    //        file when needed.
+    ; // WIP
+#endif // LF_UNTHREADED
+
     LF_PRINT_DEBUG("Started tracing.");
 }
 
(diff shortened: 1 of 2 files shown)


ChadliaJerad:
Oops, the PR breaks Arduino, Zephy... tests...
At the same time, replacing:
```
typedef void* lf_mutex_t;
typedef void* lf_cond_t;
typedef void* lf_thread_t;
```
with:
```
typedef void lf_mutex_t;
typedef void lf_cond_t;
typedef void lf_thread_t;
```
unfixes the compilation of `test/C/src/CountTest.lf`, when `threading` is set to `false`... 

Any hint on the usefulness of `_LF_TRACE`?

ChadliaJerad:
The failing tests seem related to [this PR](https://github.com/lf-lang/lingua-franca/pull/1660) though.

edwardalee (commented):
This does not feel like the right solution to me.  Ideally, platform.h would not need to know anything at all about tracing. That file shouldn't mention tracing.

One possible way to solve the problem that the trace file does not get written is that instead of just ignoring the call to create a thread, the unthreaded runtime could execute the function passed in the current thread.

erlingrj:
Maybe the easiest would be to start by *disabling tracing for uthreaded runtime*, then we need to figure out how to adapt `trace.c` to either:
A) Not use the threaded API but use some other abstraction (e.g. enter/exit critical section)
B) Conditionally include code with the preprocessor based on THREADED/UNTHREADED

ChadliaJerad:
Closed as solved in #178.

PR #183, opened 2023-03-22
Added lf_wait_for_event()

edwardalee:
In the platform.h definitions, we have `lf_critical_section_enter()`, `lf_critical_section_exit()`, and `lf_notify_of_event`. But missing from this is `lf_wait_for_event()`.  This PR adds that function, including implementations for Zephyr, Arduino, and nrf52, although I'm not entirely sure those implementations are correct.

In lingua-franca, there is a corresponding branch `wait-for-event` that adds two tests, `test/C/src/WaitForEvent.lf` and `test/C/src/WaitForEvent.lf2`.  I've verified that these tests work for threaded and unthreaded runtimes.

The reason for adding this is that we can make tracing work for all platforms if we add this function, I think.

### DIFF
--- a/core/platform/lf_arduino_support.c
+++ b/core/platform/lf_arduino_support.c
@@ -175,6 +175,14 @@ int lf_notify_of_event() {
    return 0;
 }
 
+int lf_wait_for_event(interval_t timeout) {
+    int result = lf_sleep(timeout);
+    if (result) {
+        return LF_TIMEOUT;
+    }
+    return 0;
+}
+
 #else
 #warning "Threaded support on Arduino is still experimental"
 #include "ConditionWrapper.h"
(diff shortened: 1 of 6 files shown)


erlingrj:
I don't see immediately how this is different than `lf_sleep_until_locked`. The semantics of this function is to sleep until either:
1) You reach the specified time ( `timeout` in your case) or
2) `lf_notify_of_event` is invoked (i.e. an event 

So, in the same  way as `reactor_threaded.c` implements it using the platform API, `reactor.c` could implement it using the single threaded API (with `lf_sleep_until_locked`).

Or am I missing something?

edwardalee:
Never mind. It turns out tracing doesn't really need this, so let's not insert additional complexity.

PR #190, opened 2023-03-31
[C-Generics] Helper Macros

mkhubaibumer:
This PR contains two headers
 - api/generics.h (provides convenient wrappers around builtins to infer-type based on assignment and type conditionals)
 - utils/converter.h (provides a unified interface to call type-converters if a converter library is provided and defined convention is followed


petervdonovan:
In order to merge this, it might be good to know why these files cannot be maintained as a separate project from `reactor-c`.

It has been explained that these files are generally useful and that they likely will be used by anyone who uses generics, but I am not sure that this will turn out to be true. It depends on the idioms that emerge for dealing with generics on the C target, and I am not sure that we are ready yet to commit to/endorse a particular set of idioms. Also, if files of this sort are in `reactor-c`, then I am worried that it will become the LF maintainers' responsibility to maintain them, to limit backwards-incompatible changes to them, to think about and discuss what is in them, etc.

edwardalee (commented):
These look pretty useful to me, but I have question: How portable are the builtin functions (or macros)?  Will this create problems for embedded platforms?

mkhubaibumer:
> These look pretty useful to me, but I have question: How portable are the builtin functions (or macros)? Will this create problems for embedded platforms?

@edwardalee These builtins are available for both clang and gcc
As far as embedded platforms is concerned this will be a nop and the user will have to change their code if builtins are not defined.


But for the platforms that they are defined this makes the life very easy and we have been using them in our code-base heavily

@petervdonovan this is a basic set of features supported by the c-compiler to make things generic and is purely optional



We can share a few examples of these macros in action if you want but they pack a lot of power to be ignored

edwardalee (changes requested):
I agree that these be useful in user-written reaction code. But we need to keep use of these out of the reactor-c codebase if we want to preserve the ability to run on embedded platforms.  So I think a better place to put them would be in the `util` directory, where there are other such user-level utilities.

petervdonovan:
Because these are header files, I do not think that they will cause problems if they are not `#include`d. But I do agree that the `util` directory would be a good place for them.

edwardalee:
I am more worried about people starting to use them in the core. The build system cannot fix that.

petervdonovan:
I also agree that it is not a big deal to have to add something to the target properties. This is C; people are responsible for making sure their projects build. Later, we can even let Lingo generate a configuration file that adds these files by default if we want.

mkhubaibumer:
> I am more worried about people starting to use them in the core. The build system cannot fix that.

@edwardalee This generics.h can be inclosed in a `#ifdef REACTOR_CODE` block  and while generating reactor code we can define `REACTOR_CODE` symbol for all reactors
I think that can solve this issue

lhstrh:
@mkhubaibumer, it seems that the easiest way to address @edwardalee's concern is to just move the files to `util`. Could you take care of this? After that we should be ready to merge.

edwardalee:
I have opened an issue #1797 for the test failure, which is unrelated to this PR.

PR #192, opened 2023-04-13
Transient federates

ChadliaJerad:
This is PR enables the support of transient federates, introduced in this [discussion](https://github.com/lf-lang/lingua-franca/discussions/1504).
This is still a work in progress. The PR is meant for checking backward compatibility using the CI.

This branch works with `transient-federates` branch of `lingua-franca`. Currently:
 - every federate instance stores its persistence status (`is_transient`), and records its own `start_time`.
 - The `RTI` distinguishes the startup phase (where all persistent federates have to join in order to start the execution), from the one after, where only transient federates can join and leave at arbitrary times.
 - when a transient joins during the startup phase, its `start_time` is set along with the persistent federates. If, however, it joins after the startup phase, its start_time will be:
    1. either the physical time at which it joined, if it has no upstream or downstream federates,
    2. or the maximum among the Next Event Tags (ignoring the microstep) of its upstream and downstream federates, otherwise.
 - `test/C/src/federated/transient/Timers.lf` showcases 1.
 - An example that showcases 2. is WIP.

The semantics will further be detailed in [discussion](https://github.com/lf-lang/lingua-franca/discussions/1504).


ChadliaJerad:
Closed as replaced with https://github.com/lf-lang/reactor-c/pull/358.

PR #193, opened 2023-04-13
Documentation and code cleanup

edwardalee:
This is a starting point to clean up the APIs and their docs for user-facing APIs. It includes a number of code organization changes, including removing variables and functions from user scope. PR [#1696](https://github.com/lf-lang/lingua-franca/pull/1696) should be merged first, or else some tests will fail.

This first step focuses mainly on tag.h and tag.c.  The generated doxygen docs are much better, now suitable for pointing users to them.  This PR is already big enough that I think we should merge this incrementally.  I.e., merge this now, then continue by picking the next part of the API, such as target.h, to work on.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,5 +1,5 @@
 All files in this repository are copyrighted by their respective authors
-and/or institutions (as identified in each file) and bear a BSD 2-Clause License:
+and/or institutions and bear a BSD 2-Clause License:
 
 Copyright (c) 2021-2023, by the Authors
 
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,7 @@
-BSD 2-Clause License
+All files in this repository are copyrighted by their respective authors
+and/or institutions (as identified in each file) and bear a BSD 2-Clause License:
 
-Copyright (c) 2021, The Lingua Franca Coordination Language
+Copyright (c) 2021-2023, by the Authors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -2,7 +2,6 @@ All files in this repository are copyrighted by their respective authors
 and/or institutions (as identified in each file) and bear a BSD 2-Clause License:
 
 Copyright (c) 2021-2023, by the Authors
-All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/README.md
+++ b/README.md
@@ -4,32 +4,21 @@
 # Reactor-C: A reactor runtime implementation in C
 
 ## Documentation
-To generate and view documentation, follow the following steps:
-- Install `python3`, `pip3` and `doxygen`
-- Install the required Python modules:
-  - `pip3 install sphinx`
-  - `pip3 install sphinx_sitemap`
-  - `pip3 install sphinx-rtd-theme`
-  - `pip3 install breathe`
-  - `pip3 install exhale`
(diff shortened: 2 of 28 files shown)


petervdonovan (approved):
Thank you for deleting the duplicated documentations and some of the unconventional uses of header files. It is nice to see the number of LOC go down a bit. I also agree that it is good to remove non-API things from `tag.h`; this was a significant missing part of #189.

LICENSE.md
```diff
@@ -1,7 +1,7 @@
-BSD 2-Clause License
+All files in this repository are copyrighted by their respective authors
```
lhstrh, line 1: I think that's the formal use of the `LICENSE.md` file is to apply to the entire repo, so I'm not sure we need to state this explicitly. That said, is there any reason to deviate from the `LICENSE.md` in the `lingua-franca` repo? There, we simply state:
```
This software is licensed under the BSD 2-Clause License.

Copyright (c) 2019-2023, the Lingua Franca contributors.
All rights reserved.

(...)
```

lhstrh (commented):
These changes look good, and I'm a big fan of removing these long license notices at the beginning of files. I think we should stick to a single `LICENSE.md` that we can use in all of our repos, however.

LICENSE.md
edwardalee: The copyright should be held by "authors" not "contributors" (those identified by @author tags).  For a book, for example, a copy editor does not hold the copyright.  So if it's OK with you, I will also update the LICENSE.md file in lingua-franca.  Also, "contributors" could include those who provide financial support.

lhstrh: OK, that's a good point. Let me have another look at the wording before we also change the other one.

lhstrh: Maybe we can just keep the default
```
BSD 2-Clause License
```
and then have something like:
```
Copyright (c) 2021-2023, the authors as specified in each individual file in this repository.
All rights reserved.
```

lhstrh: But we might also just keep it simple and write
```
Copyright (c) 2021-2023, the authors.
```
because if you commit code but forget to add yourself as an author, you would still be a copyright holder.

edwardalee: Just committing code does not make you an author. Again, think of a copy editor.

Also, copyright and license are two different things.  Copyright holders are the people authorized to grant a license. The LICENSE.md does not need to bear a copyright notice (and in fact, the copyright notice would only apply to the text in that file, which I don't think we can claim copyright to).

lhstrh: When you select BSD-2 License on GitHub, it creates a default text that includes a Copyright notice. That's what a BSD-2 license looks like -- _it includes a copyright notice_. I agree that this is strange, but if we don't want that we need to select another license and not use BSD-2. If the BSD-2 template has a copyright notice in it, and the use of the template is for copyright owners to enter their name after the copyright statement, then there it makes no sense to interpret the copyright statement to apply to the statement itself. We didn't write that statement, so we can't copyright it.

A copyright notice also isn't necessary in order to be able to claim copyright. If I can prove that I authored something (e.g., by having sent something in a sealed envelope to myself) or checked in code to GitHub, I can claim copyright over that. I think the only utility this particular copyright statement has is to make it clear that contributors aren't signing away their right when they check code into our repository.

What problem can you identify with the statement `Copyright (c) 2021-2023, the authors`?

I recall that we have had a similar conversation a while back, and that resulted in the `LICENSE.md` that's checked into `lingua-franca`...

edwardalee: Ok, let's go with that.

PR #194, opened 2023-04-14
Switch to more general thread creation in Zephyr support

siljesu:
lf_thread_create() uses a static NUMBER_OF_THREADS instead of a static NUMBER_OF_WORKERS. NUMBER_OF_THREADS will be decided by both number of workers and user threads in the application. When federated support is added, this macro will be expanded further (i.e. clock synchronization thread, socket listening threads, etc.).

I added a locking/unlocking of a mutex to lf_thread_create() for thread safety.

I have not added a way to actually set the USER_THREADS macro, so as of yet it is set to a default value of 1.


edwardalee (approved):
LGTM, modulo the FIXME indicating that there needs to be a way to define USER_THREADS. What is the plan for this?

erlingrj:
We could add user_threads as a target property under `platform`. I.e.

```
target C {
  platform : {
    name: Zephyr,
    board: qemu_cortex_m3,
    user_threads: 3
  }
}
```

This is however a temporary fix until we have separated out the target properties with `lingo`.

erlingrj:
Will you have the chance to add this target property to `lingua-franca` @siljesu ?

siljesu:
I have added it now. See newest commit and PR in lingua-franca repo: https://github.com/lf-lang/lingua-franca/pull/1721.

I also changed the default value to 0, as this makes more sense.

erlingrj (approved):
LGTM!

PR #195, opened 2023-04-14
Fix for Arduino to account avoid duplicate definition of timespec

arengarajan99:


petervdonovan (approved):
This is harmless, but I don't understand why it is necessary. This entire file already has an include guard. Is there another file that also defines `timespec`?

lhstrh:
AFAIK, there is one in the Arduino library that is used with some platforms.

PR #201, opened 2023-04-26
Memory reporting

edwardalee:
This simple PR moves the check for memory leaks in token to after where template tokens are freed. When token types were made persistent, we started getting warnings about unfreed tokens. But these warnings are spurious. This PR eliminates those warnings.


lhstrh (approved):
These changes look good to me, but the scope of this PR is larger than the name suggests.

edwardalee:
> These changes look good to me, but the scope of this PR is larger than the name suggests.

Rats. This PR seems to have accidentally included changes in the file-access branch.  I guess that PR needs to be merged first.

PR #203, opened 2023-04-27
Reactor local time

erlingrj:
This PR makes the logical tag a reactor-level variable (not system-level). This means each reactor can be at different tags. This is needed for LET scheduling and static scheduling.

The changes are mainly:
- Update API with a pointer to the a `self_base_t` for the timing functions. `lf_tag()` and `lf_time_logical`.
- Ensure backwards compatibility with macros 
- Update any calls to: `lf_tag` and `lf_time_logical` throughout the runtime. Either they must be passed a pointer to the reactor from which they are invoked. Or passed NULL to get the system-level logical time

### DIFF
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -302,9 +302,6 @@ int next(void) {
 
 /**
  * Stop execution at the conclusion of the next microstep.
- * FIXME: If this is called from a LET reaction we will either:
- * A) requrest a stop tag "in the past" if we are using reactors local tag
- * B) request stop sometime "in the futre" if we use global tag <- This is what we are doing now
  */
 void lf_request_stop() {
 	tag_t new_stop_tag;
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -249,7 +249,7 @@ int next(void) {
         next_tag.time = event->time;
         // Deduce the microstep
         if (next_tag.time == current_tag.time) {
-            next_tag.microstep = lf_tag().microstep + 1;
+            next_tag.microstep = current_tag.microstep + 1;
         } else {
             next_tag.microstep = 0;
         }
@@ -302,6 +302,9 @@ int next(void) {
 
 /**
  * Stop execution at the conclusion of the next microstep.
+ * FIXME: If this is called from a LET reaction we will either:
+ * A) requrest a stop tag "in the past" if we are using reactors local tag
+ * B) request stop sometime "in the futre" if we use global tag <- This is what we are doing now
  */
 void lf_request_stop() {
 	tag_t new_stop_tag;
--- a/core/tag.c
+++ b/core/tag.c
@@ -15,7 +15,10 @@
 #include "tag.h"
 #include "util.h"
 #include "platform.h"
+#include "reactor.h"
 #include "util.h"
+#include "lf_types.h"
+
 
 /**
  * An enum for specifying the desired tag when calling "lf_time"
@@ -148,8 +151,14 @@ instant_t _lf_time(_lf_time_type type) {
 
 ////////////////  Functions declared in tag.h
 
-tag_t lf_tag() {
-    return current_tag;
+tag_t lf_tag(void* self) {
+    if (self == NULL) {
+        return current_tag;
+    } else if (((self_base_t *) self)->executing_reaction) {
+        return ((self_base_t *) self)->current_tag;
+    } else {
+        return current_tag;
+    }
 }
 
 int lf_tag_compare(tag_t tag1, tag_t tag2) {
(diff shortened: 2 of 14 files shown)


erlingrj:
@lsk567 this PR introduces a local tag to each reactor, this is required for the static scheduler also.

core/tag.c
```diff
@@ -148,8 +151,14 @@ instant_t _lf_time(_lf_time_type type) {
...
-tag_t lf_tag() {
-    return current_tag;
+tag_t lf_tag(void* self) {
+    if (self == NULL) {
+        return current_tag;
+    } else if (((self_base_t *) self)->executing_reaction) {
+        return ((self_base_t *) self)->current_tag;
+    } else {
+        return current_tag;
+    }
```
edwardalee, line 161:
```suggestion
tag_t lf_tag(void* self) {
    if (self != NULL && ((self_base_t *) self)->executing_reaction != NULL) {
        return ((self_base_t *) self)->current_tag;
    } else {
        return current_tag;
    }
```

include/api/api.h
```diff
@@ -185,6 +185,6 @@ int lf_tag_compare(tag_t tag1, tag_t tag2);
 /**
  * Return the current tag, a logical time, microstep pair.
```
edwardalee, line 186: This documentation needs enhancement. What does the argument mean? What does a NULL argument mean?

edwardalee (changes requested):
This looks very useful moving forward. However, it is not clear to me what a NULL argument means when calling the tag/time functions. Can you clarify?

erlingrj:
> This looks very useful moving forward. However, it is not clear to me what a NULL argument means when calling the tag/time functions. Can you clarify?

Thanks for feedback. I will clarify the documentation. When calling tag/time functions with a NULL argument then the global tag/time is returned. So there still is a notion of a global time which exists outside of any particular reactor. This global tag is what is advanced when popping a new event of the event queue and this global tag is copied into the reactor self-structs when executing a reaction

core/tag.c
```diff
@@ -148,8 +151,14 @@ instant_t _lf_time(_lf_time_type type) {
 
 ////////////////  Functions declared in tag.h
 
-tag_t lf_tag() {
```
lhstrh, line 151: It seems more natural to me to leave this function as it was. With your change, there are lots of places where it ends up being called with `NULL` as an argument, which is kind of ugly. Why not reference `self->current_tag` directly with a local time is needed? It's not less readable than `lf_tag(self)`.

lhstrh: I see that there is a dependency on whether there currently is an executing reaction. If you want to include that logic as a convenience, then maybe it's better to create a separate function called `lf_local_tag` that takes a reference to the self struct?

edwardalee: > I see that there is a dependency on whether there currently is an executing reaction. If you want to include that logic as a convenience, then maybe it's better to create a separate function called `lf_local_tag` that takes a reference to the self struct?

I don't think there is any semantic meaning to a global tag, so I don't think this is a good idea. I think that really there is *only* a local tag. It just so happens that we have had in the past a phenomenon that the local tag is advanced together for all reactors on the same platform.  This is not the case for federated, nor will it be the case for enclaves or LET.

My suggestion is to leave this as is, and document that users should *never* pass NULL to these functions, and that in the future, this will result in an error. I think all the places where NULL is passed now are internal, and to implement enclaves, these will have to replaced with non-NULL calls.

core/reactor.c
```diff
@@ -302,6 +302,9 @@ int next(void) {
 
 /**
  * Stop execution at the conclusion of the next microstep.
+ * FIXME: If this is called from a LET reaction we will either:
+ * A) requrest a stop tag "in the past" if we are using reactors local tag
+ * B) request stop sometime "in the futre" if we use global tag <- This is what we are doing now
```
erlingrj, line 307:
```suggestion
```

lsk567:
Thanks for working on this, @erlingrj. I am in favor of the changes you made here. One suggestion on reducing the "ugliness" pointed out by @lhstrh is to define a macro `LF_GLOBAL_TAG`.
```C
#define LF_GLOBAL_TAG NULL
```
Then in the documentation, we can instruct the LF maintainers to pass in `LF_GLOBAL_TAG` if the global tag is needed. Having`lf_tag(LF_GLOBAL_TAG)` instead of `lf_tag(NULL)` should make the code more readable.

edwardalee:
> Thanks for working on this, @erlingrj. I am in favor of the changes you made here. One suggestion on reducing the "ugliness" pointed out by @lhstrh is to define a macro `LF_GLOBAL_TAG`.

If we do this, I think it should be carefully documented that this only makes sense for runtimes that have barrier synchronizations at tag advancement. The existence of a global tag is transitory at best, or, more realistically, already non-existent, since there is no such thing for federated execution.

lhstrh:
My response was in reaction to the presence of a lot of calls with a `NULL` argument. If the answer is, like @edwardalee suggested, that we shouldn't have these calls and that they will disappear, then that fully addresses my concern.

include/api/api.h
erlingrj: DOne

erlingrj:
I have applied suggestions and updated docs of `lf_tag`.

erlingrj:
Inside reaction-bodies, the users will just use `lf_tag` which is a macro expanded to `lf_tag(self)` which works cause `self` is in scope. If the user wants to get the current tag inside a function somewhere they would need to pass that self reference manually. I don't think there is a scenario where the user would want to pass in `NULL`. Maybe if `lf_tag` was called asynchronously from an ISR.

lhstrh:
@erlingrj note that this has gotten out of sync and incurred a conflict. Other than that, I think this is ready to merge?

erlingrj:
> @erlingrj note that this has gotten out of sync and incurred a conflict. Other than that, I think this is ready to merge?

We need to revise this wrt to the new concept of an environment which holds the current tag.

erlingrj:
I am gonna close it for now

PR #204, opened 2023-04-28
Added `LF_PACKAGE_DIRECTORY`

edwardalee:
Companion to PR in lingua-franca.


edwardalee:
> I think we should call this `LF_PACKAGE_DIRECTORY` instead. To me, the definition of "project" is muddy currently, but the notion of "package" is clear.

I guess if it's clear to you, it's worth changing, so I've done that. To me, both terms are equally mysterious.

lhstrh (approved):

PR #206, opened 2023-05-06
Attempt to suppress "has no symbols" warnings.

petervdonovan:
These warnings appear on macOS because we use the preprocessor to conditionally include the contents of entire source files.

Since I do not use macOS it is difficult for me to check whether this solves the problem, but I think it should.


edwardalee (commented):
This looks like the right idea, but it doesn't seem to work for me. ??  I still see the warnings.

lhstrh:
Subsumed by #214.

PR #208, opened 2023-05-14
Revised Scheduler API

erlingrj:
This is still a draft PR. Docs have not been updated. Just want to run the CI.
---
TODO:

- [x] Create a common "main-loop" for both threaded and unthreaded execution
- [x] Move reaction-chain optimization out of `reactor_common.c`
- [x] Re-implement chain optimization inside the NP scheduler
- [ ] Support chain optimization in single-threaded runtime
- [ ] Add solid documentation for all new functions and logic
- [ ] We want to avoid duplicating the chain optimization. Find a way to abstract it behind a common API that all the schedulers can use.

---

This PR addresses the issue that the scheduler API is not really "consistent". The API is:

1. `lf_sched_get_ready_reaction`
2. `lf_sched_done_with_reaction`
3. `lf_sched_trigger_reaction`

But due to an optimization there can be executed reactions without a corresponding call to `lf_sched_get_ready_reaction`. This happens if a reaction produces an output that enables a single other reaction. Then the runtime will go ahead and execute this reaction right away. This is problematic for the LET scheduler and the static scheduler where we would like to have complete control over which worker executes which reaction. It should be a scheduler optimization because whether or not you should execute this downstream reaction depends on its deadline relative the deadline of the first reaction on the reaction queue.

This proposal removes this "chain optimization" from `reactor_common.c` and puts it in the scheduler. To simplify this we also unify the threaded and the unthreaded runtime. I have created a common `lf_main_loop` which both the workers of the threaded and the unthreaded runtime should run. It depends on a few functions which both threaded and unthreaded runtime must implement.


erlingrj:
I am interested in feedback on this before I proceed, @lhstrh @edwardalee .
The main thing is:
1. Removing chain optimization from `reactor_common.c`
2. Creating a common `lf_main_loop` for both the threaded and unthreaded runtime which uses a small API which must be implemented by both.
3. Re-implementing the chain-optimzation behind the scheduler API (and eventually in reactor_common.c). Currently I re-implemented the chain optimization inside the `scheduler_NP.c`, but I think we should try to avoid reimplementing it inside each scheduler, rather it could live in `scheduler_common.c` maybe or even `scheduler_chain.c`

lhstrh:
> I am interested in feedback on this before I proceed, @lhstrh @edwardalee . The main thing is:
> 
>     1. Removing chain optimization from `reactor_common.c`
> 
>     2. Creating a common `lf_main_loop` for both the threaded and unthreaded runtime which uses a small API which must be implemented by both.
> 
>     3. Re-implementing the chain-optimzation behind the scheduler API (and eventually in reactor_common.c). Currently I re-implemented the chain optimization inside the `scheduler_NP.c`, but I think we should try to avoid reimplementing it inside each scheduler, rather it could live in `scheduler_common.c` maybe or even `scheduler_chain.c`

All the above sounds good to me...

edwardalee:
Do we know how to make the chain optimization work?

erlingrj:
> Do we know how to make the chain optimization work?

Yes. With the new scheduler API, if there is only a single call to `lf_sched_enable_downstream_reaction` from a worker in between calling `lf_sched_get_ready_reaction` and the call to `lf_sched_reaction_done`, then, we have a chain and the next call to `lf_sched_get_ready_reaction` can return that enabled reaction, and it doesn't have to go through the reaction queue.

This change was motivated by the LET scheduler. With our focus shifted to enclaves, I don't see the immediate urgency to finish this now

edwardalee:
Ah, OK, I misinterpreted "chain optimization" to mean something to do with "chain IDs," which I think we still don't know how to make work.

lhstrh:
We know how to make a correct assignment but we have no algorithm for making the assignment in polynomial time...

PR #209, opened 2023-05-15
Support for watchdogs

edwardalee:
This PR, co-authored by @Benichiwa , is a companion to [PR 1730 in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/1730), which adds watchdog support for the C target. It also factors out the watchdog support into its own .c and .h files.


core/reactor_common.c
```diff
@@ -1292,17 +1300,32 @@ trigger_handle_t _lf_schedule_int(lf_action_base_t* action, interval_t extra_del
...
+
  * Invoke the given reaction
  *
  * @param reaction The reaction that has just executed.
  * @param worker The thread number of the worker thread or 0 for unthreaded execution (for tracing).
  */
 void _lf_invoke_reaction(reaction_t* reaction, int worker) {
+
+#ifdef LF_THREADED
+    if (((self_base_t*) reaction->self)->reactor_mutex != NULL) {
```
lhstrh, line 1312: We've seen a increase this kind the use of preprocessor directives inside of function bodies throughout the code base, and this has me worried about code maintainability. Here is an excerpt from the Linux kernel coding style:

> Wherever possible, don’t use preprocessor conditionals (#if, #ifdef) in .c files; doing so makes code harder to read and logic harder to follow. Instead, use such conditionals in a header file defining functions for use in those .c files, providing no-op stub versions in the #else case, and then call those functions unconditionally from .c files. The compiler will avoid generating any code for the stub calls, producing identical results, but the logic will remain easy to follow.

In short, I think we should stop using this anti pattern. A good start with this would be try to stop adding more occurrences of it, at least inside of function bodies.

core/threaded/watchdog.c
```diff
@@ -0,0 +1,83 @@
...
+extern watchdog_t* _lf_watchdogs;
+
+/**
+ * @brief Initialize watchdog mutexes.
+ * For any reactor with one or more watchdogs, the self struct should have a non-NULL
+ * `reactor_mutex` field which points to an instance of `lf_mutex_t`.
+ * This function initializes those mutexes.
+ */
+void _lf_initialize_watchdog_mutexes() {
+    for (int i = 0; i < _lf_watchdog_number; i++) {
```
lhstrh, line 23:
```suggestion
    for (int i = 0; i < _lf_watchdog_count; i++) {
```

core/threaded/watchdog.c
```diff
@@ -0,0 +1,83 @@
...
+ * @author Edward A. Lee
+ * @copyright (c) 2023, The University of California at Berkeley.
+ * License: <a href="https://github.com/lf-lang/reactor-c/blob/main/LICENSE.md">BSD 2-clause</a>
+ * @brief Definitions for watchdogs.
+ */
+
+#include <assert.h>
+#include "watchdog.h"
+
+extern int _lf_watchdog_number;
```
lhstrh, line 13:
```suggestion
extern int _lf_watchdog_count;
```

core/threaded/watchdog.c
```diff
@@ -0,0 +1,83 @@
...
+void _lf_initialize_watchdog_mutexes() {
+    for (int i = 0; i < _lf_watchdog_number; i++) {
+        self_base_t* current_base = _lf_watchdogs[i].base;
+        if (current_base->reactor_mutex != NULL) {
+            lf_mutex_init((lf_mutex_t*)(current_base->reactor_mutex));
+        }
+    }
+}
+
+/**
```
lhstrh, line 31: This comment can be removed. It already exists in the `.h` file.

lhstrh (changes requested):
This is looking pretty good! My comments are minor...

lhstrh:
Here are some more recommendations regarding the conditional compilation:

> Prefer to compile out entire functions, rather than portions of functions or portions of expressions. Rather than putting an ifdef in an expression, factor out part or all of the expression into a separate helper function and apply the conditional to that function.

> If you have a function or variable which may potentially go unused in a particular configuration, and the compiler would warn about its definition going unused, mark the definition as __maybe_unused rather than wrapping it in a preprocessor conditional. (However, if a function or variable always goes unused, delete it.)

> Within code, where possible, use the `IS_ENABLED` macro to convert a Kconfig symbol into a C boolean expression, and use it in a normal C conditional:

```
if (IS_ENABLED(CONFIG_SOMETHING)) {
        ...
}
```
> The compiler will constant-fold the conditional away, and include or exclude the block of code just as with an `#ifdef`, so this will not add any runtime overhead. However, this approach still allows the C compiler to see the code inside the block, and check it for correctness (syntax, types, symbol references, etc). Thus, you still have to use an `#ifdef` if the code inside the block references symbols that will not exist if the condition is not met.

> At the end of any non-trivial `#if` or `#ifdef` block (more than a few lines), place a comment after the `#endif` on the same line, noting the conditional expression used. For instance:
```
#ifdef CONFIG_SOMETHING
...
#endif /* CONFIG_SOMETHING */
```

This is from: https://www.kernel.org/doc/html/v4.10/process/coding-style.html.

petervdonovan (approved):

core/reactor_common.c
```diff
@@ -56,6 +56,14 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "hashset/hashset.h"
 #include "hashset/hashset_itr.h"
 
+#ifdef LF_THREADED
+#include "watchdog.h"
+
+// Code generated global variables.
+extern int _lf_watchdog_number;
```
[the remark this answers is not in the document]
edwardalee: Good point. This is changed.

core/reactor_common.c
erlingrj: Two possible solutions:
1. Implement `void lf_invoke_reaction` in both `reactor.c` and `reactor_threaded.c` which eventually calls `_lf_invoke_reaction` in `reactor_common.c`. Then we can lock the mutex in the version implement in `reactor_threaded.c`
2. Just put the mutex pointer on the self struct, also for the unthreaded runtime. This, of course, adds additional memory footprint for the unthreaded runtime which we do intend for resource constraint platforms...

lhstrh: We can address this in a separate PR as @edwardalee suggested. I've created an issue https://github.com/lf-lang/reactor-c/issues/210 for it.

core/threaded/watchdog.c
edwardalee: I don't see this comment (or function) in watchdog.h. (?)

PR #212, opened 2023-05-18
Environments

erlingrj:
This PR puts most global variables in the C-target on an environment struct. This touches almost all files and functions in the C target. Most of the changes are related to adding an `environment_t* env` argument as the first argument to all functions that modify some of the global variables.

Companion PR in Lingua-franca: https://github.com/lf-lang/lingua-franca/pull/1772

NB:
**This branch was merged with the RTI refactoring branch way back and it would be easiest to merge RTI refactoring #224  before merging this**
I.e. please dont look at the RTI changes introduced here, they will rather be merged into main with #224 and then merged into this branch 

---

### Things to consider
- All the schedulers were based on using a global variable holding the scheduler instance. This is changed by having the scheduler on the environment struct. To enable the addition of custom data structures for each scheduler there is  a `custom_scheduler_data_t*` on the scheduler instance struct. This can be typedef'ed by each scheduler so it can store additional things. It is used by e.g. the adaptive scheduler
- I have moved all the functionality for the adaptive scheduler into that file instead of using different header files for data_collection, worker_States and worker_assignments
- I took the liberty to rename many variables that had uneccesarily long names, e.g. `_lf_sched_instance->_lf_sched_num_timer_triggers` -> `scheduler->num_timer_triggers`
- How to set keepalive for each enclave is not resolved in this PR
- Watchdogs are still global. I am waiting with moving them to the environment until they are stable/working.
- In the tracing subsystem. The internal tracing API now e.g. `tracepoint_reaction_start`, takes an environment pointer as the first argument. This is to get the current logical tag where the event takes place. The user-facing API, e.g. `tracepoint_user` takes a pointer to a self-struct as an argument, I dont think we should require an environment pointer directly from the user since it is hidden in the base-fields of the self-struct. So there is an inconsistency in the API there, but I think it is OK. 
- An increasing use of dynamic memory allocation

---
TODO:
- [x] Update all schedulers, not just NP
- [x] Extend environments to modes as well
- [x] Consider adding watchdogs to the environments
- [x] Extend environments to unthreaded runtime
- [x] Support federated execution with environments 
The simplest federated program now work.

### Step 1 Hold all global state in an environment struct
- [x] Add environment struct
- [x] Use tag from environment
- [x] Use mutex and cond_var from environment
- [x] Move all priority queues to environment
- [x] Move all token alloc stuff to environment
- [x] Move barriers to environment
- [x] Move all the other relevant global variables to the environment


cmnrd:
Looks like the CI configuration is broken for this PR. Perhaps due to the CI changes in the LF repo @petervdonovan?

erlingrj:
I addressed @edwardalee's comments. Do @petervdonovan or @lhstrh wanna add something, or should we just merge this?

core/threaded/scheduler_NP.c
```diff
@@ -52,63 +52,60 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "trace.h"
 #include "util.h"
 
-/////////////////// External Variables /////////////////////////
-extern lf_mutex_t mutex;
-
 /////////////////// Scheduler Variables and Structs /////////////////////////
-_lf_sched_instance_t* _lf_sched_instance;
+// lf_scheduler_t* scheduler;
```
petervdonovan, line 56: This can be removed

include/core/lf_types.h
```diff
@@ -271,6 +290,75 @@ typedef struct allocation_record_t {
...
+ */
+
+typedef struct mode_environment_t mode_environment_t;
+
+
+typedef struct environment_t {
+    bool initialized;
+    int id;
+    tag_t current_tag;
+    tag_t stop_tag; // Make global protected by global mutex
```
petervdonovan, line 314: Is this an open TODO? I am not sure how to interpret this comment

core/threaded/scheduler_adaptive.c
```diff
@@ -37,79 +37,721 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-    if (init_called) return;
-    worker_states_init(number_of_workers);
-    worker_assignments_init(number_of_workers, params);
-    init_called = true;
+    if(!init_sched_instance(env, &env->scheduler, number_of_workers, params)) {
+        // Already initialized
+        return;
+    }
+    
+    // FIXME: Check for out of mem
```
petervdonovan, line 715: It looks like this has been addressed... Thank you

include/core/threaded/scheduler_instance.h
```diff
@@ -43,18 +43,22 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif  // NUMBER_OF_WORKERS
 
 #include "semaphore.h"
-#include "scheduler.h"
+#include <stdbool.h>
 
-extern lf_mutex_t mutex;
+#define DEFAULT_MAX_REACTION_LEVEL 100
```
petervdonovan, line 48: This is not introduced by this PR, but perhaps we should open an issue about this. If the max reaction level is not known, it is not safe to assume it is 100.

petervdonovan (approved):
Sorry for my tardiness. The current state of this PR looks good to me. I have only trivial comments.

erlingrj:
Thanks @petervdonovan and @edwardalee. I will merge this and the corresponding LF PR, when we reach agreement on the namespace pollution from `platform.h`

PR #213, opened 2023-05-18
Platform abstraction layer for the RTI

siljesu:
In order to port the RTI to different platforms in the future, it should use the platform abstractions already available. This requires changing all pthread functions and types to their corresponding functions and types in platform.h. Additionally, the rti_mutex and two condition variables must be defined outside of the RTI_instance_t, since they no longer can be initialized with PTHREAD_(...)_INITIALIZER macros.

### DIFF
--- a/core/federated/RTI/rti_lib.h
+++ b/core/federated/RTI/rti_lib.h
@@ -23,7 +23,6 @@
 #include <strings.h>    // Defines bzero().
 #include <assert.h>
 #include <sys/wait.h>   // Defines wait() for process to change state.
-#include <pthread.h>
 
 #include "platform.h"   // Platform-specific types and functions
 #include "util.h" // Defines print functions (e.g., lf_print).
--- a/core/federated/RTI/rti_lib.h
+++ b/core/federated/RTI/rti_lib.h
@@ -71,7 +71,7 @@ typedef enum fed_state_t {
  */
 typedef struct federate_t {
     uint16_t id;            // ID of this federate.
-    pthread_t thread_id;    // The ID of the thread handling communication with this federate.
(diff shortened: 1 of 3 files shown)


core/federated/RTI/rti_lib.h
```diff
@@ -209,6 +199,18 @@ typedef struct RTI_instance_t {
     bool tracing_enabled;
 } RTI_instance_t;
 
+/**
+ * RTI synchronization variables
+ */ 
+// The main mutex lock.
```
edwardalee, line 205:
```suggestion
/**
 * The main mutex lock for the RTI.
 */ 
```

core/federated/RTI/rti_lib.h
```diff
@@ -209,6 +199,18 @@ typedef struct RTI_instance_t {
...
+extern lf_mutex_t rti_mutex;
+
+// Condition variable used to signal receipt of all proposed start times.
```
edwardalee, line 208:
```suggestion
/**
 * Condition variable used to signal receipt of all proposed start times.
 */
```

core/federated/RTI/rti_lib.h
```diff
@@ -209,6 +199,18 @@ typedef struct RTI_instance_t {
...
+extern lf_cond_t received_start_times;
+
+// Condition variable used to signal that a start time has been sent to a federate.
```
edwardalee, line 211:
```suggestion
/**
 * Condition variable used to signal that a start time has been sent to a federate.
 */
```

edwardalee (approved):
Great catch!  Let's merge this.  Small suggests for comments...

PR #214, opened 2023-05-18
Suppress "no symbols" warnings emitted by ranlib

petervdonovan:
This addresses warnings reported on Mac such as:
` /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libcore.a(scheduler_adaptive.c.o) has no symbols`

These warnings are expected because we use the preprocessor to exclude the contents of files that relate to features that are not used.

I have checked that this change suppresses the warnings on the 2013 Mac that we have been using.


lhstrh (approved):
:rocket:

PR #219, opened 2023-05-22
Change nanosleep to lf_sleep in federate and RTI code

siljesu:
Platform abstraction for sleep is already available as lf_sleep(), and should be used instead of nanosleep() in the federate and RTI code. Timespec structs can then be removed entirely and just use the available macros or variables in nanoseconds as input to lf_sleep().

CONNECT_RETRY_INTERVAL is changed to be in nanoseconds instead of seconds, as the other macros.

### DIFF
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -1109,20 +1109,14 @@ void* clock_synchronization_thread(void* noargs) {
     interval_t ns_to_wait = start_time - lf_time_physical();
 
     if (ns_to_wait > 0LL) {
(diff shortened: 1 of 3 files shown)


erlingrj:
This looks good to me. For reference, Silje is merging in the changes required to run the RTI as a user thread on the Zephyr platform.

@Benichiwa on the macos target we are getting a strange-looking Watchdog failure. Can you make anything of it? It is surely not caused by this PR.

edwardalee:
This Watchdog failure looks to me like something is trying to happen after the program has shut down.  I will take a look.

edwardalee (approved):
LGTM!

edwardalee (approved):
LGTM!

PR #220, opened 2023-05-22
Refactor of obsolete gethostbyname() in connect_to_rti()

siljesu:
According to the linux man-page: 

> The gethostbyname*(), gethostbyaddr*(), herror(), and hstrerror() functions are obsolete. Applications should use getaddrinfo(3), getnameinfo(3), and gai_strerror(3) instead.

Currently, gethostbyname() is used in connect_to_rti(). This PR modifies the function to use getaddrinfo() instead. The functionality should be unchanged. If no host matches the given hostname, getaddrinfo() will return a non-zero value and exit. If at least one address structure matches the given hostname, it will set the port the be _uport_ and try to connect. Notice that getaddrinfo() returns a linked list of matching address structures, if more than one matches the hostname and _hints_. According to the man page:

>  There are several reasons why the linked list may have more than one addrinfo structure, including: the network host is multihomed, accessible over multiple protocols (e.g., both AF_INET and AF_INET6); or the same service is available from multiple socket types (one SOCK_STREAM address and another SOCK_DGRAM address, for example). Normally, the application should try using the addresses in the order in which they are returned.

So in the current application, I think it should never return more than one address structure (the _hints_ structure specifies AF_INET and SOCK_STREAM specifically). And if it does, we essentially ignore all but the first one, and if the first one fails, try with the next port number instead.

### DIFF
--- a/core/federated/federate.c
+++ b/core/federated/federate.c
@@ -1046,12 +1046,12 @@ void connect_to_rti(const char* hostname, int port) {
         char str[6];
         sprintf(str,"%u",uport);
 
-        // Get address structures matching hostname, port and hints criteria
-        // There should only ever be one matching address structure, and
-        // we connect to that.
+        // Get address structure matching hostname and hints criteria, and
+        // set port to the port number provided in str. There should only 
+        // ever be one matching address structure, and we connect to that.
         int server = getaddrinfo(hostname, &str, &hints, &res);
         if (server != 0) {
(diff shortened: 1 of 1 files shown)


edwardalee (approved):
Looks good to me!

edwardalee:
This should be merged, but I don't understand the test failures. There is no information.

PR #222, opened 2023-05-25
Output error when trying to use --auth (-a) for RTI built without -DAUTH=ON

hokeun:
Add an error message when trying to use the --auth (-a) option for the RTI built without the -DAUTH=ON option. Also fix a minor build warning.

Relevant issue: https://github.com/lf-lang/lingua-franca/issues/1777

### DIFF
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -1586,7 +1586,7 @@ bool authenticate_federate(int socket) {
     write_to_socket(socket, message_length, rti_hello_buffer);
 
     // Check HMAC of received FED_RESPONSE message.
-    size_t hmac_length = SHA256_HMAC_LENGTH;
+    unsigned int hmac_length = SHA256_HMAC_LENGTH;
     size_t federation_id_length = strnlen(_RTI.federation_id, 255);
     size_t fed_id_length = sizeof(uint16_t);
 
@@ -1997,6 +1997,11 @@ int process_args(int argc, const char* argv[]) {
            i++;
            i += process_clock_sync_args((argc-i), &argv[i]);
         } else if (strcmp(argv[i], "-a") == 0 || strcmp(argv[i], "--auth") == 0) {
+            #ifndef __RTI_AUTH__
+            fprintf(stderr, "Error: --auth requires the RTI to be built with the -DAUTH=ON option.\n");
+            usage(argc, argv);
+            return 0;
+            #endif
             _RTI.authentication_enabled = true;
         } else if (strcmp(argv[i], "-t") == 0 || strcmp(argv[i], "--tracing") == 0) {
             _RTI.tracing_enabled = true;


core/federated/RTI/rti_lib.c
```diff
@@ -1586,7 +1586,7 @@ bool authenticate_federate(int socket) {
     write_to_socket(socket, message_length, rti_hello_buffer);
 
     // Check HMAC of received FED_RESPONSE message.
-    size_t hmac_length = SHA256_HMAC_LENGTH;
+    unsigned int hmac_length = SHA256_HMAC_LENGTH;
```
hokeun, line 1589: This is to fix the following build warning:

```
/Users/hokeunkim/Development/lingua-franca/org.lflang/src/lib/c/reactor-c/core/federated/RTI/rti_lib.c:1608:19: warning: incompatible pointer types passing 'size_t *' (aka 'unsigned long *') to parameter of type 'unsigned int *' [-Wincompatible-pointer-types]
         rti_tag, &hmac_length);
                  ^~~~~~~~~~~~
/opt/homebrew/Cellar/openssl@3/3.1.0/include/openssl/hmac.h:56:54: note: passing argument to parameter 'md_len' here
                    unsigned char *md, unsigned int *md_len);
```

core/federated/RTI/rti_lib.c
```diff
@@ -1997,6 +1997,11 @@ int process_args(int argc, const char* argv[]) {
            i++;
            i += process_clock_sync_args((argc-i), &argv[i]);
         } else if (strcmp(argv[i], "-a") == 0 || strcmp(argv[i], "--auth") == 0) {
+            #ifndef __RTI_AUTH__
+            fprintf(stderr, "Error: --auth requires the RTI to be built with the -DAUTH=ON option.\n");
+            usage(argc, argv);
+            return 0;
```
edwardalee, line 2003: Shouldn't this exit with a non-zero value? Otherwise, tests will pass?

edwardalee (changes requested):

core/federated/RTI/rti_lib.c
hokeun: This return value is actually not the exit code.
The return value is checked here: https://github.com/lf-lang/reactor-c/blob/main/core/federated/RTI/rti.c#L65
And return value zero actually leads to a non-zero exit code.

hokeun: As I mentioned above, the return value is not an exit code.

lhstrh (approved):
LGTM!

hokeun:
This change will make sure that the RTI outputs an error and terminates with `return -1` in main (https://github.com/lf-lang/reactor-c/blob/main/core/federated/RTI/rti.c#L67) when it's executed with an authentication option while it's not built with the OpenSSL support, as shown below. I believe this will fail the test when running the SimpleFederatedAuthenticated test with an RTI without OpenSSL (authentication) support.


```
./test/C/bin/SimpleFederatedAuthenticated
Federate SimpleFederatedAuthenticated in Federation ID '3c1c172f45cb7d2dab5c5ad93c47549fae6fbf815964fa09'
RTI: Federation ID: 3c1c172f45cb7d2dab5c5ad93c47549fae6fbf815964fa09
Error: --auth requires the RTI to be built with the -DAUTH=ON option.

Command-line arguments:

  -i, --id <n>
   The ID of the federation that this RTI will control.

  -n, --number_of_federates <n>
   The number of federates in the federation that this RTI will control.

  -p, --port <n>
   The port number to use for the RTI. Must be larger than 0 and smaller than 65535. Default is 15045.

  -c, --clock_sync [off|init|on] [period <n>] [exchanges-per-interval <n>]
   The status of clock synchronization for this federate.
       - off: Clock synchronization is off.
       - init (default): Clock synchronization is done only during startup.
       - on: Clock synchronization is done both at startup and during the execution.
   Relevant parameters that can be set:
       - period <n>(in nanoseconds): Controls how often a clock synchronization attempt is made
          (period in nanoseconds, default is 5 msec). Only applies to 'on'.
       - exchanges-per-interval <n>: Controls the number of messages that are exchanged for each
          clock sync attempt (default is 10). Applies to 'init' and 'on'.

  -a, --auth Turn on HMAC authentication options.

  -t, --tracing Turn on tracing.

Command given:
RTI -i 3c1c172f45cb7d2dab5c5ad93c47549fae6fbf815964fa09 -a -n 2 -c init exchanges-per-interval 10

#### Launching the federate federate__fed1.
#### Launching the federate federate__fed2.
#### Bringing the RTI back to foreground so it can receive Control-C.
./test/C/bin/SimpleFederatedAuthenticated: line 69: fg: job has terminated
#### Received ERR signal on line 69. Invoking cleanup().
Killing federate 63269.
Killing federate 63270.
#### Killing RTI 63267.
./test/C/bin/SimpleFederatedAuthenticated: line 17: kill: (63267) - No such process
```

edwardalee (approved):
LGTM! Thanks for fixing this.

PR #224, opened 2023-05-26
Refactoring of the RTI implementation

ChadliaJerad:
RTI refactoring to separate common coordination functions between federates and enclaves.
There are two types of RTI instances, one for federates, and one for enclaves.

PR #228, opened 2023-05-31
RTI exit while saving the trace file

ChadliaJerad:
This PR enables saving the trace file of the RTI, also after receiving a kill (Ctrl-C) signal.
This will help with debugging failing federated executions.

### DIFF
--- a/core/federated/RTI/rti.c
+++ b/core/federated/RTI/rti.c
@@ -80,7 +80,7 @@ int main(int argc, const char* argv[]) {
     lf_cond_init(&received_start_times, &rti_mutex);
     lf_cond_init(&sent_start_time, &rti_mutex);
 
-    // Catch the Ctrl-C signal, for a clean exit that does not loose the trace information
+    // Catch the Ctrl-C signal, for a clean exit that does not lose the trace information
     signal(SIGINT, exit);
     if (atexit(termination) != 0) {
         lf_print_warning("Failed to register termination function!");
--- a/core/federated/RTI/rti.c
+++ b/core/federated/RTI/rti.c
@@ -47,6 +47,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 #include "rti_lib.h"
+
+/** Termination function */
+void termination();
+
 unsigned int _lf_number_of_workers = 0u;
 
 extern RTI_instance_t _RTI;
@@ -61,6 +65,12 @@ int main(int argc, const char* argv[]) {
     lf_mutex_init(&rti_mutex);
     lf_cond_init(&received_start_times, &rti_mutex);
     lf_cond_init(&sent_start_time, &rti_mutex);
+
+    // Catch the Ctrl-C signal, for a clean exit that does not loose the trace information
+    signal(SIGINT, exit);
+    if (atexit(termination) != 0) {
+        lf_print_warning("Failed to register termination function!");
+    }
 
     if (!process_args(argc, argv)) {
         // Processing command-line arguments failed.
@@ -79,9 +89,20 @@ int main(int argc, const char* argv[]) {
     }
     int socket_descriptor = start_rti_server(_RTI.user_specified_port);
     wait_for_federates(socket_descriptor);
+    
+    termination();
+
+    return 0;
+}
+
+/**
+ * @brief A clean termination of the RTI will write the trace file, if tracing is
+ * enabled, before exiting.
+ */
+void termination() {
     if (_RTI.tracing_enabled) {
         stop_trace();
-    }
+        printf("RTI trace file saved.\n");
+    }   
     printf("RTI is exiting.\n");
-    return 0;
 }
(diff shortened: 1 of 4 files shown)


core/federated/RTI/rti.c
```diff
@@ -47,6 +47,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 #include "rti_lib.h"
+
+/** Termination function */
+void termination();
```
edwardalee, line 52: Instead of a forward declaration here, why not just define the function here?

core/federated/RTI/rti.c
```diff
@@ -79,9 +89,20 @@ int main(int argc, const char* argv[]) {
     }
     int socket_descriptor = start_rti_server(_RTI.user_specified_port);
     wait_for_federates(socket_descriptor);
+    
+    termination();
```
edwardalee, line 93: Won't this result in the termination() function being called twice?

core/federated/RTI/rti_lib.h
```diff
@@ -23,6 +23,7 @@
 #include <strings.h>    // Defines bzero().
 #include <assert.h>
 #include <sys/wait.h>   // Defines wait() for process to change state.
+#include <signal.h>     // To trap ctrl-c and invoke a clean stop to save the trace file, if needed.
```
edwardalee, line 26: Isn't this #include only needed in rti.c? If so, it should be put there.

edwardalee (changes requested):
Some questions and suggestions. Thanks for fixing the printf format warnings too!

core/federated/RTI/rti.c
erlingrj: I sometimes do this when I want to avoid cluttering the beginning of the source file with "unimportant" function definitions.

core/federated/RTI/rti.c
```diff
@@ -79,9 +89,20 @@ int main(int argc, const char* argv[]) {
...
+
+/**
+ * @brief A clean termination of the RTI will write the trace file, if tracing is
+ * enabled, before exiting.
+ */
+void termination() {
     if (_RTI.tracing_enabled) {
         stop_trace();
-    }
+        printf("RTI trace file saved.\n");
```
erlingrj, line 105: I think we  should use the `lf_print` functions rather than `printf`, because there are platforms which doesn't support `printf` but still can provide some implementation of `lf_print`.

core/federated/RTI/rti_lib.h
erlingrj: I guess some of the include statements above also don't belong in the header file

erlingrj (commented):
Nice addition to the RTI :)

core/federated/RTI/rti.c
ChadliaJerad: I generally want to have `main` as the first function in a file...
But I will update it so that the coding style will be the same.

core/federated/RTI/rti_lib.h
ChadliaJerad: > Isn't this #include only needed in rti.c? If so, it should be put there.
Moved to `rti.c`.

> I guess some of the include statements above also don't belong in the header file
Will check them.

core/federated/RTI/rti.c
ChadliaJerad: Indeed!
The RTI was not using the abstraction provided by LF, as pointed out by @siljesu.
Will update them.

edwardalee:
I have no idea why these tests are not passing. Looks like a socket connection was broken while downloading gradle? @petervdonovan : any ideas?

ChadliaJerad:
BTW, `lf-gedf-np` and `lf-gedf-np-ci` tests succeed in #224, which already has updates merged from `rti-exit-save-trace` (into `rti-refactoring` branch).

petervdonovan:
> I have no idea why these tests are not passing. Looks like a socket connection was broken while downloading gradle? @petervdonovan : any ideas?

Three tests failed, all for different reasons:
- I agree that there was a network issue while downloading Gradle. This isn't really our fault, although I suppose we could look into caching like what Marten implemented for Windows in `lingua-franca`.
- `lf-gedf-np-ci` tests are failing due to the bug that I pushed to master a few days ago. I expect the fix to be merged in the next couple of hours.
- I don't know why the `lf-adaptive` tests failed, but it looks like some kind of flaky test failure related to federated communication. There isn't a lot of information to go on:
```
Federate DistributedCMakeIncludeSeparateCompile in Federation ID 'f6dd8f3f58e226751a9650be2bdc2372fd558d9f8a5271d5'
    #### Launching the federate federate__f1.
    #### Launching the federate federate__b1.
    #### Bringing the RTI back to foreground so it can receive Control-C.
    RTI -i ${FEDERATION_ID} -n 2 -c init exchanges-per-interval 10
    Federate 0: ERROR: Read 0 bytes, but expected 1. errno=57
    Federate 0: FATAL ERROR: Failed to read response from RTI.
```

PR #231, opened 2023-06-02
Fix protocol of HMAC authentication to start from federate.

Jakio815:
This PR fixes lf-lang/lingua-franca#1777.

When the RTI `-a` option is off, and the `.lf` file's `auth=true` option is on, the RTI and federate both waits for each other's message. The previous HMAC authentication protocol started from the RTI sending a message, but I changed it to the federate sending the message to fix the problem above.

The main protocol is like the following with a 3 way handshake.
**1. FED_NONCE (Federate -> RTI)
2. RTI_RESPONSE  (RTI -> Federate) 
3. FED_RESPONSE  (Federate -> RTI)**
Details are below.

### 1. FED_NONCE (Federate -> RTI) ###
The federate sends a FED_NONCE message to the RTI when TCP connection is established. The message is in this order:
- One byte equal to MSG_TYPE_FED_NONCE.
- Two bytes for the federate ID.
- Eight bytes for randomly created federate 's nonce.
### 2. RTI_RESPONSE  (RTI -> Federate)  ###
The RTI sends a response message to the federate. It creates its own nonce, and makes an HMAC tag using the federation ID as the key. The message is in the following order:
- One byte equal to MSG_TYPE_RTI_RESPONSE.
- Eight bytes for RTI's nonce.
- 32 bytes for HMAC tag based on SHA256. 

 The HMAC tag is composed of the following order:
- One byte equal to MSG_TYPE_RTI_RESPONSE.
- Two bytes for the federate ID.
- Eight bytes for received federate's nonce.

### 3. FED_RESPONSE  (Federate -> RTI) ###
The federate sends a response message to the RTI. It checks the received HMAC created by the RTI, and also makes a HMAC tag. The message is in the following order:
- One byte equal to MSG_TYPE_FED_RESPONSE.
- 32 bytes for HMAC tag based on SHA256.

The HMAC tag is composed of the following order:
- One byte equal to MSG_TYPE_FED_RESPONSE.
- Eight bytes for received RTI's nonce.

### RTI Check ###
The RTI finally checks the received HMAC created by the federate.

## Major changes ##
- The federate now starts the authentication.
- The first message from the federate now contains the federate's ID. Previously it happened on the second message.
- The second message from the RTI now does not contain the federate's ID.


include/core/federated/net_common.h
```diff
@@ -336,7 +336,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * between the RTI and the federate.
  * The next eight bytes are RTI's 8-byte nonce (RTI nonce).
  */
```
hokeun, line 338: Can you update the comments for messages?

hokeun (approved):
Looks good to me!

core/federated/RTI/rti_lib.c
lhstrh, line 1: The `authenticate_federate` function is rather long, and it seems easy to factor it out into smaller functions.

lhstrh (changes requested):
This looks OK to me, although I only did a spot check. Please resolve the conflict with `main` and it would be great if you could break down the super long functions into smaller ones.

Jakio815:
Sure. I'll take a look at it as soon as possible. Thanks!

Jakio815:
@lhstrh The function is about 60 lines, and I see that many other functions are even longer. Also, it's kind of hard to split it up, because it is processing the handshake, and I won't be able to split it by similar functions.

I think it would just make the code more dirty, adding more #ifdef #elifs. What do you think?

lhstrh:
> @lhstrh The function is about 60 lines, and I see that many other functions are even longer. Also, it's kind of hard to split it up, because it is processing the handshake, and I won't be able to split it by similar functions.
> 
> I think it would just make the code more dirty, adding more #ifdef #elifs. What do you think?

OK, fair enough!

lhstrh (approved):

PR #232, opened 2023-06-02
Zephyr federated support

siljesu:
To be able to run federates on Zephyr, the following changes must be made to reactor-c. Notice that https://github.com/lf-lang/reactor-c/pull/220 has been merged into this PR as well. This is because Zephyr does not offer support for the obsolete _gethostbyname()_. So, if https://github.com/lf-lang/reactor-c/pull/220 is not approved, then this PR must add a macro separating the two implementations. 

In the Zephyr platform-specific functions, macros are added to allow for static thread stack definition for federated execution.

Additionally, the regex library is not supported by Zephyr.

### DIFF
--- a/core/federated/clock-sync.c
+++ b/core/federated/clock-sync.c
@@ -31,12 +31,15 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 #ifdef FEDERATED
+#ifdef PLATFORM_ZEPHYR
+#else
+#include <sys/socket.h>
+#include <netinet/in.h>
+#endif
 #include <errno.h>
 #include <math.h>
 #include <string.h>
 #include <stdlib.h>
-#include <sys/socket.h>
-#include <netinet/in.h>
 
 #include "platform.h"
 #include "clock-sync.h"
--- a/core/federated/federate.c
+++ b/core/federated/federate.c
@@ -33,6 +33,8 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifdef FEDERATED
 #ifdef PLATFORM_ARDUINO
 #error To be implemented. No support for federation on Arduino yet.
+#elif PLATFORM_ZEPHYR
+#warning Federated support on Zephyr is still experimental.
 #else
 #include <arpa/inet.h>  // inet_ntop & inet_pton
 #include <netdb.h>      // Defines getaddrinfo(), freeaddrinfo() and struct addrinfo.
@@ -2414,6 +2416,7 @@ void terminate_execution() {
     // possibility of deadlock. To ensure this, this
     // function should NEVER be called while holding any mutex lock.
     lf_mutex_lock(&outbound_socket_mutex);
+    // FIXME: Should this be _fed.number_of_outbound_p2p_connections instead?
     for (int i=0; i < NUMBER_OF_FEDERATES; i++) {
         // Close outbound connections, in case they have not closed themselves.
         // This will result in EOF being sent to the remote federate, I think.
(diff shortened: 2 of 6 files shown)


siljesu:
Good points. I think you are right; the regex library should be included in Newlib. 

However, I am getting linker errors when using the regex functions:

```
.../zephyr-sdk-0.15.2/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.1.0/../../../../arm-zephyr-eabi/bin/ld.bfd: core/libcore.a(net_util.c.obj): in function `extract_rti_addr_info':
.../src-gen/federate__unitB/core/federated/net_util.c:698: undefined reference to `regcomp'
.../zephyr-sdk-0.15.2/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.1.0/../../../../arm-zephyr-eabi/bin/ld.bfd: 
.../src-gen/federate__unitB/core/federated/net_util.c:703: undefined reference to `regexec'
.../zephyr-sdk-0.15.2/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.1.0/../../../../arm-zephyr-eabi/bin/ld.bfd: 
.../src-gen/federate__unitB/core/federated/net_util.c:716: undefined reference to `regfree'
collect2: error: ld returned 1 exit status
```
Any ideas? It is similar to https://github.com/zephyrproject-rtos/zephyr/issues/8502 - could it be a bug in Zephyr? Or maybe some configuration option I have missed?

erlingrj:
Could be a configuration issue. How can I reproduce it here?

siljesu:
Try to build any zephyr program with -b mimxrt1170_evk_cm7 and the option CONFIG_NEWLIB_LIBC=y. This should not be able to compile, since regcomp takes parameters. With this program I am not getting a linker error, but another error in regex.h. So it seems the linker error was my fault. But still I am getting the error shown below.

```
#include <zephyr/kernel.h>
#include <zephyr/drivers/gpio.h>
#include <regex.h>

void main(void)
{
	regcomp();
}
```

```
zephyr-sdk-0.15.2/arm-zephyr-eabi/arm-zephyr-eabi/sys-include/regex.h:43:9: error: unknown type name 'off_t'
   43 | typedef off_t regoff_t;

```

erlingrj:
DId you resolve it?

siljesu:
No, I am still getting a linker issue, even if the correct library files are included. It could be related to linker flags, maybe. I guess you should be able to reproduce it properly now, with the new pull request in lf-west-template.

erlingrj:
The problem is resolved by using picolibc rather than newlibc. But I remember picking newlibc because I had problems with the other C library options. But you could try changing form newlibc to picolibc in the prj.conf file that is in the lingua-franca repository. It is somewhere under resources/platform/zephyr I think

erlingrj:
@siljesu What is the status on this? Will you have time to finalize it?

erlingrj:
To get this merged we need to do the following:

- [ ] Switch from newlib to picolibc and verify that all tests still pass (https://github.com/lf-lang/lingua-franca/pull/1918)
- [ ] Add *building* of federated tests to the CZephyr integration test suite. Ideally the whole federated test suite.
- [ ] Get the lf-west-template PR ready  (https://github.com/lf-lang/lf-west-template/pull/3)
- [ ] Test on actual boards

lhstrh:
@erlingrj would you be able to turn this into a brief project description for the embedded systems class? We could add it [here](https://www.lf-lang.org/embedded-lab/Project.html).

erlingrj:
Could be a nice project. I have seen that Zephyr supports Sockets over USB, so it should, in theory, be possible to run federated LF programs spanning a Pico and a Linux PC connected by USB. The Pico would be a better example platform for Federated Embedded LF as it is more common. The boards Silje were using are more rare, and in general, ethernet on microcontroller boards is not that common. By what time should I propose a project?

lhstrh:
No hurry with this, but what you describe sounds great. I think students will figure out their project sometime next month...

PR #233, opened 2023-06-08
pico lf

gundralaa:


core/platform/lf_pico_support.c
```diff
@@ -0,0 +1,329 @@
...
+    multicore_reset_core1();
+}
+
+/**
+ * Fetch the value of an internal (and platform-specific) physical clock and
+ * store it in `t`. in nanoseconds
+ *
+ * Ideally, the underlying platform clock should be monotonic. However, the
+ * core lib tries to enforce monotonicity at higher level APIs (see tag.h).
+ * TODO: might want to use the RTC
```
edwardalee, line 69: Maybe this should be an ordinary comment in the function body? It refers to the implementation, not the interface.

core/platform/lf_pico_support.c
```diff
@@ -0,0 +1,329 @@
...
+int lf_mutex_init(lf_mutex_t* mutex) {
+    recursive_mutex_init(mutex);
+    return 0;
+}
+
+/**
+ * Lock a mutex.
+ *
+ * @return 0 on success, platform-specific error number otherwise.
+ * TODO: should this block?
```
edwardalee, line 238: Acquiring a mutex always needs to block.

PR #236, opened 2023-06-09
Make _lf_count_token_allocations extern instead of static

erlingrj:
I kept getting warnings with the Zephyr toolchain about `_lf_count_token_allocations` being unused. It is very uncommon to declare static variables in header files. It means that each file that includes this header will get its own copy of the variable. Is that intended? My guess is that its a typo and should be `extern` creating a single global variable shared across compilation units.


edwardalee (approved):
Yes, this looks right. Good catch!

PR #238, opened 2023-06-10
Further RTI refactoring for enclaved scheduling

erlingrj:
I am sorry to propose these additional RTI changes, after we have merged in the other RTI refactorings. But how I understand the function of the local and remote RTI, I felt that a substantial refactoring was necessary to make everything more clear. This is still a draft and I would like your feedback. I will try to summarize the changes and my reason for them below:

### New data structures
- `rti_local_t` struct for storing info specific for enclaved coordination. Currently no additions beyond the `rti_common_t` but I anticipate that it will be necessary
- `federation_rti_t` -> `rti_remote_t` struct with info specific to federate coordination 
- `enclave_rti_t` -> `rti_common_t` Struct with info common to enclaved and federated coordination
- `enclave_t` -> `reactor_node_info_t` Generic information about an enclave/federation, e.g. num upstream etc
- `federate_t` -> `federate_info_t` federate-specific information. Added the `info` part because it isn't "the" federation.
- `enclave_info_t` enclave-specific information.

As you see, I have created new enclave specific structures and created generic structures which both enclave-specific and federate-specific structures inherit from as follows:

```mermaid
classDiagram
class rti_common_t {
reactor_node_info_t **reactor_nodes;
int number_of_reactor_nodes
...
}
class rti_remote_t {
rti_common_t base
...
}

class rti_local_t {
rti_common_t base
}

class enclave_info_t {
reactor_node_info_t base
}

class federate_info_t {
reactor_node_info_t base
}


    rti_common_t <|-- rti_remote_t
    rti_common_t <|-- rti_local_t

    reactor_node_info_t <|-- federate_info_t
    reactor_node_info_t <|-- enclave_info_t
    rti_common_t *-- reactor_node_info_t
```

### File renaming
- `rti.c` -> `main.c`. I think this is clearer. `main.c` is only used for building the remote RTI program
- `rti_lib.*`  -> `rti_remote.*`. I don't think rti_lib was a good name. This compilation unit contains functions specific to the remote RTI. Thus renamed as such
- `enclave.*` -> `rti_common.*`. I think this was  confusing. This compilation unit contains the functionality common to both the remote and local RTI. Its header file should not be included by anyone except rti_remote and rti_local. 
- `enclave_impl.c` -> `rti_local.c`. This file will contain functionality specific to the local RTI which deals with enclaves. It is paired by a header file which provides the only API to the threaded reactor-c runtime. To support federated enslaved scheduling, also the federate.h API must be "behind" this local_rti. 

### Other
- I decided to avoid "matching" struct definitions to enable recasting and instead put the `rti_common_t` as the first field in `rti_remote_t`. There is some casting needed because `rti_common_t` has a pointer array of `reactor_node_info_t` which must be cast back and forward between `federate_info_t`. But it is not too bad. I made a simple macro for it.
- I didn't like these global variables defined with extern. Now, each source file contains a static global pointer. It is updated by the `initialize` function they all expose. I thought this was cleaner


ChadliaJerad:
> I am sorry to propose these additional RTI changes, after we have merged in the other RTI refactorings. But how I understand the function of the local and remote RTI, I felt that a substantial refactoring was necessary to make everything more clear.

The previous refactoring was just an increment, very far away from being final :-)

> ### New data structures
> * `rti_local_t` struct for storing info specific for enclaved coordination. Currently no additions beyond the `rti_common_t` but I anticipate that it will be necessary
> * `federation_rti_t` -> `rti_remote_t` struct with info specific to federate coordination
> * `enclave_rti_t` -> `rti_common_t` Struct with info common to enclaved and federated coordination
> * `enclave_t` -> `reactor_node_info_t` Generic information about an enclave/federation, e.g. num upstream etc
> * `federate_t` -> `federate_info_t` federate-specific information. Added the `info` part because it isn't "the" federation.
> * `enclave_info_t` enclave-specific information.

The new structure is quite straightforward, indeed.
What I find confusing, however, is the term `recator_node_info_t`. I see it more as the `environment_t` you created...

Structure-wise, this is clear.
Behavior-wise (which boils up to the structure), `federates` and `enclaves` need to have instances of `rti_local_t`, so that one federate can have one or more enclaves. 
Furthermore, the challenge will be in distinguishing between local and remote coordination decisions... Can this be reflected in the structure? That will be ideal.


> ### File renaming
> * `rti.c` -> `main.c`. I think this is clearer. `main.c` is only used for building the remote RTI program
> * `rti_lib.*`  -> `rti_remote.*`. I don't think rti_lib was a good name. This compilation unit contains functions specific to the remote RTI. Thus renamed as such
> * `enclave.*` -> `rti_common.*`. I think this was  confusing. This compilation unit contains the functionality common to both the remote and local RTI. Its header file should not be included by anyone except rti_remote and rti_local.
> * `enclave_impl.c` -> `rti_local.c`. This file will contain functionality specific to the local RTI which deals with enclaves. It is paired by a header file which provides the only API to the threaded reactor-c runtime. To support federated enslaved scheduling, also the federate.h API must be "behind" this local_rti.

I agree that renaming `enclave.*` into `rti_common.*` is more straightforward.  

> ### Other
> * I decided to avoid "matching" struct definitions to enable recasting and instead put the `rti_common_t` as the first field in `rti_remote_t`. There is some casting needed because `rti_common_t` has a pointer array of `reactor_node_info_t` which must be cast back and forward between `federate_info_t`. But it is not too bad. I made a simple macro for it.

This is a simple design decision :-) Each approach has its own pros and cons.
I am personally flexible and fine with following your programming style ;-) 

> * I didn't like these global variables defined with extern. Now, each source file contains a static global pointer. It is updated by the `initialize` function they all expose. I thought this was cleaner

They couldn't and shouldn't remain as global variables anyway. It was just an increment.

Thanks for the diagrams! They are important.
 
I will proceed with the code review tomorrow.

erlingrj:
Thanks for the feedback. You dont need to give a detailed review of the code, I think I will rather merge it into my new enclaves branch and then eventually into main. I also don't like `reactor_node_info_t` but I couldn't come up with a better name. It should represent what `enclave_info_t` and `federate_info_t` have in common. It could also be something like `coordination_node_t` or something

erlingrj:
Regarding the `rti_local_t`. We dont need one per enclave/federate. Only a single `rti_local_t` per program. In the future, it will handle both local coordination of enclaves as well as "escalating" to the `remote_rti_t`

edwardalee (approved):
I like this refactoring a lot. This continues to improve!

I think the term `reactor_node`, however doesn't really work for me. `coordination_node` would be better. Or `scheduling_node`? Why was `enclave` not the right term for this?

core/federated/RTI/rti_common.h
```diff
@@ -49,63 +50,67 @@ typedef struct enclave_t {
...
-
-typedef struct {
-    tag_t tag;           // NEVER if there is no tag advance grant.
-    bool is_provisional; // True for PTAG, false for TAG.
-} tag_advance_grant_t;
+void logical_tag_complete(reactor_node_info_t* enclave, tag_t completed);
 
 /** 
- * Initialize the enclave with the specified ID.
+ * Initialize the reactor- with the specified ID.
```
edwardalee, line 105: Here, `enclave` seems like the right term. The `@param` entries are now inconsistent with the doc.

PR #240, opened 2023-06-15
Less namespace pollution

erlingrj:
This is still a draft, but wanted some feedback. Essentially, I didn't find a way to remove `windows.h` and friends from `platform.h` since we need it to typdef `lf_mutex_t` etc. I think the only real possibility is to only allow using pointers to lf_thread_t, lf_mutex_t and lf_cond_t in the runtime. And those are just void pointers. They can then be passed into platform files and re-cast. 

I didn't particularly like that idea, so I made a change to avoid including `platform.h` into `lf_types.h` this way, at least the user wont get their namespace polluted unless they manually include `platform.h`

### DIFF
--- a/include/core/federated/federate.h
+++ b/include/core/federated/federate.h
@@ -36,6 +36,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "tag.h"
 #include "lf_types.h"
+#include "environment.h"
 #include "platform.h"
 
 #ifndef ADVANCE_MESSAGE_INTERVAL


core/platform/lf_C11_threads_support.c
edwardalee, line 1: Should this file have function comments? Shouldn't they just be in  `platform.h`? Comments here probably should only be about the particular customization to C11.  Same question applies to all the .c files.

core/platform/lf_C11_threads_support.c
```diff
@@ -0,0 +1,123 @@
+#include "platform.h"
+#include <threads.h>
+#include <stdlib.h>
+#include <stdint.h> // For fixed-width integral types
+
+
+/**
+ * Create a new thread, starting with execution of lf_thread
+ * getting passed arguments. The new handle is tored in thread.
```
edwardalee, line 9:
```suggestion
 * getting passed arguments. The new handle is stored in the thread argument.
```

core/platform/lf_C11_threads_support.c
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @return 0 on success, error number otherwise (see thrd_create()).
+ */
+int lf_thread_create(lf_thread_t* thread, void *(*lf_thread) (void *), void* arguments) {
+    return thrd_create((thrd_t*)thread, (thrd_start_t)lf_thread, arguments);
+}
+
+/**
+ * Make calling thread wait for termination of the thread.  The
+ * exit status of the thread is stored in thread_return, if thread_return
```
edwardalee, line 19:
```suggestion
 * exit status of the thread is stored in thread_return if thread_return
```

core/platform/lf_C11_threads_support.c
```diff
@@ -0,0 +1,123 @@
...
+    switch (return_value) {
+        case thrd_timedout:
+            return_value = LF_TIMEOUT;
+            break;
+
+        default:
+            break;
+    }
+    return return_value;
+}
```
edwardalee, line 123:
```suggestion
}

```

core/platform/lf_POSIX_threads_support.c
```diff
@@ -0,0 +1,146 @@
...
+    switch (return_value) {
+        case ETIMEDOUT:
+            return_value = LF_TIMEOUT;
+            break;
+
+        default:
+            break;
+    }
+    return return_value;
+}
```
edwardalee, line 146:
```suggestion
}

```

edwardalee (approved):
This looks like a pretty good solution to me! Minor suggestions included.

petervdonovan:
Overall this approach is the one that I would most prefer and I think we should merge it when the tests pass.

I agree with Edward's observation about duplicated comments.

core/platform/lf_C11_threads_support.c
erlingrj: Thanks, have addressed this now

petervdonovan (approved):
Thanks Erling!

edwardalee (approved):
LGTM!

PR #242, opened 2023-06-26
Scheduling enclaves

erlingrj:
This PR implements the first version of scheduling enclaves in the C target. This PR is paired with https://github.com/lf-lang/lingua-franca/pull/1872.

----
Things not supported yet:
1. Enclaved federations. 
3. Physical connections between enclaves
4. Ability to specify a maximum lead between enclaves
5. Smart mutexes wich can catch violations of specified order of acquisition


erlingrj:
@edwardalee  and @lhstrh. I am tagging you for a review now. There is still a test failure on macOS on a test with coordinated shutdown of enclaves. I will look at this when I get my hands on a MacBook next week

erlingrj:
The compiler errors now are due the fact that in reactor-c the preprocessor flags for picking the scheduler has changed, from `NP` -> `SCHED_NP` and so on. I think that my first step in November should be to craft a minimal `lingua-franca` PR which can me merged together with this PR. It should *not* introduce the AST transformation and enable enclaves. But just let us merge these refactorings of the RTI, which I think are good, into master.

core/federated/RTI/rti_local.c
```diff
@@ -0,0 +1,206 @@
...
+#include "platform.h"
+#include "environment.h"
+#include "trace.h"
+#include "reactor.h"
+
+
+// The following are macros for to be used in the enter and exit
+// of the local RTI API. We extract this to macros to make sure that 
+// mutexes are acquired and released in the correct order.
+#define LTC_LOCKED_PROLOGUE(enclave) \
```
edwardalee, line 24: It's not clear how this macro helps at all with making sure the order of mutex locks is correct.  This just unlocks the enclave mutex, presumably setting the stage for acquiring the global mutex?  If anything, this seems to obscure the order.  Perhaps better would be to document what the order should be.  Is the rule that you should never acquire the global mutex while holding an enclave mutex?  What about acquiring the mutex of another enclave?

core/federated/RTI/rti_local.c
```diff
@@ -0,0 +1,206 @@
...
+    do { \
+        lf_mutex_unlock(&enclave->env->mutex); \
+    } while(0)
+
+#define LTC_LOCKED_EPILOGUE(enclave) \
+    do { \
+        lf_mutex_lock(&enclave->env->mutex); \
+    } while(0)
+
+#define NET_LOCKED_PROLOGUE(enclave) \
```
edwardalee, line 34: What is the value of `rti_next_event_tag_locked` requiring that the enclave mutex be held when it is called if it is going to immediately release that lock and acquire the global mutex?  Wouldn't it be better to just require that the global mutex be held?

core/federated/RTI/rti_local.c
```diff
@@ -0,0 +1,206 @@
...
+        lf_mutex_unlock(&enclave->env->mutex); \
+        lf_mutex_lock(rti_local->base.mutex); \
+    } while(0)
+
+#define NET_LOCKED_EPILOGUE(enclave) \
+    do { \
+        lf_mutex_unlock(rti_local->base.mutex); \
+        lf_mutex_lock(&enclave->env->mutex); \
+    } while(0)
+#define UPDATE_OTHER_PROLOGUE() \
```
edwardalee, line 45: Same question regarding `rti_update_other_net_locked`.
These macros don't seem to me to ensure anything about order and just obscure what is going on.

edwardalee (changes requested):
This is just a review of rti_local.h/c.  I don't understand the locking strategy here.

edwardalee:
> The compiler errors now are due the fact that in reactor-c the preprocessor flags for picking the scheduler has changed, from `NP` -> `SCHED_NP` and so on. I think that my first step in November should be to craft a minimal `lingua-franca` PR which can me merged together with this PR. It should _not_ introduce the AST transformation and enable enclaves. But just let us merge these refactorings of the RTI, which I think are good, into master.

I have started doing this in `lingua-franca/enclaves3`, which is pointed to by `reactor-c/enclaves3`, which includes some small extensions of your `reactor-c/enclaves2` branch.  Feel free to push directly to either `enclave3` branch or to your `reactor-c/enclaves2` branch, which we can then merge into `reactor-c/enclaves3`.

core/federated/RTI/rti_local.c
erlingrj: You might be right that I could also just do the locking and unlocking inside the functions and document it. For the LTC function, we only release the local enclave mutex becasue the locking of the RTI mutex is done inside the function we call next, in `rti_common.c`.

core/federated/RTI/rti_local.c
erlingrj: It is less about requiring it to hold the mutex and more about that calls to this function only appears within critical sections of the enclave. So thats why I moved the unlocking and locking of the enclave mutex inside the function. 

If I remember correctly, there isnt a way to grab the "global" mutex before calling the function. The global mutex is the RTI mutex and we only have it in scope inside the RTI files

core/federated/RTI/rti_local.c
erlingrj: I get your point. It is not about ensuring things, but refactoring out these few critical lines so that there is zero chance of someone else, by mistake reordering them while doing some changes to the function. If you find it confusing we can remove the macros and do the locking and unlocking in "plaintext" in the function

erlingrj (commented):
Answering questions

core/tag.c
```diff
@@ -134,6 +134,7 @@ int lf_tag_compare(tag_t tag1, tag_t tag2) {
 
 tag_t lf_delay_tag(tag_t tag, interval_t interval) {
     if (tag.time == NEVER || interval < 0LL) return tag;
+    if (tag.time == FOREVER && tag.microstep == FOREVER_MICROSTEP) return tag;
```
edwardalee, line 137:
```suggestion
    if (tag.time >= FOREVER - interval) return tag;
```

include/core/reactor.h
```diff
@@ -418,6 +418,36 @@ trigger_handle_t _lf_schedule(environment_t* env, trigger_t* trigger, interval_t
...
 
 
+/**
+ * @brief Retrives the enclave ids of the direct upstream enclaves.
+ * The implementation is code-generated.
+ * 
+ * @param enclave_id The enclave for which to search for direct upstream
+ * @param result This pointer is dereferenced and pointed to the resulting array.
+ * @return int The number of direct upstreams.
+ */
```
edwardalee, line 428:
```suggestion
/**
 * @brief Get the array of ids of enclaves directly upstream of the specified enclave.
 * This updates the specified result pointer to point to a statically allocated array of IDs
 * and returns the length of the array. The implementation is code-generated.
 * 
 * @param enclave_id The enclave for which to report upstream IDs.
 * @param result The pointer to dereference and update to point to the resulting array.
 * @return The number of direct upstream enclaves.
 */
```

include/core/reactor.h
```diff
@@ -418,6 +418,36 @@ trigger_handle_t _lf_schedule(environment_t* env, trigger_t* trigger, interval_t
...
+int _lf_get_upstream_of(int enclave_id, int** result);
+
+/**
+ * @brief Retrives the enclave ids of the direct downstream enclaves.
+ * The implementation is code-generated.
+ * 
+ * @param enclave_id The enclave for which to search for direct downstream.
+ * @param result This pointer is dereferenced and pointed to the resulting array.
+ * @return int The number of direct downstream.
+ */
```
edwardalee, line 438:
```suggestion
/**
 * @brief Get the array of ids of enclaves directly downstream of the specified enclave.
 * This updates the specified result pointer to point to a statically allocated array of IDs
 * and returns the length of the array. The implementation is code-generated.
 * 
 * @param enclave_id The enclave for which to report downstream IDs.
 * @param result The pointer to dereference and update to point to the resulting array.
 * @return The number of direct downstream enclaves.
 */
```

include/core/reactor.h
```diff
@@ -418,6 +418,36 @@ trigger_handle_t _lf_schedule(environment_t* env, trigger_t* trigger, interval_t
...
+int _lf_get_downstream_of(int enclave_id, int** result);
+
+/**
+ * @brief Retrives the delays on the connections to the direct upstream enclaves.
+ * The implementation is code-generated.
+ * 
+ * @param enclave_id The enclave for which to search for upstream delays.
+ * @param result This pointer is dereferenced and pointed to the resulting array.
+ * @return int The number of direct upstreams.
+ */
```
edwardalee, line 448:
```suggestion
/**
 * @brief Retrive the delays on the connections to direct upstream enclaves.
 * This updates the result pointer to point to a statically allocated array of delays.
 * The implementation is code-generated.
 * 
 * @param enclave_id The enclave for which to search for upstream delays.
 * @param result The pointer to dereference and update to point to the resulting array.
 * @return int The number of direct upstream enclaves.
 */
```

edwardalee (approved):
I've finally gone through this whole PR and it looks good to me! Modulo small comments, I think we can merge this as soon as we can get tests to pass in lingua-franca/master with minimal changes to lingua-franca.  This will not yet provide support for enclaves, but will get us much closer.

I'm proposing further changes in the enclaves3 branch, but we can review and merge those after merging this into main.

core/federated/RTI/rti_common.h
```diff
@@ -203,37 +208,20 @@ void notify_provisional_tag_advance_grant(enclave_t* e, tag_t tag);
...
-tag_advance_grant_t next_event_tag(enclave_t* e, tag_t next_event_tag);
 
 /**
  * @brief Update the next event tag of an enclave.
  *
- * This will notify downstream enclaves with a TAG or PTAG if appropriate.
+ * This will notify downstream scheduling_nodes with a TAG or PTAG if appropriate.
  *
- * This function assumes that the caller is holding the rti_mutex.
+ * This function assumes that the caller is holding the mutex.
```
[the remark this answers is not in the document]
erlingrj: It is the RTI mutex. This diff is out-of-date for some reason

erlingrj:
@edwardalee. I have addressed all comments. I removed the PROLOGUE/EPILOGUE macros which locked/unlocked mutex. I have made the documentation clearer with respect to which mutex are assumed held etc. Tests are passing here and in LF? Shall we merge?

PR #245, opened 2023-06-27
Minor fix of the federate id in the tracepoint

ChadliaJerad:
This PR fixes the federate identifier when calling the tracepoint function.


petervdonovan (approved):
LGTM

PR #246, opened 2023-06-28
Memory leak in Python target fixed

jackykwok2024:
This pull request addresses a memory leak issue observed in the current implementation. As shown in the following screenshot, it is noticeable that with each iteration, the size of Python objects constructed using C extensions continues to grow.

Untitled picture

This problem could be attributed to several key issues, which are as follows:

1.	PyObject Reference Count: The reference counts for the majority of `PyObjects` are not being appropriately managed. 
2.	Port Assignment: Rather than using the `_LF_SET` macro to assign values to the` generic_port_instance_struct` in the current implementation, the more suitable method would be to use `lf_set_token` and define a destructor for the token to ensure proper garbage collection.

3. Function Definition: The existing` _lf_free_token_value` function is incorrect and fails to invoke the destructor.

Comments have been added to the code within the commits to provide additional context and insight into the changes made. Other related changes have also been made in the code generator repository ([LINK](https://github.com/lf-lang/lingua-franca/pull/1873)).

Please feel free to comment if further modifications are required. Similar issues persist during the federated multiport setting. To address this, a separate PR will be submitted later.

### DIFF
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -130,21 +130,23 @@ lf_token_t* lf_writable_copy(lf_port_base_t* port) {
 
 ////////////////////////////////////////////////////////////////////
 //// Internal functions.
-
 void _lf_free_token_value(lf_token_t* token) {
     if (token->value != NULL) {
         // Count frees to issue a warning if this is never freed.
         _lf_count_payload_allocations--;
         // Free the value field (the payload).
-        // First check whether the value field is garbage collected (e.g. in the
-        // Python target), in which case the payload should not be freed.
-#ifndef _LF_GARBAGE_COLLECTED
         LF_PRINT_DEBUG("_lf_free_token_value: Freeing allocated memory for payload (token value): %p",
-                token->value);
-        if (token->type->destructor == NULL) {
-            free(token->value);
-        } else {
+            token->value);
+        // First check whether the token value field is not NULL
+        // in which case token's destructor should be invoked.
+        if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
+        }
+        // If Python Target is enabled and destructor is NULL
+        // Token values should be freed
+#ifndef _PYTHON_TARGET_ENABLED
+        else {
+            free(token->value);
         }
 #endif
         token->value = NULL;
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -143,11 +143,11 @@ void _lf_free_token_value(lf_token_t* token) {
         }
         // If Python Target is not enabled and destructor is NULL
         // Token values should be freed
-#ifndef _PYTHON_TARGET_ENABLED
         else {
+#ifndef _PYTHON_TARGET_ENABLED
             free(token->value);
-        }
 #endif
+        }
         token->value = NULL;
     }
 }
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -138,7 +138,6 @@ void _lf_free_token_value(lf_token_t* token) {
         LF_PRINT_DEBUG("_lf_free_token_value: Freeing allocated memory for payload (token value): %p",
             token->value);
         // First check the token's destructor field and invoke it if it is not NULL.
-        // in which case token's destructor should be invoked.
         if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
         }
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -142,7 +142,7 @@ void _lf_free_token_value(lf_token_t* token) {
         if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
         }
-        // If Python Target is enabled and destructor is NULL
+        // If Python Target is not enabled and destructor is NULL
         // Token values should be freed
 #ifndef _PYTHON_TARGET_ENABLED
         else {
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -137,7 +137,7 @@ void _lf_free_token_value(lf_token_t* token) {
         // Free the value field (the payload).
         LF_PRINT_DEBUG("_lf_free_token_value: Freeing allocated memory for payload (token value): %p",
             token->value);
-        // First check whether the token value field is not NULL
+        // First check the token's destructor field and invoke it if it is not NULL.
         // in which case token's destructor should be invoked.
         if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
--- a/python/lib/python_port.c
+++ b/python/lib/python_port.c
@@ -99,7 +99,7 @@ PyObject* py_port_set(PyObject* self, PyObject* args) {
 
     if (val) {
         LF_PRINT_DEBUG("Setting value %p.", val);
-        Py_XDECREF(port->value);
+        python_count_decrement(port->value);
 
         // 1. Call lf_new_token function to create a new token
         // 2. Set destructor and assign value to the port 
--- a/python/lib/python_port.c
+++ b/python/lib/python_port.c
@@ -51,6 +51,15 @@ void python_count_decrement(void* py_object) {
     Py_XDECREF((PyObject*)py_object);
 }
 
+/**
+ * Decrease the reference count of PyObject for output port.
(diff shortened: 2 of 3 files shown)


core/lf_token.c
```diff
@@ -130,22 +130,24 @@ lf_token_t* lf_writable_copy(lf_port_base_t* port) {
...
-        if (token->type->destructor == NULL) {
-            free(token->value);
-        } else {
+            token->value);
+        // First check whether the token value field is not NULL
+        // in which case token's destructor should be invoked.
+        if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
         }
+        // If Python Target is enabled and destructor is NULL
```
edwardalee, line 145:
```suggestion
        // If Python Target is not enabled and destructor is NULL
```

python/lib/python_port.c
```diff
@@ -36,9 +36,22 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#include "api/api.h"
+#include "api/set.h"
 
 PyTypeObject py_port_capsule_t;
 
+//////////// destructor Function(s) /////////////
+/**
+ * Called when putting token to the recycling bin. This function decreases the 
+ * reference count of PyObject and allows python to collect garbage later.
+ * @param py_object A PyObject with count 1*
```
edwardalee, line 48: To conform with our coding style, comment should be something like this;
```suggestion
 * Decrease the reference count of PyObject. When the reference count hits zero,
 * Python can free its memory.
 * @param py_object A PyObject with count 1 or greater.
```
We use the imperative tense for comments, and the comment should describe what the function does, not where it's called. (I'm not sure what you meant by count `1*`).

python/lib/python_port.c
```diff
@@ -85,13 +98,16 @@ PyObject* py_port_set(PyObject *self, PyObject *args) {
     }
 
     if (val) {
-        LF_PRINT_DEBUG("Setting value %p.", val);
+        LF_PRINT_DEBUG("Setting value lol %p.", val);
```
edwardalee, line 101: What's the meaning of `lol` here?

python/lib/python_port.c
```diff
@@ -85,13 +98,16 @@ PyObject* py_port_set(PyObject *self, PyObject *args) {
...
         Py_XDECREF(port->value);
```
edwardalee, line 102:
```suggestion
        python_count_decrement(port->value);
```

python/lib/python_port.c
```diff
@@ -85,13 +98,16 @@ PyObject* py_port_set(PyObject *self, PyObject *args) {
...
-        Py_INCREF(val);
-        // Call the core lib API to set the port
-        _LF_SET(port, val);
 
-        Py_INCREF(val);
+        // 1. Call lf_new_token function to create a new token
+        // 2. Set destructor and assign value to the port 
+        // Py_INCREF(val) is unnecessary 
+        lf_token_t* token = lf_new_token((void*)port, val, 1);
+        lf_set_destructor(port, python_count_decrement);
```
edwardalee, line 108: It's not a big deal, but setting the destructor needs to be done only once. Could it be done instead in `py_port_capsule_new`?

edwardalee (approved):
Fantastic!  Some minor suggestions included.

python/lib/python_port.c
jackykwok2024: Hello Edward,
Thanks for the comment. I considered moving this to the `py_port_capsule_new` function as well. However, during testing, I found that the `py_port_capsule_new` function wasn't called at all during the federated execution. Additionally, the `py_port_set` function is invoked once per iteration, and the destructor is indeed set only once.

python/lib/python_port.c
jackykwok2024: Fixed the typo👍

core/lf_token.c
```diff
@@ -130,22 +130,24 @@ lf_token_t* lf_writable_copy(lf_port_base_t* port) {
...
-        // First check whether the value field is garbage collected (e.g. in the
-        // Python target), in which case the payload should not be freed.
-#ifndef _LF_GARBAGE_COLLECTED
         LF_PRINT_DEBUG("_lf_free_token_value: Freeing allocated memory for payload (token value): %p",
-                token->value);
-        if (token->type->destructor == NULL) {
-            free(token->value);
-        } else {
+            token->value);
+        // First check whether the token value field is not NULL
```
lhstrh, line 140:
```suggestion
        // First check the token's destructor field and invoke it if it is not NULL.
```

core/lf_token.c
```diff
@@ -130,22 +130,24 @@ lf_token_t* lf_writable_copy(lf_port_base_t* port) {
...
+        // in which case token's destructor should be invoked.
```
lhstrh, line 141:
```suggestion
```

core/lf_token.c
```diff
@@ -130,22 +130,24 @@ lf_token_t* lf_writable_copy(lf_port_base_t* port) {
...
+        if (token->type->destructor != NULL) {
             token->type->destructor(token->value);
         }
+        // If Python Target is not enabled and destructor is NULL
+        // Token values should be freed
+#ifndef _PYTHON_TARGET_ENABLED
```
lhstrh, line 147: I think it would be more readable if the macro is moved inside of the `else` block.

python/lib/python_port.c
```diff
@@ -86,12 +99,15 @@ PyObject* py_port_set(PyObject *self, PyObject *args) {
 
     if (val) {
         LF_PRINT_DEBUG("Setting value %p.", val);
-        Py_XDECREF(port->value);
-        Py_INCREF(val);
-        // Call the core lib API to set the port
-        _LF_SET(port, val);
+        python_count_decrement(port->value);
+
+        // 1. Call lf_new_token function to create a new token
```
lhstrh, line 104: The first two lines of this comment strike me as redundant (remove?), but the third line seems to lack context. Could you elaborate?

lhstrh (commented):
Looks good! I left some minor comments...

python/lib/python_port.c
petervdonovan: I assume that it is because the reference count is already set to one when it is created by `PyArg_ParseTuple`, and then no references are saved by this function?

petervdonovan (approved):
Other than the lingering comments, this looks good to me. I like that it hooks the reference count into our destructor mechanism -- this seems like the right way to get our memory management system to interact properly with CPython's. As before, I'm really grateful that I didn't have to do this because this looks really hard.

core/lf_token.c
jackykwok2024: Agree👍

petervdonovan (changes requested):
We should not merge this yet because there are test failures; see the discussion at https://github.com/lf-lang/lingua-franca/pull/1873

python/lib/python_port.c
```diff
@@ -99,12 +108,14 @@ PyObject* py_port_set(PyObject* self, PyObject* args) {
...
         LF_PRINT_DEBUG("Setting value %p.", val);
+        Py_INCREF(val);
         python_count_decrement(port->value);
        
         lf_token_t* token = lf_new_token((void*)port, val, 1);
-        lf_set_destructor(port, python_count_decrement);
+        lf_set_destructor(port, output_port_destructor);
         lf_set_token(port, token);
-
+        Py_INCREF(val);
```
petervdonovan, line 117: Why is `Py_INCREF(val)` written twice in this commit?

If the "right" destructor decrements the output by one, then probably we should preserve the invariant that the portion of the reference count that is created by our runtime is always either zero or one, depending on whether any part of our runtime has a reference to the memory. So, we should never increment the reference count twice in a row nor decrement it twice in a row.

petervdonovan (approved):
Tests seem to be passing without a segfault in the companion PR, and the most problematic memory leak seems to be fixed. I think this is ready to merge.

lhstrh (approved):
:rocket: :rocket: :rocket:

PR #248, opened 2023-07-05
Use of correct federate ID in tracing of absent messages

ChadliaJerad:
This PR fixes the wrong federate id when tracing a received ABS message by the RTI.


petervdonovan (approved):
Thanks Chadlia!

lhstrh (approved):
LGTM! Please squash and merge when all the tests pass.

PR #249, opened 2023-07-11
Critical sections enabled outside of the context of a reactor

edwardalee:
This PR makes it possible for users to enter and exit critical sections and access time both inside and outside of the context of a reactor.  This enables resurrecting the MQTT examples in the example repo.

This has a [companion PR](https://github.com/lf-lang/lingua-franca/pull/1901) in lingua-franca.

### DIFF
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -38,23 +38,11 @@ lf_mutex_t rti_mutex;
 lf_cond_t received_start_times;
 lf_cond_t sent_start_time;
 
-/**
- * Enter a critical section where logical time and the event queue are guaranteed
- * to not change unless they are changed within the critical section.
- * this can be implemented by disabling interrupts.
- * Users of this function must ensure that lf_init_critical_sections() is
- * called first and that lf_critical_section_exit() is called later.
- * @return 0 on success, platform-specific error number otherwise.
- */
-extern int lf_critical_section_enter() {
+extern int lf_critical_section_enter(environment_t* env) {
     return lf_mutex_lock(&rti_mutex);
 }
 
-/**
- * Exit the critical section entered with lf_lock_time().
- * @return 0 on success, platform-specific error number otherwise.
- */
-extern int lf_critical_section_exit() {
+extern int lf_critical_section_exit(environment_t* env) {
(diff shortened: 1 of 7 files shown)


petervdonovan:
My understanding is that one of the key objectives of LF is to allow users to avoid the explicit use of mutexes/critical sections.

I assume that the MQTT example that you refer to is [this file](https://github.com/lf-lang/examples-lingua-franca/blob/f851ad6357a2dbaa43f2451de87ac6a2ab346287/C/src/mqtt/lib/MQTTPublisher.lf), where a mutex is acquired while accessing the MQTT API, presumably because the MQTT API is not thread-safe. If this is the case, then why not enforce mutual exclusion of accesses to the MQTT API by requiring all accesses to the MQTT API to be performed within the same reactor instance?

In order for this approach to work, it would be necessary for reactors to pass various arguments along to the MQTT reactor, and the MQTT reactor would have to pass back any handles and outputs that it gets from the API, and everything would have to be wired to that one reactor, and altogether it would make for a cumbersome way to access a simple C API. However, I think that we should consider this approach anyway because doing so would draw attention to the areas where the LF language could be improved. If the standard way of using LF is to use it to structure an application but to resort to explicit critical sections as soon as it becomes inconvenient to use LF to avoid concurrency bugs, then LF applications will not be safer than applications written in popular general-purpose languages.

erlingrj:
I agree with what you are writing, @petervdonovan. But there might still be reasons for exposing the platform API to the users. They might want to synchronize an async context in a `lf_thread_t` with either another `lf_thread_t` or with a reactor. We earlier had some debate on this here: https://github.com/lf-lang/reactor-c/issues/161

petervdonovan (approved):
OK, I will not belabor a point that has already been discussed. I am satisfied with having commented that the particular use of critical sections in the MQTT examples may not be appropriate.

edwardalee:
I think @petervdonovan 's suggestion regarding MQTT is a really good one and it may not be all that hard to implement. But I also agree with @erlingrj that at least having the possibility of accessing those functions is OK.  We don't have to encourage it, but might lead to some interesting innovations.  Maybe @petervdonovan could open an issue in the examples repo to redesign the MQTT code?

Issue #229, opened 2023-06-01
`AsyncCallbackDrop.lf` failure


lhstrh:
We have seen occasional failures of `AsyncCallbackDrop.lf`and the problem appears to be the following:

> I think the erroneous trace looks as follows: the callback acquires the lock and schedules, then the runtime acquires the lock and starts handling the event, and then the second event gets pushed on to the event queue and is handled.

There is no assurance that the the two calls to schedule happen atomically. Instead, we should not rely on the event queue to determine minimum spacing but keep a separate record of what the tag of the last-handled event was.

PR #251, opened 2023-07-20
Fix for fatal error raised during shutdown when decrementing a tag barrier that is zero

petervdonovan:
This is a follow-up to [this comment](https://github.com/lf-lang/lingua-franca/pull/1684#issuecomment-1644724836) and seems to fix an error that can be reproduced in `master` by running `PingPongDistributedPhysical` perhaps a hundred times. After making this change I have run `PingPongDistributedPhysical` more than a hundred times without errors.

The change that fixes this problem is the single-line change in `handle_stop_granted_message`; the other changes are separate.

This fix seems dangerous and looks like it can cause deadlock if increments and decrements are ordered incorrectly. However, I do not think that it will cause deadlock because it seems like the federate/RTI communication protocol is supposed to ensure that the increment happens first, if it happens at all. Furthermore, I do not see a way around a fix like this because the logic for deciding whether or not to increment the tag barrier seems nontrivial.

Although the change in `handle_tagged_message` is not necessary in order to fix this bug, I think that it does fix a separate logical error where the mutex is released but the tag barrier is not decremented when we do an early return.

### DIFF
--- a/core/federated/federate.c
+++ b/core/federated/federate.c
@@ -2031,20 +2031,19 @@ void handle_tagged_message(int socket, int fed_id) {
         // Before that, if the current time >= stop time, discard the message.
         // But only if the stop time is not equal to the start time!
         if (lf_tag_compare(env->current_tag, env->stop_tag) >= 0) {
-            lf_mutex_unlock(&env->mutex);
             lf_print_error("Received message too late. Already at stop tag.\n"
             		"Current tag is " PRINTF_TAG " and intended tag is " PRINTF_TAG ".\n"
             		"Discarding message.",
 					env->current_tag.time - start_time, env->current_tag.microstep,
 					intended_tag.time - start_time, intended_tag.microstep);
-            return;
+            goto release;
         }
 
         LF_PRINT_LOG("Calling schedule with tag " PRINTF_TAG ".", intended_tag.time - start_time, intended_tag.microstep);
         schedule_message_received_from_network_locked(env, action->trigger, intended_tag, message_token);
     }
 
-
+    release:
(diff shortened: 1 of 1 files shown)


lhstrh:
Note that the failure you're seeing in CI is due to https://github.com/lf-lang/reactor-c/issues/229

edwardalee (approved):
Wow, this is a subtle one and a great catch. The root problem is that PingPongDistributedPhysical is a nondeterministic program because of the physical connections and the way it is written, it is possible for it to call `lf_request_stop` more than once. The federate.c code contains a guard against that eventuality that can lead to this mismatch of barriers.  

Another possible fix would be to make the PingPongDistributedPhysical test more deterministic by changing:
```
  reaction(startup, serve) -> send {=
    printf("At logical time %lld, Ping sending %d.\n", lf_time_logical_elapsed(), self->pingsLeft);
    lf_set(send, self->pingsLeft--);
  =}

  reaction(receive) -> serve {=
    if (self->pingsLeft > 0) {
      lf_schedule(serve, 0);
    } else {
      lf_request_stop();
    }
  =}
```
to
```
  reaction(startup, serve) -> send {=
    printf("At logical time %lld, Ping sending %d.\n", lf_time_logical_elapsed(), self->pingsLeft);
    lf_set(send, self->pingsLeft);
  =}

  reaction(receive) -> serve {=
    if (self->pingsLeft-- > 0) {
      lf_schedule(serve, 0);
    } else {
      lf_request_stop();
    }
  =}
```
But I think @petervdonovan 's fix is better because there really is nothing preventing user code from calling `lf_request_stop()` more than once.

PR #252, opened 2023-07-21
If clock sync is on, link math

petervdonovan:
This fixes compilation errors observed on Ubuntu when the following target properties are used in a federated program:
```
  clock-sync: on,
  clock-sync-options: {
    local-federates-on: true,
  },
```


lhstrh (approved):
Great! Thanks for the fix!

PR #255, opened 2023-07-26
Python 3.11 support

cmnrd:
This PR contains @edwardalee's accompanying changes for https://github.com/lf-lang/lingua-franca/pull/1902. There wasn't a PR in reactor-c yet, so I thought I just open one.


lhstrh:
Are the changes compatible with Python 3.10? Or are we just upgrading the requirement to 3.11?

edwardalee:
These changes seem to work with both python 3.10 and 3.11. With python 3.11, we get a deprecation warning that should be easy to fix:
```
pythontarget.c:600:9: warning: 'Py_SetPath' is deprecated [-Wdeprecated-declarations]
        Py_SetPath(wcwd);
```
I think the tests are running with python 3.10, so it seems these changes work with 3.10.  There is a [companion PR in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/1902) that is needed to actually run with python 3.11.

PR #259, opened 2023-08-15
Move tracing utils

erlingrj:
Moving tracing utilities from `lingua-franca` to `reactor-c`. Also install them to `/usr/local/bin`.

PR #262, opened 2023-08-18
Updated Makefile and docs for fedsd utility

ChadliaJerad:
When installing the tracing executables to `/usr/local/bin`, `fedsd` will not work, because it is operating on a dangling symlink.

This PR provides a fix.


ChadliaJerad:
BTW, `launch-fedsd.sh` script is broken, since it relied on the paths to `fedsd.py` and `trace_to_csv` executable being known.  
Will work on this.

ChadliaJerad:
It works this way, but seems more like a hack...

edwardalee (changes requested):
One problem with the fix to the Makefile is that now a file in /usr/local/bin is a symbolic link into a user's clone of the repo.  This seems problematic to me.  Would it work instead to copy launch-fedsd.sh into /usr/local/bin (or wherever is specified with INSTALL_PREFIX)?

Also, for the other changes that try to prepend a path to trace_to_csv using an INSTALL_PREFIX argument to fedsd, this seems awkward.  If trace_to_csv is in the user's PATH, it seems like it would not be necessary.  Can we just check to see whether trace_to_csv is in the user's path and issue a friendly error message if not?

erlingrj:
I am sorry about this. I didnt test the fedsd utility. I think I agree with Edward, it is not unreasonable to demand that for  `fedsd` to work, certain executables must be on the PATH.

erlingrj:
What is, btw, the reason for having this launch shell script wrapping a Python script instead of making the Python script itself executable and putting it on the PATH?

erlingrj:
The Zephyr test failure is fixed by updating LF ref here: #263

lhstrh:
> What is, btw, the reason for having this launch shell `launch-fedsd.sh` script wrapping a Python script `fedsd.py` instead of making the Python script itself executable and putting it on the PATH?

No need for the launch script, IMO. Let make install copy the Python script and be done with it? Make sure it allows for specifying an installation prefix...

edwardalee:
> > What is, btw, the reason for having this launch shell `launch-fedsd.sh` script wrapping a Python script `fedsd.py` instead of making the Python script itself executable and putting it on the PATH?
> 
> No need for the launch script, IMO. Let make install copy the Python script and be done with it? Make sure it allows for specifying an installation prefix...

Can a python program be made executable? I was unaware of that...  How does the OS know which python to invoke?

lhstrh:
> Can a python program be made executable? I was unaware of that... How does the OS know which python to invoke?

https://realpython.com/python-shebang/

ChadliaJerad:
Just pushed a fix, where `launch-fedsd.sh` is no longer needed.
I have a question though. Since executing Python Scripts with a shebang under Windows requires an extra configuration effort, where shall we mention this? 
I thought this is the reason behind using the launch scripts...

A detail worth mentioning in this update is that one does not need saying `fedsd *.lft` anymore. Calling `fedsd` without arguments will automatically search `.lft` files. At the same time, it is still possible to operate of specific trace files that are given as arguments.

Regarding `fedsd_helper.py`, it is also copied with `fedsd.py`. Should we simply keep one file by putting the contents of the first in the second?

lhstrh:
> Just pushed a fix, where `launch-fedsd.sh` is no longer needed. I have a question though. Since executing Python Scripts with a shebang under Windows requires an extra configuration effort, where shall we mention this? I thought this is the reason behind using the launch scripts...

What configuration effort? On Windows, a bash script must be executed in WSL, and in WSL running a script with a `#!` doesn't require any configuration. The file must be executable (`+x`) though.

util/tracing/Makefile
```diff
@@ -30,8 +31,10 @@ install: trace_to_csv trace_to_chrome trace_to_influxdb
 	mv trace_to_csv $(INSTALL_PREFIX)
 	mv trace_to_chrome $(INSTALL_PREFIX)
 	mv trace_to_influxdb $(INSTALL_PREFIX)
-	ln -f -s launch-fedsd.sh $(INSTALL_PREFIX)/fedsd
-	chmod +x launch-fedsd.sh
+	cp ./visualization/fedsd.py $(INSTALL_PREFIX)
```
lhstrh, line 34: Do we want to maintain the `.py` extension after installation?

util/tracing/visualization/fedsd.py
```diff
@@ -88,9 +93,119 @@ def load_and_process_csv_file(csv_file) :
...
+
+    Args:
+     * File: the argument passed at the command line as the rti lft trace file.
+     * Array: the argument passed at the command line as array of federates lft trace files.
+    Return:
+     * File: the converted RTI trace csv file, or empty, if no RTI trace lft file is found
+     * Array: Array of files of converted federates trace csv files
+    '''
+
+    # # print("===> "+rti_lft_file+" adn "+federates_lft_files)
```
lhstrh, line 153:
```suggestion
```

util/tracing/visualization/fedsd.py
```diff
@@ -329,3 +441,5 @@ def load_and_process_csv_file(csv_file) :
 
     # Write to a csv file, just to double check
     trace_df.to_csv('all.csv', index=True)
+    print('Fedsd: Successfully generated the sequence diagram in trace_svg.html.')
+
```
lhstrh, line 445:
```suggestion

```

util/tracing/Makefile
```diff
@@ -30,8 +31,10 @@ install: trace_to_csv trace_to_chrome trace_to_influxdb
...
+	cp ./visualization/fedsd_helper.py $(INSTALL_PREFIX)
```
lhstrh, line 35: This file shouldn't end up on the path but in a `lib` directory. However, Python files are pretty small, so why not merge them?

util/tracing/Makefile
lhstrh, line 1: Isn't `INSTALL_PREFIX` meant to specify the _prefix_ of installation directories like `bin` and `lib`? In other words, if the `INSTALL_PREFIX` is `/foo/bar/`, I would expect `fedsd` to be installed in `/foo/bar/bin`. I'm not sure though because I don't often write Make files...

util/tracing/visualization/fedsd_helper.py
```diff
@@ -169,6 +169,7 @@ def svg_string_draw_arrow(x1, y1, x2, y2, label, type=''):
     str_line3 = svg_string_draw_label(x1, y1, x2, y2, label)
     return str_line1 + str_line2 + str_line3
 
+
```
lhstrh, line 172:
```suggestion
```

util/tracing/visualization/fedsd_helper.py
```diff
@@ -193,6 +194,7 @@ def svg_string_draw_side_label(x, y, label, anchor="start") :
 
     return str_line
 
+
```
lhstrh, line 197:
```suggestion
```

util/tracing/visualization/fedsd_helper.py
```diff
@@ -222,6 +224,7 @@ def svg_string_draw_dot(x, y, label) :
     str_line = str_line + '\t<text x="'+str(x+5)+'", y="'+str(y+5)+'" fill="blue">'+label+'</text>\n'
     return str_line
 
+
```
lhstrh, line 227:
```suggestion
```

util/tracing/visualization/fedsd_helper.py
```diff
@@ -239,6 +242,7 @@ def svg_string_draw_dot_with_time(x, y, time, label) :
     str_line = str_line + '\t<text x="'+str(x+5)+'", y="'+str(y+5)+'"> <tspan class="time">'+time+':</tspan> <tspan fill="blue">'+label+'</tspan></text>\n'
     return str_line
 
+
```
lhstrh, line 245:
```suggestion
```

ChadliaJerad:
> What configuration effort? On Windows, a bash script must be executed in WSL, and in WSL running a script with a `#!` doesn't require any configuration. The file must be executable (`+x`) though.

Yes yes, you're right! Same for Cygwin.
Actually, I don't have a decent understanding of LF support on Windows platform...

lhstrh (changes requested):

lhstrh:
> > What configuration effort? On Windows, a bash script must be executed in WSL, and in WSL running a script with a `#!` doesn't require any configuration. The file must be executable (`+x`) though.
> 
> Yes yes, you're right! Same for Cygwin. Actually, I don't have a decent understanding of LF support on Windows platform...

Since the introduction of WSL, there is a whole lot less to worry about. We can just tell Windows users to use WSL...

lhstrh:
To build confidence in multi-platform support, just add some tests in CI and let them run them on all platforms. I don't believe there are currently any tests, so please add those.

util/tracing/Makefile
ChadliaJerad: It is not maintained. In line 35, a symbolic link is created with the name `fedsd`.

util/tracing/visualization/fedsd.py
ChadliaJerad: A number of messages are there to see the progress.
Why remove this final one? An output example looks like this:
```
Fedsd: Successfully converted trace file rti.lft to rti.csv.
Fedsd: Successfully converted trace file federate__mid.lft to federate__mid.csv.
Fedsd: Successfully converted trace file federate__up.lft to federate__up.csv.
Fedsd: Successfully converted trace file federate__down.lft to federate__down.csv.
Fedsd: Successfully generated the sequence diagram in trace_svg.html.
```

util/tracing/Makefile
ChadliaJerad: I explicitly asked about this.
Will do.

util/tracing/Makefile
lhstrh: OK, I see what you're trying to do. But if `fedsd` is on the path, then so is `fedsd.py`. Is that the intention?
I would imagine we would want something like:
```
        mkdir -p $(INSTALL_PREFIX)/lib/fedsd
        cp ./visualization/fedsd.py $(INSTALL_PREFIX)/lib/fedsd
	ln -s $(INSTALL_PREFIX)/lib/fedsd/fedsd.py $(INSTALL_PREFIX)/bin/fedsd
        chmod +x $(INSTALL_PREFIX)/bin/fedsd
```

util/tracing/Makefile
```diff
@@ -30,8 +31,10 @@ install: trace_to_csv trace_to_chrome trace_to_influxdb
...
+	ln -f -s $(INSTALL_PREFIX)/fedsd.py $(INSTALL_PREFIX)/fedsd
```
lhstrh, line 36: Why use the `-f` flag?

util/tracing/Makefile
ChadliaJerad: If my understanding is correct, @edwardalee and @erlingrj say that it is a reasonable assumption to have all tracing utilities in the path. Unlike the first fix, this update relies on this.

util/tracing/visualization/fedsd.py
lhstrh: I'm not suggesting to remove the message, I'm suggesting to remove the excess white space below it. Did I click on the wrong line?

util/tracing/Makefile
lhstrh: That is not what I meant. My point is that you're installing stuff directly into `INSTALL_PREFIX` rather than in sub directories called `bin` and `lib`.

util/tracing/Makefile
ChadliaJerad: It is useful when one installs the tracing utilities with an `INSTALL_PREFIX` (that does not need `sudo`, and that is the path), and you are updating the utilities.
Doesn't this make sense?

util/tracing/visualization/fedsd.py
lhstrh: The convention is for the last char to be a newline, so no extra spaces.

lhstrh: GitHub marks deviations from this "end with newline" practice with red error markers in the diff. I recommend scanning the diff in the PR for obvious things like these before marking the PR ready for review.

util/tracing/Makefile
ChadliaJerad: This simply follows what was already approved in this PR #259...

erlingrj: It seems that you are right Marten. The convention is that `bin` is appended to the INSTALL_PREFIX, I will push some changes addressing that to this commit

util/tracing/Makefile
```diff
@@ -27,11 +29,12 @@ trace_to_influxdb: trace_to_influxdb.o trace_util.o
 	$(CC) -o trace_to_influxdb trace_to_influxdb.o trace_util.o $(LIBS)
 
 install: trace_to_csv trace_to_chrome trace_to_influxdb
-	mv trace_to_csv $(INSTALL_PREFIX)
-	mv trace_to_chrome $(INSTALL_PREFIX)
-	mv trace_to_influxdb $(INSTALL_PREFIX)
-	ln -f -s launch-fedsd.sh $(INSTALL_PREFIX)/fedsd
-	chmod +x launch-fedsd.sh
+	mv trace_to_csv $(BIN_INSTALL_PATH)
```
lhstrh, line 32:
```suggestion
	cp trace_to_csv $(BIN_INSTALL_PATH)
```
https://stackoverflow.com/questions/39892692/how-to-implement-make-install-in-a-makefile

util/tracing/Makefile
```diff
@@ -27,11 +29,12 @@ trace_to_influxdb: trace_to_influxdb.o trace_util.o
...
+	mv trace_to_chrome $(BIN_INSTALL_PATH)
```
lhstrh, line 33:
```suggestion
	cp trace_to_chrome $(BIN_INSTALL_PATH)
```

util/tracing/Makefile
```diff
@@ -27,11 +29,12 @@ trace_to_influxdb: trace_to_influxdb.o trace_util.o
...
+	mv trace_to_influxdb $(BIN_INSTALL_PATH)
```
lhstrh, line 34:
```suggestion
	cp trace_to_influxdb $(BIN_INSTALL_PATH)
```

lhstrh (commented):
Modulo suggestions, this looks good, but there are still no tests, so we don't have any assurance that this works and will keep working...

lhstrh:
Since it appears that what's in `main` is broken and this is a fix, let's merge this. But we do need tests.

lhstrh (approved):

PR #265, opened 2023-08-28
Fix launch script

petervdonovan:
It may also be necessary to make the installed `fedsd` symlink point to the right location. I am not sure if I will have time to finish this.


ChadliaJerad:
This has been addressed in #262 and we even did refactor  `fedsd` installation and usage. Doesn't it?

petervdonovan:
You're right! I just hadn't seen that.

PR #269, opened 2023-09-02
Mixed microsteps and time

petervdonovan:
This attempts to fix an STP violation that was reported over email. I believe that it actually fixes two distinct bugs which manifested in the same program.

It is maybe not yet ready for review. I think it may be justified to add another test case to the main repository because it does not seem like we have adequate test cases for federated programs that use both the time dimension and the microstep dimension.

### DIFF
--- a/include/core/tag.h
+++ b/include/core/tag.h
@@ -111,6 +111,15 @@ int lf_tag_compare(tag_t tag1, tag_t tag2);
 tag_t lf_delay_tag(tag_t tag, interval_t interval);
 
 /**
+ * Return the latest tag that cannot have an event provided that
+ * - `antitag` does not have an event, and
+ * - if an event occurs at time `x`, then there exists a tag `antitag` such that
+ *   `x = antitag + interval`, where `+` is the usual `+` that we use for tags
+ *   which is neither commutative nor a group operation.
+ */
+tag_t lf_delay_antitag(tag_t antitag, interval_t interval);
+
+/**
  * Return the current logical time in nanoseconds.
  * On many platforms, this is the number of nanoseconds
  * since January 1, 1970, but it is actually platform dependent.
--- a/include/core/tag.h
+++ b/include/core/tag.h
@@ -94,16 +94,18 @@ int lf_tag_compare(tag_t tag1, tag_t tag2);
 
 /**
  * Delay a tag by the specified time interval to realize the "after" keyword.
- * If either the time interval or the time field of the tag is NEVER,
- * return the unmodified tag.
- * If the time interval is 0LL, add one to the microstep, leave
- * the time field alone, and return the result.
+ * Any interval less than 0 (including NEVER) is interpreted as "no delay",
+ * whereas an interval equal to 0 is interpreted as one microstep delay.
+ * If the time field of the tag is NEVER or the interval is negative,
+ * return the unmodified tag. If the time interval is 0LL, add one to
+ * the microstep, leave the time field alone, and return the result.
  * Otherwise, add the interval to the time field of the tag and reset
- * the microstep to 0.
- * If the sum overflows, saturate the time value at FOREVER.
- *
- * Note that normally it makes no sense to call this with a negative
- * interval (except NEVER), but this is not checked.
+ * the microstep to 0. If the sum overflows, saturate the time value at
+ * FOREVER. For example:
+ * - if tag = (t, 0) and interval = 10, return (t + 10, 0)
+ * - if tag = (t, 0) and interval = 0, return (t, 1)
+ * - if tag = (t, 0) and interval = NEVER, return (t, 0)
+ * - if tag = (FOREVER, 0) and interval = 10, return (FOREVER, 0)
  *
  * @param tag The tag to increment.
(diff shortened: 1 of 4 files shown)


include/core/tag.h
```diff
@@ -110,6 +110,15 @@ int lf_tag_compare(tag_t tag1, tag_t tag2);
...
 tag_t lf_delay_tag(tag_t tag, interval_t interval);
 
+/**
+ * Return the latest tag that cannot have an event provided that
+ * - `antitag` does not have an event, and
+ * - if an event occurs at time `x`, then there exists a tag `antitag` such that
+ *   `x = antitag + interval`, where `+` is the usual `+` that we use for tags
+ *   which is neither commutative nor a group operation.
+ */
+tag_t lf_delay_antitag(tag_t antitag, interval_t interval);
```
lhstrh, line 120: The description makes sense, but I find the terminology confusing. The fact that the given tag doesn't have an event doesn't change anything about the tag, so I don't see why we should be calling it an "antitag." Maybe it would be better to name the method more descriptively? Also, are you use that you meant "latest", not "earliest"?

petervdonovan: > I don't see why we should be calling it an "antitag."

I can write `absent_tag` to indicate that it is the tag of something that does not happen.

> Also, are you use that you meant "latest", not "earliest"?

Yes, I meant "latest." To find the earliest tag at which nothing can happen is trivial -- that is the tag that is as far as possible into the past (nothing is going to happen in the past). But actually what I wrote was not technically right. I meant "The maximum of the set of all tags at which nothing can happen," which is similar to "latest"

lhstrh: Now I'm even more confused :laughing: 

What do you mean with "tags at which nothing can happen"? Something can happen at any tag in the future. The absence of an event in the queue at a particular tag does not rule out that an event with that tag could later be added.

petervdonovan: > Something can happen at any tag in the future.

The whole design of federated execution is based on the fact that is not true. We have after delays. If nothing happens on one side of an after delay at time $g$, and the delay is $h$ then nothing will happen on the other side of the after delay before time $g + h$.

lhstrh: That is true, but but timers and actions can cause events to be inserted in the queue. By "nothing happens" you mean with respect to a particular upstream federate? I'm not trying to be dense, but the more you explain about the use of this function, the easier it will be to come up with more fitting terminology...

petervdonovan: Nothing will happen **at the port** on the other side of the after delay.

I have no strong opinions about the terminology and would be OK with whatever terminology we choose. It is good to strike a reasonable balance between choosing sensible terminology and spending too much time debating over terminology.

include/core/tag.h
```diff
@@ -110,6 +110,15 @@ int lf_tag_compare(tag_t tag1, tag_t tag2);
  */
 tag_t lf_delay_tag(tag_t tag, interval_t interval);
 
+/**
+ * Return the latest tag that cannot have an event provided that
```
lhstrh, line 114: How about something like:
"Given a `next_event_tag` of an upstream event producer and a `delay` associated the connection to that event producer, return the largest tag smaller than `next_event_tag + delay` for which input from the producer can be assumed to be absent."

The name of the function could then be `lf_largest_tag_smaller_than_sum` or something like that...

lhstrh: Is there actually any significance to the absence of input? Could it just be described as a numerical operation on tags? "Get the largest tag smaller than the sum of the given `tag` and `delay`?"

petervdonovan: That sounds like a good solution!

edwardalee: I have changed lf_delay_antitag to lf_delay_strict and adjusted the implementation and the documentation. I believe this addresses the concerns and gives a simpler interpretation.

include/core/tag.h
edwardalee: @petervdonovan identified and fixed a real bug in the RTI. This should also resolve an issue I had been seeing that the RTI was sending way too many PTAG messages when these were not needed.  I have adjusted the terminology and implementation and believe this is ready to merge (modulo tests passing).

edwardalee:
It would be great if @petervdonovan could look at my changes, but I can't seem to request him as a reviewer because he is the author of this PR.

edwardalee (approved):
I have adjusted the terminology and implementation. This PR resolves the problem I reported. I have added a test in companion [PR in the lingua-franca repo](https://github.com/lf-lang/lingua-franca/pull/2002).

petervdonovan:
The changes look good to me except I noticed a few tests are failing. I am sure the problem it is only a minor detail.

petervdonovan:
I reproduced the problem and partially reverted the latest commit; this seemed to work on my machine. I think this shows that the problem is not as simple as strictness vs. nonstrictness and instead is inherently tied to the non-injectivity of right addition. When you are just adding a microstep, right addition acts like it is injective and so in that case we do not need strictness.

I think we should probably just merge this since it is so easy to break it or to make the inline documentation confusing in subtle ways :sweat_smile:

PR #271, opened 2023-09-14
Fedsd fix self_id

ChadliaJerad:
This PR fixes a bug that @jackykwok2024 reported when using `fedsd` utility with `Python` target.

Previously, we relied on the last event of a trace (after processing) not being about advancing time. A federate will send a resignation message after the last time advance. This behavior is valid in the C target, but it is not the case in the Python target. 

In this PR, `fedsd` retrieves the correct `self_id` of a federate.

### DIFF
--- a/util/tracing/visualization/fedsd.py
+++ b/util/tracing/visualization/fedsd.py
@@ -334,10 +334,8 @@ def load_and_process_csv_file(csv_file) :
     # which boils up to having 'RTI' in the 'event' column
     df = df[df['event'].str.contains('Sending|Receiving|Scheduler advancing time ends') == True]
 
+    # Determine the "self id" in the trace file based on the first 'Receiving' or 'Sending' message (or use -1, the id of the RTI, if there is none).
     id = -1
-    # Fix the parameters of the event 'Scheduler advancing time ends'
-    # where 'self_id' should not be -1, but rather the value from a sending or 
-    # receiving event
     for index, row in df.iterrows():
         if ('Sending' in row['event'] or 'Receiving' in row['event']) :
             id = row['self_id']
--- a/util/tracing/visualization/fedsd.py
+++ b/util/tracing/visualization/fedsd.py
@@ -334,9 +334,14 @@ def load_and_process_csv_file(csv_file) :
     # which boils up to having 'RTI' in the 'event' column
     df = df[df['event'].str.contains('Sending|Receiving|Scheduler advancing time ends') == True]
 
+    id = -1
     # Fix the parameters of the event 'Scheduler advancing time ends'
-    # We rely on the fact that the first row of the csv file cannot be the end of advancing time
-    id = df.iloc[-1]['self_id']
+    # where 'self_id' should not be -1, but rather the value from a sending or 
+    # receiving event
+    for index, row in df.iterrows():
+        if ('Sending' in row['event'] or 'Receiving' in row['event']) :
+            id = row['self_id']
+            break
     df['self_id'] = id
     df = df.astype({'self_id': 'int', 'partner_id': 'int'})
 


edwardalee (approved):
LGTM!

lhstrh (commented):
I looked at this, but I have to admit that I neither understand the original comment/code nor the change. In the new comment, it says that "`'self_id'` should not be `-1`," but the former is a string (which happens to be used as a key) and the latter is a number, so they can never be equal... I'm also unclear on what _should_ means to convey. Are you trying to describe an invariant? I don't see an actual check where anything gets compared against `-1`, either, so I'm a bit confused... The for loop appears to assign to `id` the first `'Receiving'` event that it can find. Could you explain why it does that?

ChadliaJerad:
Sure!
Every federate trace file, after being converted to a csv file, is loaded as a dataframe (`df`).
A preprocessing will only keep the rows containing `Sending` messages, `Receiving` messages, or `Scheduler advance time` in `df`. Only these rows are needed to build the sequence diagram.

In order to identify the interacting components (federates and RTI), `df` uses the columns `self_id` and `partner_id`. The former specifies the federate whose trace is being processed, while the latter specifies the partner (which can be the RTI  or another federate). The id of the RTI is -1, while the id of a federate is >= 0. 

The event `Scheduler advance time`  has -1 in `self_id`. Consequently, in order to identify whose trace file it is, we need to read the `self_id` of a sending or receiving message.
Note that initially, events have different interpretations of these values.

The fix checks that we are reading the `self_id` from a message event. Otherwise, the trace file will be misinterpreted as the RTI's. Before the fix, the last row was used, and the comment explicitly said: "# We rely on the fact that the first row of the csv file cannot be the end of advancing time" (first should be last, actually!)

A counter-example was found, that's why this fix correctly reads the actual `self_id`.

lhstrh:
OK, thanks, that clear it up!

I would then make a change a long these lines:

```
    # Determine the the "self id" based on the first `Receiving` message (or use -1, the id of the rti, if there is none).
    id = -1
    for index, row in df.iterrows():
        if ('Sending' in row['event'] or 'Receiving' in row['event']) :
            id = row['self_id']
            break
```

Does that make sense?

ChadliaJerad:
A minor tweak of the tweak:
```
# Determine the "self id" in the trace file based on the first `Receiving` or `Sending` message (or use -1, the id of the RTI, if there is none).
```
Ok?

lhstrh:
> A minor tweak of the tweak:
> 
> ```
> # Determine the "self id" in the trace file based on the first `Receiving` or `Sending` message (or use -1, the id of the RTI, if there is none).
> ```
> 
> Ok?

:+1:

util/tracing/visualization/fedsd.py
```diff
@@ -334,9 +334,12 @@ def load_and_process_csv_file(csv_file) :
     # which boils up to having 'RTI' in the 'event' column
     df = df[df['event'].str.contains('Sending|Receiving|Scheduler advancing time ends') == True]
 
-    # Fix the parameters of the event 'Scheduler advancing time ends'
-    # We rely on the fact that the first row of the csv file cannot be the end of advancing time
-    id = df.iloc[-1]['self_id']
+    # Determine the "self id" in the trace file based on the first 'Receiving' or 'Sending' message (or use -1, the id of the RTI, if there is none).
```
lhstrh, line 337:
```suggestion
    # Determine the "self id" in the trace file based on the first 'Receiving' message (or use -1, the id of the RTI, if there is none).
```
Looks like we're not checking for 'Sending' messages?

edwardalee: Looks to me like the code checks for both.

ChadliaJerad: We do!
The `if` statement in the `for` loop says (Line 340):
```
if ('Sending' in row['event'] or 'Receiving' in row['event']) :
``` 
that is, if `row['event']` has `Sending` *OR* `Receiving` in it.

lhstrh: Oh, missed that. Looks good!

lhstrh (approved):

PR #272, opened 2023-09-15
Do not link with thread library when we are compiling for Zephyr

erlingrj:
Another small fix. When compiling for certain boards I get a linker error when it tries to link with the thread library. I do not think it is necessary to link this when we are linking with the Zephyr RTOS


lhstrh (approved):
Sounds reasonable. I suppose that Zephyr provides its own?

PR #273, opened 2023-09-18
Fix argparse for fedsd

erlingrj:
The docs are not up-to-date, fedsd doesnt work with `fedsd *.lft`, you have to do `fedsd -r rti.lft -f fed1.lft fed2.lft`.

This fixes the argparse setup so that the command line parsing works properly. Previously it turned the list of federate lfts into a double list [[lft1, fft2, ...,]] which crashed the rest of the script

### DIFF
--- a/util/tracing/README.md
+++ b/util/tracing/README.md
@@ -23,8 +23,9 @@ directory.
 ```
     sudo make install
 ```
-Will install the tracing executables to `/usr/local/bin` to install them to a different location, use the `INSTALL_PREFIX` flag, e.g.
+Will install the tracing executables to the `bin` directory under `/usr/local/`. To install them to a different location, use the `INSTALL_PREFIX` flag, e.g.
 
 ```
-  make install INSTALL_PREFIX=~/.local/bin
+  make install INSTALL_PREFIX=~/.local/
 ```
+which will install the executables under `~/.local/bin/`.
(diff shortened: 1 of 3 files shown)


ChadliaJerad:
> The docs are not up-to-date, fedsd doesn't work with `fedsd *.lft`, you have to do `fedsd -r rti.lft -f fed1.lft fed2.lft`.
> 
In the last update, `fedsd` does not work anymore with `fedsd *.lft`, but one needs to simply call `fedsd` instead. The [README.md](https://github.com/lf-lang/reactor-c/blob/main/util/tracing/visualization/README.md) file (under `util/tracing/visualization`) was updated accordingly. It says:
````
Once the federation stopped executing, running `fedsd` will operate on all the `.lft` files in the current directory:
````
and 
````
It is also possible to operate on specific files. In such a case, run `fedsd` with `-r` flag to provide the RTI trace file, and `-f` flag to provide the list of federate trace files. Both arguments are optional.
````
Actually, I didn't see any value in explicitly saying `*.lft`, if we need to use all local files. 
Does this make sense? 
  
> This fixes the argparse setup so that the command line parsing works properly. Previously it turned the list of federate lfts into a double list [[lft1, fft2, ...,]] which crashed the rest of the script

Apologies for the overlook. Thanks for the fix!

ChadliaJerad (approved):

edwardalee (approved):
LGTM.

erlingrj:
@ChadliaJerad, these are the docs I am referring to: https://www.lf-lang.org/docs/handbook/tracing?target=c#tracing-federated-programs

PR #274, opened 2023-09-18
Use real-time sockets in federated execution

erlingrj:
Got a bug report from @lhstrh which I was able to reproduce in the following trivial federated program:

```
target C {
    timeout: 100 msec
}
reactor Src {
    timer t (0, 50 msec)
    output out:int
    reaction(t) -> out {=
        lf_set(out, 42);
    =}
}
reactor Sink {
    input in:int 
    reaction(in) {=
        lf_print("Got %d at %li", in->value, lf_time_physical_elapsed());
    =}
}
federated reactor {
    src = new Src()
    sink = new Sink()

    src.out -> sink.in
}
```

I observed a strange ~40msec latency between Src and Sink. I tracked it down be a communication latency between Src and the RTI. After alot of digging I have found out that there is something called Nagle algorithm, enabled by default, which bundles short TCP messages together to avoid too much network traffic. So it was delaying our messages. In this draft I have just added socket options for disabling with `TCP_NODELAY`. It is still a draft because we can discuss which sockets this should apply for, maybe all? After we decide, lets put creation of sockets in `net_util.c` so we dont litter federate.c with these low-level socket option details.

@lhstrh this PR should fix your issue. You should recompile and reinstall the RTI also based on this PR.

### DIFF
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -69,7 +69,7 @@ int create_server(int32_t specified_port, uint16_t port, socket_type_t socket_ty
             lf_print_error_and_exit("Failed to disable Nagle algorithm on socket server.");
         }
     } else if (socket_type == UDP) {
-        socket_descriptor = socket(AF_INET, SOCK_DGRAM, 0);
+        socket_descriptor = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
         // Set the appropriate timeout time
         timeout_time = (struct timeval){.tv_sec = UDP_TIMEOUT_TIME / BILLION, .tv_usec = (UDP_TIMEOUT_TIME % BILLION) / 1000};
     }
--- a/core/federated/RTI/rti_lib.c
+++ b/core/federated/RTI/rti_lib.c
@@ -24,6 +24,7 @@
  */
 
 #include "rti_lib.h"
+#include <netinet/tcp.h>
 #include <string.h>
 
 // Global variables defined in tag.c:
@@ -53,6 +54,20 @@ int create_server(int32_t specified_port, uint16_t port, socket_type_t socket_ty
     int socket_descriptor = -1;
     if (socket_type == TCP) {
         socket_descriptor = socket(AF_INET, SOCK_STREAM, 0);
+
+        // Disable Nagle algorithm which bundles together small TCP messages to
+        //  reduce network traffic
+        int flag = 1;
+        int result = setsockopt(socket_descriptor,            /* socket affected */
+                                IPPROTO_TCP,     /* set option at TCP level */
+                                TCP_NODELAY,     /* name of option */
+                                (char *) &flag,  /* the cast is historical
+                                                        cruft */
+                                sizeof(int));    /* length of option value */
+        
+        if (result < 0) {
+            lf_print_error_and_exit("Failed to disable Nagle algorithm on socket server.");
+        }
     } else if (socket_type == UDP) {
         socket_descriptor = socket(AF_INET, SOCK_DGRAM, 0);
         // Set the appropriate timeout time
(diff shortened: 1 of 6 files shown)


erlingrj:
We should also investigate whether delayed ACKs can be problematic for us also: https://en.wikipedia.org/wiki/Nagle%27s_algorithm#Interaction_with_delayed_ACK

erlingrj:
The `DistributedStopDecentralized.lf` test now fails. This means either 1) I made a mistake here, 2) We shouldnt use TCP_NODELAY for federat-to-federate sockets or 3) Race condition in federated infrastructure related to stop requests which surface with this change

edwardalee:
Using the debugger, I just found a serious bug in federate.c where a local variable "result" was shadowed. I've pushed a fix to this bug to this branch, but this fix should probably be merged with main regardless.  It is interesting that the bug did not show up until @erlingrj changed the sockets to use TCP_NODELAY. In any case, at least on my machine, `DistributedStopDecentralized.lf` does not deadlock anymore.

core/federated/RTI/rti_lib.c
```diff
@@ -53,6 +54,20 @@ int create_server(int32_t specified_port, uint16_t port, socket_type_t socket_ty
     int socket_descriptor = -1;
     if (socket_type == TCP) {
         socket_descriptor = socket(AF_INET, SOCK_STREAM, 0);
+
+        // Disable Nagle algorithm which bundles together small TCP messages to
+        //  reduce network traffic
+        int flag = 1;
+        int result = setsockopt(socket_descriptor,            /* socket affected */
+                                IPPROTO_TCP,     /* set option at TCP level */
+                                TCP_NODELAY,     /* name of option */
```
lhstrh, line 63: This code is copied three times in this PR. Should we just put this in a library function named `int disable_delayed_transmission(int sockfd)` or something like that?

lhstrh:
> Using the debugger, I just found a serious bug in federate.c where a local variable "result" was shadowed. I've pushed a fix to this branch.

I think the bug was in Erling's change, not the original code, right?

edwardalee:
Yes, the bug was not in the original code (I initially thought it might have been, but nothing would have worked).

core/federated/net_util.c
```diff
@@ -54,6 +56,34 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /** Number of nanoseconds to sleep before retrying a socket read. */
 #define SOCKET_READ_RETRY_INTERVAL 1000000
 
+int create_real_time_tcp_socket_errexit() {
+    int sock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    if (sock < 0) {
+        lf_print_error_and_exit("Could not open TCP socket. Err=%d", sock);
+    }
+    // Disable Nagle algorithm which bundles together small TCP messages to
```
lhstrh, line 64:
```suggestion
    // Disable Nagle's algorithm which bundles together small TCP messages to
```

erlingrj:
After Christian's fix of the error reporting system, I will now investigate this test failure and fix it so we can merge it soon.

erlingrj:
@lhstrh, I got a test failure on a decentralized federated program on macOS.  No debug output. Unable to reproduce in Ubuntu. Went away when re-running the tests. I have taken a close look at this PR and only  reason I can come up with is either:
1) There is a race condition in decentralized federated which sometimes is triggered now because we have less delays
2) Some unrelated CI issue is at fault.

lhstrh:
My bet would be on there being something fishy with the coordination... I don't think we can blame CI for this one (nor the changes in this PR).

erlingrj:
Then I will merge this and we will have to look at federated race conditions at another time

PR #275, opened 2023-09-18
Guard against unconnected outputs

edwardalee:
This addresses [issue 1943 in lingua-franca](https://github.com/lf-lang/lingua-franca/issues/1943), where unconnected outputs lead to segmentation fault.


lhstrh (approved):
Great! Glad this is a relatively easy fix.

Issue #87, opened 2022-06-10
PQueue priorities sometimes need more bits


petervdonovan:
**This post is motivated by the following comment:**

If I understand, the reason why this procedure (and the one above it) are complicated is that [pqueue priorities have to be 64 bits](https://github.com/lf-lang/reactor-c/blob/da27154663ceb4c877d896d3009df869759bb8a9/core/utils/pqueue.h#L45), which makes it hard to sort items by **tag** instead of **time**. This implementation looks like it might be complicated, and it looks like it might have suboptimal time complexity in programs that frequently schedule events a microstep in the future.

I [have already suggested](https://github.com/lf-lang/reactor-c/pull/16/files#r787034668) that attempts to cram priorities into a single word might not be serving us very well; maybe now is a good time to reconsider that?

_Originally posted by @petervdonovan in https://github.com/lf-lang/reactor-c/pull/61#discussion_r889619223_

I would like to emphasize that this is not the only area in which I am skeptical of our priority queue data structure. @lhstrh has, I think, mentioned calendar queues as an alternative -- if I understand, this is an exotic data structure that is fundamentally different from the traditional binary heap... I have also been unhappy with the extensive use of function pointers, which make function inlining impossible. The `find_equal_same_priority` function has also caused us some grief (time complexity stuff), as we have discussed in the scheduling channel on Slack... What I am saying here is that **a fix for this shortcoming might as well be accompanied by an entirely new data structure,** rather than just one ad hoc patch.

Hokeun's and Yunsang's recent work might be a prerequisite for this, since our benchmarks are not representative of what we expect LF programs to look like. A few days ago Soroush explained to me that typical LF programs might not actually make extensive use of the [`schedule(next, 0)`](https://github.com/lf-lang/benchmarks-lingua-franca/blob/386858370154a05efb725163d60f9bc69c94377a/C/Savina/src/concurrency/Banking.lf#L144) idiom that appears in our benchmarks. [Edward has also commented on our lack of representative performance assessments for the event queue.](https://github.com/lf-lang/lingua-franca/discussions/805#discussioncomment-1822562)

PR #276, opened 2023-09-19
Draft: Priority queue is compatible with data types larger than `unsigned long long`

byeong-gil:
This solves [this issue](https://github.com/lf-lang/reactor-c/issues/87#issuecomment-1717083825).


byeong-gil:
I attempted to exploit the `_Generic` macro function to implement the priority queue that is compatible with multiple types. However, converting `unsigned long long` to the pointer and saving various data types seems better. Close this PR.

PR #277, opened 2023-09-23
Static update instr

erlingrj:


lsk567 (approved):
Thanks for working on this, @erlingrj ! I like the solution of using `union`. It is quite elegant. :-)

lsk567:
I see a `DO NOT MERGE` in the commit. Does it mean this is not ready for merging into the feature branch?

erlingrj:
The changes of that commit is already in main/master. The best would be to bring the static-schedule branch up-to-date with master first. Then I would remove this commit from this PR. Alternatively we keep the static-schedule branch stale and merge this in. And deal with it later, it will be a merge conflict.

lsk567:
Ok, sounds good. I will merge bring the `static-schedule` branch up-to-date.

Issue #283, opened 2023-10-05
`trace_to_csv` enhancements 


mkhubaibumer:
`trace_to_csv` is a good tool to convert trace files in csv format but for bulk runs with tracing this tool doesn't provide enough options to be practical

 - This tool needs to provide the ability to process all trace files in a directory including all sub-directories for bulk runs
 - This tool needs to provide custom filtering options to user to filter out events of specific type/types

PR #284, opened 2023-10-05
283 trace to csv improvements

mkhubaibumer:
Implementation of #283


petervdonovan:
These look like good changes to me, but work is in progress to give the user even more control over the behavior of the tracing mechanism. It is possible that this more general solution will address the use case covered by this PR, and other use cases as well.

In any case, recursively searching for trace files after bulk runs will be useful, so I will get back to this PR once the infrastructure looks more like what we will want long-term.

PR #286, opened 2023-10-06
Update lingua-franca-ref.txt

erlingrj:
This should pass tests when https://github.com/lf-lang/lingua-franca/pull/2037 is merged into lingua-franca...


erlingrj:
I update the lf-ref in #288

PR #288, opened 2023-10-07
Fix overflow bug affecting 32-bit Linux platforms

erlingrj:
This bug was reported in this issue: https://github.com/lf-lang/lingua-franca/issues/2034

Another reminder of how difficult it is to juggle time representations on different architectures. In general, unless we are dealing with a type of explicit widht, like uint64_t or instant_t, then we must consider that it might be 32-bit on a 32-bit architecture. This goes for any literal value also.

### DIFF
--- a/core/platform/lf_unix_clock_support.c
+++ b/core/platform/lf_unix_clock_support.c
@@ -3,6 +3,7 @@
 #include <errno.h>
 
 #include "platform.h"
+#include "util.h"
 #include "lf_unix_clock_support.h"
 
 /**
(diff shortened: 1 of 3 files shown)


erlingrj:
The test: `DecentralizedP2PUnbalancedTimeoutPhysical` fails non-deterministically. Also before this PR. So, I think we should consider this a different bug and address it separately

core/platform/lf_unix_clock_support.c
```diff
@@ -50,6 +51,15 @@ void calculate_epoch_offset(void) {
 
 void _lf_initialize_clock() {
     calculate_epoch_offset();
+
+    struct timespec res;
+    // Adjust the clock by the epoch offset, so epoch time is always reported.
```
erlingrj, line 56:
```suggestion
```

erlingrj:
Should we get this little fix merged, @edwardalee? I have tested in on a RPi4 (with a 32bit OS).

edwardalee (approved):
LGTM!

PR #293, opened 2023-10-21
 Fix Segfault for Python 3.11

jackyk02:
**Issue**
Segfault due to Python 3.11 changes to object layout and memory allocation methods.

**Background**
Python 3.11 introduced changes to optimize CPython. One of the significant changes, as mentioned in [this PR](https://github.com/python/cpython/pull/29879), modifies the object layout to place pointers to dict and values immediately before the GC header. This change potentially affected the memory layout of custom python objects.

Before:
Untitled1
After:
Untitled2
Moreover, this commit removed ```_PyObject_GC_Calloc``` and ```_PyObject_GC_Malloc```, affecting the memory allocation for objects that participate in garbage collection.

**Solution**
 The new memory allocation relies on ```PyObject_New``` instead of ```PyObject_GC_New```.

### DIFF
--- a/python/lib/pythontarget.c
+++ b/python/lib/pythontarget.c
@@ -437,7 +437,7 @@ void destroy_action_capsule(PyObject* capsule) {
 PyObject* convert_C_port_to_py(void* port, int width) {
     // Create the port struct in Python
     PyObject* cap =
-        (PyObject*)PyObject_GC_New(generic_port_capsule_struct, &py_port_capsule_t);
+        (PyObject*)PyObject_New(generic_port_capsule_struct, &py_port_capsule_t);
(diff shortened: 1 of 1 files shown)


edwardalee (approved):
LGTM!

petervdonovan:
@Jakio815 seems to have been blocked because this PR causes segfaults in `src/federated/DistributedLoopedAction.lf` and `src/federated/DistributedLoopedPhysicalAction.lf`. It is possible that we should run the Python tests as a requirement for merging changes in `reactor-c`.

lhstrh:
I also noticed that `main` of `reactor-c` is failing, but probably for a different reason.

PR #294, opened 2023-10-24
Refactor zephyr timing implementation into two separate files

erlingrj:
Clean up Zephyr platform implementation:
- Remove the use of preprocessor to conditionally compile the two different timing APIs
- Implement the kernel timing API and counter timing API in two separate files
- remove the `_lf_` prepending of all the static local variables. They are static and local to that file, so no need to add long complicated variable names that are hard to parse 

Closes https://github.com/lf-lang/lingua-franca/issues/2063

### DIFF
--- a/core/platform/CMakeLists.txt
+++ b/core/platform/CMakeLists.txt
@@ -9,6 +9,8 @@ lf_macos_support.c
 lf_windows_support.c
 lf_nrf52_support.c
 lf_zephyr_support.c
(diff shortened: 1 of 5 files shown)


lhstrh (approved):
Nice! Looks great to me. I guess this doesn't require any changes in the code generator?

erlingrj:
Correct, no changes required in `lingua franca`

PR #297, opened 2023-10-28
Use env command rather than /usr/bin/python3

edwardalee:
Directly invoking /usr/bin/python3 bypasses any user setup for python and may invoke an outdated version of python. This PR uses the `env` command instead.


edwardalee:
Merging this small change so it doesn't get lost.

PR #299, opened 2023-10-30
Zephyr: Update default thread stack size to 2KB

erlingrj:
There was a test failure due to a stack overflow. I am going to think a bit about how we can approach this in a better way. For now, increase default stack size by x2.


lhstrh (approved):
LGTM!

PR #300, opened 2023-11-03
Token double free as error

erlingrj:
Instead of printing a warning. Treat a double-freeing of a token as a critical error. Or is there a good argument for allowing those, @edwardalee?


lhstrh:
@edwardalee, :+1: or :-1:?

PR #302, opened 2023-11-07
initial test for support for STM32

naichenzhao:


lingua-franca-ref.txt
```diff
@@ -1 +1,2 @@
 single-threaded
```
lhstrh, line 1:
```suggestion
stm32
```

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
+#if defined(PLATFORM_STM32F4)
+/*************
+ I hope this software works LOL
```
lhstrh, line 3: :rofl:

petervdonovan:
@lhstrh marked this as ready for review, but I'm not sure I remember why since IIRC we agreed that we are waiting for a cross-compilation smoke test (just to show STM32 support building successfully on a regular laptop). I will review this when the tests are passing and such a smoke test is available; feel free to ping me if I forget.

lhstrh:
I was being optimistic. We hit some snags with setting up tests.

lhstrh (dismissed):
@naichenzhao, based on a cursory review, I think that we should try to get this merged soon, but, please:
 - sanitize your comments :-) (suggestions provided)
 - resolve the conflicts

@elgeeko1 (ex-BlackRock Neurotech) is willing to do a deeper review, which you might appreciate given the emphasis of safety in your project. If you can do the requested clean ups this weekend, he'll complete the review early next week. Let's aim to have this merged by the end of next week.

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
...
+}
+
+//  + -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- +
+//  | Other functions I found -> taken from the generated main.c
+//  + -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- +
+void lf_SystemClock_Config(void) {
+    RCC_OscInitTypeDef RCC_OscInitStruct = {0};
+    RCC_ClkInitTypeDef RCC_ClkInitStruct = {0};
+
+    /** Configure the main internal regulator output voltage
```
elgeeko1, line 213: Use `//` instead of `/**` blocks here.

include/core/platform/lf_STM32f4_support.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <stm32f4xx_hal.h>
+
+// Defines for formatting time in printf for pico
+#define PRINTF_TAG "(" PRINTF_TIME ", " PRINTF_MICROSTEP ")"
+#define PRINTF_TIME "%lld"
+#define PRINTF_MICROSTEP "%d"
+
+#define LF_TIME_BUFFER_LENGTH 80
+#define _LF_TIMEOUT 1
```
elgeeko1, line 17: What units is this #def? Should it be a const instead?

include/core/platform/lf_STM32f4_support.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifdef LF_THREADED
+#error "I have no idea how to support threading"
```
elgeeko1, line 20: Suggest, "Threading is not supported on this architecture."

include/core/platform.h
```diff
@@ -70,6 +70,8 @@ int lf_critical_section_exit(environment_t* env);
     #include "platform/lf_zephyr_support.h"
 #elif defined(PLATFORM_NRF52)
     #include "platform/lf_nrf52_support.h"
+#elif defined(PLATFORM_STM32)
+#include "platform/lf_STM32f4_support.h"
```
elgeeko1, line 74: Code style, should indent.

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
...
+        // Exit when the timer is up or there is an exception
+    } while (!_lf_async_event && (now < wakeup_time));
+
+    // Disable interrupts again on exit
+    lf_disable_interrupts_nested();
+
+    if (!_lf_async_event) {
+        return 0;
+    } else {
+        LF_PRINT_DEBUG(" *The STM32 rises from sleep* \n");
```
elgeeko1, line 158: Are there timing implications of `LF_PRINT_DEBUG` that should be considered here? Is this message necessary?

elgeeko1 (changes requested):
I added comments inline, and my general comments follow. At a high level, the most difficult aspects of this implementation are understanding concurrent access to volatile variables and enabling/disabling interrupts. Without more documentation, it's hard to know if these mechanisms are safe.

1. **Reference counting for enabling / disabling interrupts**: It's not clear how reference counting is managed, and whether or not it is handled in a way that is safe. Can you add documentation that indicates when this value changes and what could lead to an error?
2. `_lf_single_threaded_notify_of_event` is volatile, but it appears it's accessed only by a single-threaded LF application. Is this variable accessed by an ISR?
3. **Unclear ownership of `_lf_num_nested_crit_sec`**: It's unclear when this value is incremented and decremented. Consequently, it's possible that race conditions exist that can result in data corruption (two processes write to the value at the same time). Can you comment more about the use of this variable, who accesses it when, and how to guarantee that it is accessed in a predictable way?
4. Consider the names `lf_critical_section_enter` and `lf_critical_section_exit` instead of `lf_disable_interupts_nested` and `lf_enable_interrupts_nested`. This conveys that purpose of the method more clearly.
5. **Inconsistent return values**. Zero is consistently success, in some cases 1 is an error, others -1 is an error. Consider standardizing on 1 as success (as it evaluates to true in boolean expressions), and zero as failure.
6. Remove references to "I" or the thought process that lead you to the code. Instead, document what happens and why it is implemented this way. This is the most concise way for another developer to understand the code.

I hope this is helpful towards hardening your application. I will be happy to review any new updates.

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
...
+    uint32_t _lf_time_us_low = TIM5->CNT;
+
+    // Combine upper and lower timers (Yoinked from lf_nrf52 support)
+    uint64_t now_us = COMBINE_HI_LO((_lf_time_us_high - 1), _lf_time_us_low);
+    *t = ((instant_t)now_us) * 1000;
+    return 0;
+}
+
+/**
+ * Make the STM32 go honk shoo mimimi for set nanoseconds
```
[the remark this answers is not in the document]
naichenzhao: I am using blocking cause the HAL_Delay() function is set to only use milliseconds. I dont really know of another way to stall other than doing it manually

include/core/platform/lf_STM32f4_support.h
naichenzhao: I kinda just copied this over when I compared it against the rp2040 platform - I wasnt sure what this was used for so I kept it just incase. If its not used in LF proper, ill just delete it

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
...
+    if (HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_0) != HAL_OK) {
+        Error_Handler();
+    }
+}
+
+void Error_Handler(void) {
+    /* USER CODE BEGIN Error_Handler_Debug */
+    /* User can add his own implementation to report the HAL error return state */
+    __disable_irq();
+    while (1) {
```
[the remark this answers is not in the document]
naichenzhao: This is what the default STM32 library so I just didnt change it.

core/platform/lf_STM32f4_support.c
```diff
@@ -0,0 +1,251 @@
...
+int _lf_interruptable_sleep_until_locked(environment_t *env, instant_t wakeup_time) {
+    // Get the current time and sleep time
+    instant_t now;
+    _lf_clock_now(&now);
+    interval_t duration = wakeup_time - now;
+
+    // Edge case handling for super small duration
+    if (duration <= 0) {
+        return 0;
+    } else if (duration < 10) {
```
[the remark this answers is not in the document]
naichenzhao: this is kind of justa. constant value I tested and seems to work, though I will make it a defined value

core/platform/lf_STM32f4_support.c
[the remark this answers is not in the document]
elgeeko1: You can consider using HAL_Delay() for the milliseconds portion of the delay, and then cycles for the rest, if you need that level of precision.

lhstrh:
@naichenzhao, were you still planning on resolving the merge conflicts and bringing this into a state that we can merge it?

lhstrh:
Note that the only test failing is due to code formatting issues. Please run `make format` to fix.

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Write the time since boot into time variable
+ */
+int _lf_clock_gettime(instant_t* t) {
+  // Timer is cooked
+  if (!t) {
+    return 1;
+  }
+  // Get the current microseconds from TIM5
+  uint32_t _lf_time_us_low = TIM5->CNT;
```
erlingrj, line 76: There is a tricky race condition here that appears during the clock cycles where a counter overflow happens. Imagine the following scenario:
You read TIM5->CNT when the value is 0xffffffff (i.e. UINT32_MAX). Then there is an overflow and `_lf_time_us_high` is incremented before you read it a few lines below here. Then you end up with the wrong value.

Two ways to solve it, either duplicate read-outs (see NRF52 code), or disable interrupts (so that the overflow IRQ is not handled)

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+static void lf_busy_wait_until(instant_t wakeup_time) {
+  instant_t current_time;
+  _lf_clock_gettime(&current_time);
+
+  // We repurpose the lf_sleep function here, just to better streamline the code
+  interval_t sleep_duration = wakeup_time - current_time;
+  lf_sleep(sleep_duration);
+}
+
+// I am pretty sure this function doesnt work
```
erlingrj, line 118: Is this outdated? This function is critical and has to work

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+    _lf_clock_gettime(&current_time);
+
+  return 0;
+}
+
+/**
+ * Make the STM32 sleep for set nanoseconds
+ * Based on the lf_nrf52 support implementation
+ */
+static void lf_busy_wait_until(instant_t wakeup_time) {
```
erlingrj, line 109: This function is not needed

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  // Enable interrupts and prepare to wait
+  _lf_async_event = false;
+  lf_enable_interrupts_nested();
+
+  do {
+    _lf_clock_gettime(&now);
+    // Exit when the timer is up or there is an exception
+  } while (!_lf_async_event && (now < wakeup_time));
```
erlingrj, line 142: This should be low-power sleep, not busy-waiting. Is there anything in the HAL for doing low-power sleep?

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+  } else {
+    return 1;
+  }
+}
+
+//  + -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- +
+//  | Code for enabling and disabling Interrupts
+//  + -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- +
+
+// disables the IRQ (checks if its already disabled)
```
erlingrj, line 158:
```suggestion
// disables the IRQ with support for nested disabling
```

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+  if (_lf_num_nested_crit_sec == 0) {
+    __disable_irq();
+  }
+
+  // update the depth of disabling interrupts
+  _lf_num_nested_crit_sec++;
+  return 0;
+}
+
+// enables the IRQ (checks if other programs still want it disabled first)
```
erlingrj, line 170: The use of the word "program" is not correct here

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+  // Left the critical section more often than it was entered.
+
+  if (_lf_num_nested_crit_sec <= 0) {
+    return 1;
+  }
+
+  // update the depth of disabling interrupts
+  _lf_num_nested_crit_sec--;
+
+  // If we have exited all important programs, we can enable them again
```
erlingrj, line 181: "Programs" is not the correct term here

erlingrj (changes requested):
Hi @naichenzhao, this looks promising. I have a few requests for changes. I think the most important is to support low-power sleep_until.

Do you have any docs for how to install SDK and the right compiler so I can test it out at my end?

low_level_platform/impl/src/lf_STM32f4_support.c
```diff
@@ -0,0 +1,238 @@
...
+  // set prescaler to (16 - 1) = 15
+  // CPU runs a 16MHz so timer ticks at 1MHz
+  //      Which means period of 1 microsecond
+  TIM5->PSC = 15;
+
+  // Setup ISR to increment upper bits
+  TIM5->DIER |= TIM_DIER_CC1IE;
+  NVIC_EnableIRQ(TIM5_IRQn);
+
+  /* This is to make the Prescaler actually work
```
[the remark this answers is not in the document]
naichenzhao: I tried playing around with the order but it seems like it still didnt work. So unless I did something wrong with my implementation, I dont know if there is a way to avoid this

low_level_platform/impl/src/lf_STM32f4_support.c
naichenzhao: There seems to be a low power sleep mode, but its only interrupt based. It seems most timed delay implementations just use a while loop. Im not sure it itll be worth it trying to create some workaround now? or just merge now and maybe improve the implementation later.

PR #304, opened 2023-11-09
[Python 3.11] Fix Segfault

jackyk02:
**Issue**
Segfault due to Python 3.11 changes to object layout and memory allocation methods.

**Background**
Python 3.11 introduced changes to optimize CPython. One of the significant changes, as mentioned in [this PR](https://github.com/python/cpython/pull/29879), modifies the object layout to place pointers to dict and values immediately before the GC header. This change potentially affected the memory layout of custom python objects.

Before:
Untitled1
After:
Untitled2
Moreover, this commit removed ```_PyObject_GC_Calloc``` and ```_PyObject_GC_Malloc```, affecting the memory allocation for objects that participate in garbage collection.

**Solution**
 The new memory allocation relies on ```PyObject_New``` instead of ```PyObject_GC_New```.

**Updates in this PR**
The earlier PR https://github.com/lf-lang/reactor-c/pull/293 only updates the memory allocation for `convert_C_port_to_py`. However, updates are also necessary for `convert_C_mode_to_py` and `convert_C_tag_to_py`. 

**Test**
`./gradlew targetTest -Ptarget=Python` is expected to pass now with the modifications introduced in this PR.

### DIFF
--- a/python/lib/modal_models/impl.c
+++ b/python/lib/modal_models/impl.c
@@ -117,7 +117,7 @@ PyObject* convert_C_mode_to_py(
 ) {
     // Create the mode struct in Python
 	mode_capsule_struct_t* cap =
-        (mode_capsule_struct_t*)PyObject_GC_New(mode_capsule_struct_t, &mode_capsule_t);
+        (mode_capsule_struct_t*)PyObject_New(mode_capsule_struct_t, &mode_capsule_t);
     if (cap == NULL) {
         lf_print_error_and_exit("Failed to convert mode.");
     }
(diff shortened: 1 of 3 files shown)


lhstrh:
Haven't seen any activity on this since last fall. Can I go ahead and close the PR? Please weigh in on this, @jackyk02 and @edwardalee.

edwardalee:
> Haven't seen any activity on this since last fall. Can I go ahead and close the PR? Please weigh in on this, @jackyk02 and @edwardalee.

This is probably a necessary part of supporting Python 3.11, but I have no idea what else would be required and we would probably have to either require Python 3.11 or find a way to support both 3.10 and 3.11.  I suspect eventually we will need to support later versions of Python, so I don't know what to do with this either.

PR #306, opened 2023-11-11
Priority queue refactoring

edwardalee:
This is a refactoring and cleanup of the pqueue infrastructure with the addition of a priority queue that uses tag ordering.
Currently, this priority queue (a `pqueue_tag_t`) is not used anywhere except in unit tests.
The first place where it should be used is to replace the queue used by the `in_transit_message_tags` field of the `federate_info_t` struct maintained by the RTI for each federate.  Also, @byeong-gil 's NDT optimization requires such a queue.

Eventually, I believe this tag sorted queue could replace the event queue and significantly simply it without a performance penalty.


erlingrj:
Will this address this issue of RTI sending mistaken PTAGs: https://github.com/lf-lang/lingua-franca/issues/2084

lhstrh:
> Will this address this issue of RTI sending mistaken PTAGs: [lf-lang/lingua-franca#2084](https://github.com/lf-lang/lingua-franca/issues/2084)

I don't think it does -- this is just a refactoring AFAIK.

byeong-gil:
Thank you for refactoring this, @edwardalee! This is really helpful for the NDT optimization. 

In this commit ([fb024fa](https://github.com/lf-lang/reactor-c/pull/306/commits/fb024fa6141237c58d765a964bf84e84c0b0f367)), I made the function [`pqueue_tag_insert_tag`](https://github.com/lf-lang/reactor-c/blob/fb024fa6141237c58d765a964bf84e84c0b0f367/core/utils/pqueue_tag.c#L150-L160) doesn't insert an existing tag again to save the memory. I modified the unit test to test it too. I think this is ok because the function is only used when elements don't contain any payloads. If a user wants to use a customized element that contains a payload as described [in this comment](https://github.com/lf-lang/reactor-c/blob/fb024fa6141237c58d765a964bf84e84c0b0f367/include/core/utils/pqueue_tag.h#L33-L47), they can use the [`pqueue_tag_insert`](https://github.com/lf-lang/reactor-c/blob/fb024fa6141237c58d765a964bf84e84c0b0f367/core/utils/pqueue_tag.c#L146-L148) function.

edwardalee (commented):
I will try to look at this more closely later, but some quick feedback: You are right that we will need the capability to prevent accumulation in the queue of elements with the same tag. This will be particularly important if tags corresponding to the timeout time or FOREVER are posted, but we could get an unbounded accumulation of these without them ever getting cleared.

However, the solution looks expensive to me. Can we avoid mallocing and then freeing an element that is used just to conduct the search? I think the element can be put on the stack.

And perhaps the de-duplication should not be done by default. How about an alternate insert function (insert if not present)?  Also, I suspect that instead of searching the queue, a low-level algorithm would be much more efficient. Probably it would just be a small variant of the bubble_up function in pqeueue_base that would perform the de-duplication while doing the insertion, instead of doing it in two separate steps.

(Actually, on reflection, probably this low-level idea won't be so easy... The bubble_up function in pqueue_base is destructive, in that it changes the data structure as it goes. If it finds a duplicate after making all these changes, then it likely will have to undo all those changes.  Probably the two-phase approach is actually better.)

byeong-gil:
Thank you for the feedback, @edwardalee! I modified the code based on your comment.

> Can we avoid mallocing and then freeing an element that is used just to conduct the search?

Now, the function [`find_same_tag`](https://github.com/lf-lang/reactor-c/blob/22423a350aa93d5a51997b191b48a2dbe905621d/core/utils/pqueue_tag.c#L75-L109) only require a tag and mallocing is not needed anymore. One disadvantage is we need another function when we want to compare a payload as well as the tag.  That's why I named the function `find_same_tag` instead of `find_equal_same_tag`.

> How about an alternate insert function (insert if not present)?

I made the function here ([`pqueue_tag_insert_tag_if_not_present`](https://github.com/lf-lang/reactor-c/blob/22423a350aa93d5a51997b191b48a2dbe905621d/core/utils/pqueue_tag.c#L159-L165)).

edwardalee:
> Thank you for the feedback, @edwardalee! I modified the code based on your comment.
> 
> > Can we avoid mallocing and then freeing an element that is used just to conduct the search?
> 
> Now, the function [`find_same_tag`](https://github.com/lf-lang/reactor-c/blob/22423a350aa93d5a51997b191b48a2dbe905621d/core/utils/pqueue_tag.c#L75-L109) only require a tag and mallocing is not needed anymore. One disadvantage is we need another function when we want to compare a payload as well as the tag. That's why I named the function `find_same_tag` instead of `find_equal_same_tag`.
> 
> > How about an alternate insert function (insert if not present)?
> 
> I made the function here ([`pqueue_tag_insert_tag_if_not_present`](https://github.com/lf-lang/reactor-c/blob/22423a350aa93d5a51997b191b48a2dbe905621d/core/utils/pqueue_tag.c#L159-L165)).

I found a way to do this without replicating code from pqueue_base and without using malloc (the elements are on the stack).

lhstrh (approved):
Looks great! Let's merge this.

PR #308, opened 2023-11-15
Preliminary support for enclaves in the C target

erlingrj:
This PR corresponds to this LF PR: https://github.com/lf-lang/lingua-franca/pull/2104

### DIFF
--- a/core/environment.c
+++ b/core/environment.c
@@ -114,15 +114,16 @@ static void environment_init_federated(environment_t* env, int num_is_present_fi
 }
 
 void environment_init_tags( environment_t *env, instant_t start_time, interval_t duration) {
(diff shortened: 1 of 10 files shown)


edwardalee (approved):
This PR looks ready to merge. Together with the companion PR in lingua-franca, it takes enclaves to the next level. There are still many limitations, but it's a step forward.

edwardalee:
There are still a number of limitations, but this is a good starting point for enclave support in the C target. Limitations:
- Enclaves not supported in federated programs
- Broadcast connections not currently supported
- Banks of enclaves are not supported
- Enclaves with multiports are not supported
- Enclaves with array port types are not supported
- Enclaves only supported with singular connections
- Enclaves cannot be involved in interleaved connections

PR #310, opened 2023-11-23
Add clang-format and run it

erlingrj:
In this PR I copy over the formatter setup from reactor-cpp. I think we should use a formatter, we can argue about what the settings of it should be. Currently I put it at a column width of 80. Reactor-cpp uses 120.

The bigger question is when we should apply it. It will create problems for any outstanding feature-branch...


erlingrj:
It is not urgent to make a decision, I think ideally we make a plan for when to apply it. Some point in the future when we have merged most of the big outstanding work.

edwardalee:
There are several outstanding PRs, so this does not seem like a good time to do this. But it is something we should do.

cmnrd:
There is never a good time for something like this. If we want to do it (which I think we should), then we have to bite the bullet (like we did in the main LF repo).

edwardalee:
Hmm... doing this now will likely result in my ongoing cleanups getting abandoned... can we coordinate and it between major phases of this?  I'm in the middle of one that has hundreds, maybe thousands of line changed.  @erlingrj also has some large pending work.

cmnrd:
Sure, I just noticed this PR today and wanted to ping it and point out that there is never a perfect time. But if there is a concrete milestone to wait for, that is fine.

edwardalee:
How about instead providing the docs on how to set VS Code to do this formatting (in the developer section of the handbook), and also the instructions for how to do the mass formatting, and I will coordinate with @erling to find a good time to do this.

cmnrd:
We currently don't have developer instructions for the individual runtimes in the handbook. And I think we should keep those instructions in the README of the runtimes. Independent of where we place it, here is a simple description:

> We use [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html) for automatically formatting the source code. Please run `clang-format <file-name>` before you commit a change. To format all files in the repository, you can use the following command:
>```
>clang-format $(find -name "*.h" -o -name "*.c") 
>```
>Most IDEs provide support for formatting with `clang-format`. VS Code, for instance, has [built-in support](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting) and automatically invokes `clang-format` when you use the IDE's formatting functionality.

PR #312, opened 2023-11-24
draft: debugging PTAG and ABS messages

byeong-gil:
This PR aims to debug the PTAG and ABS messages. Currently, in this PR ([reactor-c/enclaves3](https://github.com/lf-lang/reactor-c/pull/311)), PTAG isn't sent when a federate is not in a zero-delay cycle (ZDC). However, the semantics of PTAG should be correct even for federates that aren't in a ZDC and there is a problem with PTAG now. So I made this PR to preserve the state that does not skip PTAGs and debug them.


byeong-gil:
@edwardalee @hokeun @lhstrh I tried to resolve the problem on `AfterDelays.lf` and encountered additional difficulties. This is the summary of the efforts and I will explain this in the next meeting. 

The reason why a federate sends a wrong LTC (described in [this issue](https://github.com/lf-lang/lingua-franca/issues/2084)): Federates only consider zero-delay actions when advancing MLAA (Max Level Allowed to Advance). 

So I let federates look up every network action in this commit ([reactor-c/pull/312/commits/9307777](https://github.com/lf-lang/reactor-c/pull/312/commits/9307773fe5c2ceadd0a8f801bc7c816e6a869a39)). However, this commit released other masked problems.

* Problem on [`AfterDelays.lf`](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/AfterDelays.lf)

  There is a non-deterministic error in `AfterDelays.lf`. Let's compare two cases. I attached images below. We only need to see NET (100 msec, 0) from `sw2` (The second from the right one). In the success trace, NET (100 msec, 0) is sent and in the failure trace, it is not sent. This is caused by 'the timing of PTAG' and the code below. 
https://github.com/lf-lang/reactor-c/blob/16af56a19c1c6eeb3556d4f3c14a7c3742c67409/core/federated/federate.c#L2719-L2723
In the failed trace, a federate received PTAG (82 msec) before it completed the tag (81 msec). So it tried to send NET (82 msec) but it didn't send it because last_TAG = 82 msec >= tag = 82 msec. However, NET (82 msec) must be sent because it's stalled by MLAA and still waiting for TAG (82 msec). 

  Successfully executed trace
image

  Deadlock occurred trace
image

* Solution of the problem on `AfterDelays.lf`
  Thus, I changed the code above like this. https://github.com/lf-lang/reactor-c/blob/51da45b530b3c11296a54b46972a2d584e62bb19/core/federated/federate.c#L2721-L2727 
  This code will not send NET when 
  1. TAG >= intended NET 
  2. PTAG > intended NET 
  3. PTAG > current tag. 

  You can see that a federate sends NET (62 msec) although it received PTAG (62 msec) before in the trace below. And that NET (62 msec) allowed the RTI to send TAGs. So the deadlock didn't happen. I guess this is not an optimum solution yet it works. Because duplicated NETs are being sent. I'm trying to devise an elegant solution.

  The trace after fixing the problem
image

* Problem on [`ChainWithDelay.lf`](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/ChainWithDelay.lf)
  A problem occurred after I made this commit ([reactor-c/pull/312/commits/9307777](https://github.com/lf-lang/reactor-c/pull/312/commits/9307773fe5c2ceadd0a8f801bc7c816e6a869a39)) to consider every action in a federate. 
  image
  Let's assume that we're at the tag (33 msec). The RTI sent PTAG (33 msec) to everyone. So, `p: PhysicalPlant` sends T_MSG with the tag (33 msec). However, `c:Controller` cannot execute `reaction_2`. Because it has `reaction_1`. We know that there is no message with the tag (33 msec) for `reaction_1` as `pl:Planner`'s NET is 33 msec and there is a delay at the connection. But `c:Controller` has no sense about it and waiting for ABS, T_MSG, or TAG at 33 msec. 
  Please note that we only consider zero-delay actions when advancing MLAA, this wasn't a problem because we didn't wait for `reaction_1`.

* Solution of the problem on `FeedbackDelay.lf`

  I'm still trying to take the shape of the solution.

  The bottom line is that `c:Controller` cannot know the status of `reaction_1`. The RTI knows that information but it doesn't have a way to inform it. `pl:Planner` also knows that information and sending ABS (33 msec) will be the simplest solution. However, `pl:Planner` cannot recognize that `c:Controller` has an event at `33 msec` and it is waiting for its input for `33 msec`. Maybe `NDT` can help to solve this problem?

PR #316, opened 2023-12-02
Handle messages arriving during initial STA wait

edwardalee:
Some errors have appeared in master that are not caught.  For example, the federated STPParameter.lf test case prints the following error message:
```
Federate 1: Timer ticked at (0, 0).
Federate 0: Environment 0: ---- Spawning 2 workers.
Federate 1: ERROR: STP violation occurred in a trigger to reaction 1, and there is no handler.
**** Invoking reaction at the wrong tag!
Federate 1: Received 1.
```
In fact, the "Received 1" message is supposed to occur before the "Timer ticked" message.

I assume that this has been changed to a non-fatal error to prevent flaky test failures, but in this case, it is masking a real test failure.

This PR pulls events off the event queue before advancing to the start time so that messages received during the initial STA wait are properly handled.

However, it turns out there are more problems.  The federated test DecentralizedP2PComm succeeds but prints out the following error messages:

```
Federate 0: Received 100.
Federate 1: Received 0 late.
Federate 1: ERROR: STP offset was violated by (0, 1).
...
```

So there are more problems...

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -691,7 +691,7 @@ void _lf_initialize_start_tag(environment_t *env) {
     // Add reactions invoked at tag (0,0) (including startup reactions) to the reaction queue
     _lf_trigger_startup_reactions(env);
 
-#ifdef FEDERATED
+#if defined FEDERATED
     // If env is the environment for the top-level enclave, then initialize the federate.
     environment_t *top_level_env;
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
Looks OK, but as a general comment: I have to admit that all the of the ifdefs throughout this code I find it pretty difficult to understand what's going on...

PR #317, opened 2023-12-02
Remove absent messages

edwardalee:
This is a companion to https://github.com/lf-lang/lingua-franca/pull/2118. The main purpose of this PR is to remove absent messages from most federated programs, restricting their use to zero-delay cycles.


byeong-gil:
I tried to define and formularize what strict EIMT (`SEIMT`) should be and this is the result. I used the term `SEIMT` for simplification. 

The RTI should send `PTAG_i` to federate `i` only when 
i. `EIMT_i` is equal to the `G(NET_i)` from federate `i`.
ii. federate `i` is in a zero-delay cycle (ZDC).
iii. it can guarantee that there will be no messages that G(MSG_ij) <= G(PTAG_i) from upstream of `i` that are **not in ZDC** (`j` is a member of "immediate upstream federates that are not in a ZDC").

`SEIMT` is needed for checking the third condition. Now, let's examine cases of connections between federates.

1. A source federate is in a zero-delay cycle (ZDC).
In the first case, the network message action of this connection in federate `i` will wait for a tagged message or a port absent message and everything will be ok. Please note that a source federate sends ABS messages when it is in a ZDC.
2. A source federate is not in a ZDC.
`SEIMT` is important in this case. As a result of `zero_delay_cycle_action_table`, a federate does not wait for messages from its upstreams that are not in a ZDC. It just assumes the port is absent when no message has arrived. 
Thus, `SEIMT_i` should be calculated by **`min(EIMT_j + Delay_j)`** where `Delay_j` is a minimum delay between the federate `j` and the target federate `i`. When `SEIMT_i` > `NET_i`, the RTI knows that messages that `G(MSG_ij)` <= `G(NET)` will never be sent and is able to grant `PTAG_i` that `G(PTAG_i)` = `G(NET_i)`.

With this declaration, `SEIMT` is calculated recursively as `EIMT` calculation uses recursion. However, I'm worried about the complexity of this calculation. Maybe we can save the result of [`update_min_delays_upstream`](https://github.com/lf-lang/reactor-c/blob/61261af0b93571226e1aa91ed314d1e3b13ce0b2/core/federated/RTI/rti_common.c#L301) somewhere and recalculate it only when the structure of the federation is changed?

@edwardalee @hokeun @lhstrh I would appreciate it if you read this and give me any feedback.

edwardalee:
Yes, this looks right.  update_min_delays_upstream already caches the results.

lhstrh:
Subsumed by #323.

Issue #215, opened 2023-05-20
Action drop policy not correcly implemented


erlingrj:
Currently the drop policy with a minimum spacing only works if the prior event is still at the event queue. If the prior event is already handled then the drop policy will not enforce the minimum spacing.

Issue #145, opened 2023-01-20
trigger_t objects point to "last" event, which may have been freed or recycled.


edwardalee:
The trigger_t struct has a field `last` that points to the last event scheduled by that trigger. That field is used to implement the drop policy to enforce minimum spacing.  Unfortunately, the field may point to an event that has been recycled (or worse, freed).

The field needs to be replaced with a `last_time` field initialized to `NEVER`. However, this also means that the pqueue function `pqueue_find_equal_same_priority` needs to be redone to take a trigger and time argument instead of an event.

lhstrh:
Related to #215.

SheaFixstars:
As mentioned in this issue this is related to the `last` trigger field. 

When that event is done it gets recycled but the trigger->last is not removed so when the next event goes to run it could trigger the drop policy due to the `last` event being reused, assuming it does not crash due to freed memory. This leads to lost events as high as 50% lost in testing that I see. This issue is also hit with both "replace" and "defer" policy as well. 

- If there is a min_space over 0 then the issue is triggered, regardless of the policy.
- Only triggers when there is no entry in the queue when a new entry comes in

The test below is using a logical action with a drop policy, the queued items will only ever have one item in the queue at a time so they don't overlap and properly trigger the drop policy.

The following program is setup as follows:
 - Timer runs every 500ms
 - Schedules logical action with 1ms delay, 2000ms min_space "drop" policy
   - Prints out the Action statement
   - Current is current logical time, scheduled is the expected logical runtime
 - Logical action prints out the run statement when run
   - Time and microstep are printed out, the Msg Logical is when it was expected to run

The issue can be seen in the following:
```
(Timer) Schedule - Current: 0, Scheduled: 1000000, Count:0
(Action) Run - Time:1000000 Microstep:0, Msg Logical:1000000, Count:0
(Timer) Schedule - Current: 500000000, Scheduled: 501000000, Count:1
(Timer) Schedule - Current: 1000000000, Scheduled: 1001000000, Count:2
(Action) Run - Time:1001000000 Microstep:0, Msg Logical:1001000000, Count:2
(Timer) Schedule - Current: 1500000000, Scheduled: 1501000000, Count:3
(Timer) Schedule - Current: 2000000000, Scheduled: 2001000000, Count:4
(Action) Run - Time:2001000000 Microstep:0, Msg Logical:2001000000, Count:4
(Timer) Schedule - Current: 2500000000, Scheduled: 2501000000, Count:5
[...]

Issue #135, opened 2023-01-06
Return value of lf_critical_section_enter() and exit() are inconsistent


edwardalee:
In platform.h, these functions are defined:
```
extern int lf_critical_section_enter();
extern int lf_critical_section_exit();
```
However, reactor.c discards the return value:
```
void _lf_critical_section_enter() {
    lf_critical_section_enter();
}
void _lf_critical_section_exit() {
    lf_critical_section_exit();
}
```
Shouldn't _lf_critical_section_enter and exit return an int consistently?  And shouldn't the return value be checked?

erlingrj:
I agree, I will address this.

PR #318, opened 2023-12-08
Fix minimum spacing of actions and reference to freed event

edwardalee:
This PR addresses #145 and #215, fixing the `min_spacing` option on actions and removing the reference to a possibly freed or recycled event.  It has a companion PR in lingua-franca. Thanks @SheaFixstars for prompting this.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -1035,6 +1035,8 @@ trigger_handle_t _lf_schedule(environment_t *env, trigger_t* trigger, interval_t
                         // Leave the last_time the same.
                         return(0);
                     }
+                    _lf_recycle_event(env, dummy);
+
                     // If the preceding event _has_ been handled, then adjust
                     // the tag to defer the event.
                     intended_time = earliest_time;
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -853,10 +853,10 @@ int _lf_schedule_at_tag(environment_t* env, trigger_t* trigger, tag_t tag, lf_to
  * the will be freed. Hence, it is essential that the payload be in
  * memory allocated using malloc.
  *
- * There are three conditions under which this function will not
+ * There are several conditions under which this function will not
  * actually put an event on the event queue and decrement the reference count
  * of the token (if there is one), which could result in the payload being
- * freed. In all three cases, this function returns 0. Otherwise,
+ * freed. In all cases, this function returns 0. Otherwise,
  * it returns a handle to the scheduled trigger, which is an integer
  * greater than 0.
  *
@@ -864,7 +864,9 @@ int _lf_schedule_at_tag(environment_t* env, trigger_t* trigger, tag_t tag, lf_to
  * offset plus the extra delay is greater than zero.
  * The second condition is that the trigger offset plus the extra delay
  * is greater that the requested stop time (timeout).
- * The third condition is that the trigger argument is null.
+ * A third condition is that the trigger argument is null.
+ * Also, an event might not be scheduled if the trigger is an action
+ * with a `min_spacing` parameter.  See the documentation.
  *
  * @param env Environment in which we are executing.
  * @param trigger The trigger to be invoked at a later logical time.
@@ -964,9 +966,8 @@ trigger_handle_t _lf_schedule(environment_t *env, trigger_t* trigger, interval_t
     e->intended_tag = trigger->intended_tag;
 #endif
 
-    event_t* existing = (event_t*)(trigger->last);
     // Check for conflicts (a queued event with the same trigger and time).
-    if (trigger->period < 0) {
+    if (min_spacing <= 0) {
         // No minimum spacing defined.
         tag_t intended_tag = (tag_t) {.time = intended_time, .microstep = 0u};
         e->time = intended_tag.time;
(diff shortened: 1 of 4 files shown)


core/reactor_common.c
```diff
@@ -988,75 +989,59 @@ trigger_handle_t _lf_schedule(environment_t *env, trigger_t* trigger, interval_t
...
-                        // If the last event hasn't been handled yet, insert
-                        // the new event right behind.
-                        existing->next = e;
-                        return 0; // FIXME: return a value
-                    } else {
-                         // Adjust the tag.
-                        intended_time = earliest_time;
-                    }
+                    // Default policy is defer
+                    intended_time = earliest_time;
```
lhstrh, line 1044: We had decided that "replace" should become the default (see https://github.com/lf-lang/lingua-franca/issues/1464#issuecomment-1797990069). I think it would make that change as part of PR and I can help with that.

lhstrh:
Not sure this is related to https://github.com/lf-lang/reactor-c/issues/135. Is it?

core/reactor_common.c
edwardalee: Actually, "replace" has a very weird feature, one that convinces me it should **not** be the default.  In particular, I can ask for a value to be scheduled at, say, (1000, 0), and because of the "replace" policy, the value will appear at (500, 0). Time travel!

In any case, that change, if we make it, should be a separate PR and should be coupled with the other changes in [1464](https://github.com/lf-lang/lingua-franca/issues/1464#issuecomment-1797990069), I think.  I don't think we should slow down this PR for it.

lhstrh: The reason I'm making this connection is that we have several very long threads discussing weird corner cases and making proposals for changes. The problem with these threads is that they tend to linger, and, over time, related issues are then discussed independently, completely ignoring progress that previously had been made...

I'm not proposing a hold up, and wanted to get right to it, but your comment just now gives me pause... Have you gone over the thread that I linked to?

lhstrh: > Actually, "replace" has a very weird feature, one that convinces me it should **not** be the default. In particular, I can ask for a value to be scheduled at, say, (1000, 0), and because of the "replace" policy, the value will appear at (500, 0). Time travel!

I'm not sure that the behavior you're describing is implied by the term "replace." What if we don't only update the value, but also the tag (i.e., a wholesale replacement of the event)?

Also, it seems that most weird scenarios are tied with non-monotonic event sequences to begin with, which seem atypical.

edwardalee: The "replace" in [1464](https://github.com/lf-lang/lingua-franca/issues/1464#issuecomment-1797990069) is not referring the policy to use to handle `min_spacing`.  That issue is talking about something else altogether, which is what to do when two `lf_schedule` calls on the same action are requesting an event at the same future tag.  The proposal in [1464](https://github.com/lf-lang/lingua-franca/issues/1464#issuecomment-1797990069) is that the second event will replace the first.  This would be largely independent of the `min_spacing` property and its policy.

The `min_spacing` policy kicks in not when two calls to `lf_schedule` ask for the same tag, but when they as for two _times_ that are too close together.  The `replace` policy could do one of two things:

1. Replace the earlier event's value with the new value (what this PR does), or
2. Remove the earlier event and insert the new event at the later tag.

Both of these have undesirable features.  The first does time travel, the second discards an earlier event and therefore slows down responses.  In fact, the second one could result in a system _never_ responding because each event gets replaced with a later event, and before the later event can be replaced, it too gets replaced.  This is nastier than time travel, IMHO, so I opted for 1.

A better solution might be to eliminate the "replace" policy altogether.  That would be better than making it the default.

SheaFixstars: I think that there is merit in both of the replace policy actions that are mentioned. The developer may want one behavior over the other depending on the exact requirement and what tradeoff is acceptable.

As such would it be appropriate to suggest that the _current_ `replace` policy gets renamed to an `update` policy which may convey the actual action of what is happening a bit better. 
And a new `replace` policy is added that implements the 2. point of replacing the previous event in its entirety thus acting at the later time.

If the `replace` policy is eliminated then I believe it is necessary for the ability to control event queues, such as removing an event. To be able to remove stale data from a queue for newer data to be used as an example. 
But should that default policy change even be for all actions or an action with the same time but different tag? As my reading of the proposal looks like it morphed from scheduling actions at the same time and how to handle that to the default policy for all actions.

edwardalee: I agree that update would be better name for the replace policy.  But I'm not sure we want to even offer option 2.  The risk of having a never-ending chain of never processed events seems like it could create safety risks.

edwardalee: This PR seems to be stalled on what I consider to be a secondary issue that should be split off into a separate issue.  At most, this PR could do the rename of "replace" policy to "update" policy.  However, such a change is not backward compatible, requires updates in two repos, and requires updating the handbook web pages.  I don't have the bandwidth to make this change right now.  Should we merge this and open an issue for that change?

SheaFixstars: Apologies, I was not intending to suggest that the naming be dealt with currently as I am aware it is a breaking change. My biggest concern is the initial bug as it is affects some of my projects quite a bit.

edwardalee: OK, I'm going to mark this resolved and move forward.

core/reactor_common.c
```diff
@@ -988,75 +989,59 @@ trigger_handle_t _lf_schedule(environment_t *env, trigger_t* trigger, interval_t
             }
             // Hook the event into the list.
             found->next = e;
+            trigger->last_time = intended_tag.time;
```
lhstrh, line 992: I thought that we wanted to keep track of the _last time of scheduling_, not tag of time of the last-scheduled event, because those can be all over the place. In other words, shouldn't this be as follows?
```suggestion
            trigger->last_time = lf_time_logical_elapsed();
```

lhstrh (changes requested):
Looks great, modulo suggested change.

SheaFixstars (changes requested):
Thanks for the fix with this issue.

core/reactor_common.c
edwardalee: No.  The time at which lf_schedule is called is irrelevant.

edwardalee: Note further that this will be used typically in physical actions, and logical time is not well defined outside the context of a reaction.

lhstrh: The times at which `schedule` is invoked will be monotonically increasing. We have no assurance of that with the intended times, which could really undermine the mechanism. Generally, the problem I see is that you can schedule clusters of events that are very densely packed (i.e., with very little spacing), as long as the clusters have sufficient spacing between them and you schedule the events in an interleaving/round robin manner.

The same is true if we were to go by the times of scheduling (be it logical or physical) instead of the times of the events scheduled, however. The only way to actually achieve spacing on the logical time line is to scan the event queue and look for conflicts. As you pointed out, this is an expensive operation, however, and gives rise to tricky corner cases that need to be handled...

Exactly, what guarantees can we derive from the spacing as it is implemented now? It seems to me that the ability to specify an additional delay is what creates chaos and spoils our ability to provide meaningful guarantees. Perhaps we should not allow additional delay for actions that have a minimum spacing. We could have a keyword (`dynamic`) for actions that allow an additional delay and use the target language's type system to catch situations where the action was not declared accordingly. That way, we have strong guarantees for ordinary (e.g., `static`) actions and weaker guarantees for dynamic ones.

lhstrh (approved):

PR #324, opened 2023-12-27
Fix LF_ASSERT

erlingrj:
Thanks @edwardalee  for spotting this mistake. LF_ASSERT was not expanded correctly when we used build-type: Release.


edwardalee (approved):
This change is necessary to ensure correctness, and the test failures are just the usual macOS federated failures that all branches are getting. However, I think a better solution is needed for the long run with more customized macros (e.g. for checking memory allocation). This change will likely result in compiler warnings about the result of expressions not being used when using build-type: Release.

erlingrj:
I am not getting compiler warnings when compiling with `build-type: Release`. We can make more elaborate macros also. But I think the LF_ASSERT macro still has its use.

edwardalee:
Unfortunately, with this change, I get hundreds of warnings as follows:
```
/Users/edwardlee/LinguaFranca/test/C/src-gen/modal_models/ConvertCaseTest/core/environment.c:209:20: warning: expression result unused [-Wunused-value]
    LF_ASSERT(env->shutdown_reactions, "Out of memory");
```
We need a different solution.

erlingrj:
Unfortunate that I am not getting them. Will see if I can turn those warnings on. Probably solvable with a void cast. Give me a sec

PR #325, opened 2023-12-28
Bugfix to avoid deadlock on STP violation

edwardalee:
When an STP violation occurs with decentralized coordination, the `last_known_status_tag` of the network input port was being left at the intended tag, but should have been updated to the tag at which the input will be handled.  This can cause a deadlock with decentralized coordination and may be causing at least some of the flaky test failures in CI.  The bug is only triggered when an STP violation occurs, and this is more likely in CI.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -382,6 +382,9 @@ void _lf_pop_events(environment_t *env) {
                                     event->trigger,
                                     event->intended_tag.time - start_time, event->intended_tag.microstep,
                                     env->current_tag.time - start_time, env->current_tag.microstep);
+                        // Need to update the last_known_status_tag of the port because otherwise,
+                        // the MLAA could get stuck, causing the program to lock up.
+                        event->trigger->last_known_status_tag = env->current_tag;
                     }
                 }
 #endif


lhstrh (approved):
This is a very subtle issue, but the fix makes sense to me!

lhstrh:
Should we just merge this? `main` is already failing...

petervdonovan:
OK, I won't stand in the way of that. I agree that this looks like a necessary fix for a very subtle bug, and if main is broken anyway then I have no problem with just merging this.

But I would like to know why the macOS tests were just canceled while running the federated tests. Is it because they ran for too long? If the federated tests on macOS all deadlocked and timed out, that could be a reason why the macOS tests overall might time out and get canceled. And in fact it looks like not a single run of the federated macOS tests has completed successfully, which is worrisome.

lhstrh:
The current state of affairs is that federated execution in macOS is broken. It's proven very difficult to fix it because the failure is a deadlock, and there is no output. Moreover, we cannot reproduce the issue on a local machine, yet, it shows up reliably in CI. We've been stuck on this for a number of weeks now. Indeed, it is worrisome, which is why we've pushed off the release, but we're still intent on fixing it. If there's anything you could do to help pinpoint the problem or fix it, that would be greatly appreciated.

petervdonovan:
Ah, I see. So this is a fix for a different problem. Yep, sounds good, let's merge this.

lhstrh:
AFAIK, @edwardalee has been a lot of code review and refactoring, motivated by the aforementioned problems but with a wider scope (in part, I presume, because there is not enough information for a more targeted approach).

lhstrh:
> But I would like to know why the macOS tests were just canceled while running the federated tests.

On this: we only have 5 macOS runners, so canceling checks is often necessary to free up runners for tasks that we do want to see the output of.

edwardalee:
Closing this because it is subsumed by #323.

PR #326, opened 2023-12-28
LF_ASSERT: Cast to void to avoid unused warnings

erlingrj:
@edwardalee can you test if this removes your warnings? I am not able to reproduce the warnings here with Linux, even with -Wall -Wpedantic and -Wextra


edwardalee (approved):
Yes, this eliminates the warnings. Thanks.

PR #330, opened 2024-01-15
Draft: Create Network driver instead of directly using sockets.

Jakio815:
lf-lang/lingua-franca#2256
# Open problems
[] OPENSSL_init_crypto(OPENSSL_INIT_NO_ATEXIT, NULL); - the "atexit" problem.

# Netdriver
Current LF code of the reactor-c uses TCP and UDP sockets directly, limiting to scale network options. This PR introduces a network driver to support various network types.
```
typedef struct netdrv_t {
  void* priv;
  unsigned int read_remaining_bytes;
  int federate_id;
} netdrv_t;
```
The `void* priv` will be casted depending on the network type.

# APIs.
The main design is as a client and server. The RTI is always a server. Federates can be servers and clients. As a client, it should be able to connect to the RTI, and also connect to other federates. As a server, it should be able to accept connections from other federates for two reasons: decentralized coordination and physical connections.
So, the main design follows TCP socket server-client design. 
The RTI will have a netdriver that accepts other federate connections, and a netdriver will be allocated for each federate.
The federate will have a netdriver to connect to the RTI as a client.
Next, a netdriver to accept other federate's connection as a server.
Finally netdrivers to communicate with each inbound and outbound p2p connections.

The description below separates the APIs as functions for clients, servers, and common each.
### netdrv_t* initialize_netdrv();
Initializes netdriver, and allocates memory.
### void close_netdrv(netdrv_t* drv);
Frees memory allocated by netdriver.
### int write_to_netdrv(netdrv_t* drv, size_t num_bytes, unsigned char* buffer);
This sends the buffer to the netdriver.
TCP: write()
MQTT: publish()
SST: encapsulate message with SST, and send.

The below two each has additional error handling.
`int write_to_netdrv_close_on_error(netdrv_t* drv, size_t num_bytes, unsigned char* buffer);`
`void write_to_netdrv_fail_on_error(netdrv_t* drv, size_t num_bytes, unsigned char* buffer, lf_mutex_t* mutex,char* format, ...);`
### ssize_t read_from_netdrv(netdrv_t* drv, unsigned char* buffer, size_t buffer_length);
`ssize_t read_from_netdrv_close_on_error(netdrv_t* drv, unsigned char* buffer, size_t buffer_length);`
`void read_from_netdrv_fail_on_error(netdrv_t* drv, unsigned char* buffer, size_t buffer_length, lf_mutex_t* mutex, char* format, ...);`

### int create_server(netdrv_t* drv, int server_type, uint16_t port);
TCP: creates server socket, binds, and listens to server socket.
MQTT: Connects to broker. Subscribe to Topic “{Federation_ID}_RTI”
SST: creates server socket, init_SST(): loads keys

### netdrv_t* establish_communication_session(netdrv_t* netdrv);
TCP: accept() for federate socket requests, return fed_socket.
Mqtt:
1. (RTI is already subscribing to “{Federation_ID}_RTI”)
2. (Each federate publishes fed_id to “{Federation_ID}_RTI”)
3. RTI subscribes to “{Federation_ID}_fed{fed_id}_to_RTI“
4. (Federate subscribes to ”{Federation_ID}_RTI_to_fed{fed_id}”)
* Now there are two one-way channels for each RTI-fed connection.

SST: accept() sockets. RTI does handshake with federates. SST_session_ctx is created for each federate.

### int create_client(netdrv_t* drv, int federate_id);
TCP: Creates a real-time TCP socket.
MQTT: Federate connects to broker.
SST: init_SST().

### int connect_to_netdrv(netdrv_t* drv);
TCP: connect()
MQTT: Federate publishes to “{Federation_ID}_RTI”, then subscribes to ”{Federation_ID}_RTI_to_fed{fed_id}”
SST: Requests session key from Auth, and request connection to RTI or federate server.

# Major changes
## MQTT

| | |**\_RTI**| **\_RTI\_to\_fedID** | **\_fedID\_to\_RTI** | **\_fedServID** | **\_fedServID\_to\_fedCliID** | **\_fedCliID\_to\_fedServID** |
|:----------:|:----------:|:----------:|:----------:|:----------:|:----------:|:----------:|:----------:|
|RTI           |rti\_netdrv|sub           |                |                |                |                |                |
|                |fed\_netdrv|              |pub          |sub          |                 |                |                | 
|Fed         |rti\_netdrv|pub          |sub           |pub         |                |                 |               |
|                |serv\_netdrv|             |                |                |sub           |pub          |sub          |
|               |cli\_netdrv |                |                |                |pub          |sub          |pub          |

## SST
Netdriver supports SST.
```
reactor C {
  comm-type:SST
}
```
RTI must be built with `-DCOMM_TYPE=SST`.
### Instructions
1. Install sst-c-api
OpenSSL 3.0 above must be installed.
```
$ git clone https://github.com/iotauth/sst-c-api.git
$ cd sst-c-api && mkdir build && cd build
$ cmake ../
$ make
$ sudo make install
```
2. Generate keys and execute `Auth`.
Java 11 or above must be installed.
```
$ git clone https://github.com/iotauth/iotauth.git
$ cd iotauth/examples
$ ./generateAll.sh -g configs/lf_sst.graph 
$ cd ../auth/auth-server/
$ java -jar target/auth-server-jar-with-dependencies.jar -p ../properties/exampleAuth101.properties
```
The path must be in the same directory with `lingua-franca` for the script to work. This is for key paths, and auto generated `.config` files.
```
.
├── iotauth
├── lingua-franca
```
3. Install RTI with `-DCOMM_TYPE=SST`.
4. Add `comm-type:SST` to `.lf` code.
### Input for configuration is required.
Each federate binaries and the RTI needs a config as an input with the `-sst` or `--sst` option.

The code generator copies the skeleton config from `reactor-c/core/federated/network/SSTskeleton.config`.
It writes three components.
- entityInfo.name
- authInfo.pubkey.path
- entityInfo.privkey.path

The generated federate's sst config will be saved such as `test/C/fed-gen/SimpleFederated/src-gen/federate__fed1/core/federated/network/federate__fed1.config`

The RTI's config will be saved such as `test/C/fed-gen/SimpleFederated/src-gen/rti.config`. However, this RTI config is only for local environments.
```
./federate__fed1 -sst federate__fed1.config
./RTI -sst rti.config
```

# Refactoring
### federate.c
`lf_create_server()`: 
Separate the part sending the address_advertisement message to the RTI. The implementation is in each lf_STACK_support.c. In TCP connections, it will send the port as it did. In MQTT connections, it will do nothing, because it does not need the RTI to connect to another federate. (The MQTT will do a handshake using the federateID, which is already known).

`lf_connect_to_federate()`:
This is a very large function having three parts.
1-1. Send the RTI an ADDRESS_QUERY message, sending the remote federate's ID which it wants to connect to.
1-2. Receive ADDRESS_QUERY_REPLY message from RTI, getting the port number and IP address of the requested federate.
2. Connect to the remote federate, using the port and IP address.
3-1. Send my federateID and federationID
3-2. Receive an ACK.

I splitted this into three parts, using helper functions. However, part 1, sending address query messages, will have different implementations depending on the network communication type. So, it is defined in netdriver.h, and implemented in lf_STACK_support.c

### Change response message of RTI to federate MSG_TYPE_ADDRESS_QUERY
 ** THIS IS ALREADY FIXED IN #392 
The federate first sends MSG_TYPE_ADDRESS_QUERY with total 5bytes (1 Header + 2 federate id).
Then, the RTI responds with the same header MSG_TYPE_ADDRESS_QUERY with (1 Header + 4 port_num + 4 IP address).

Due to my approach needing unique message headers, the response of the RTI will be MSG_TYPE_ADDRESS_QUERY_REPLY.


edwardalee (commented):
I realize this is just an early draft, but this looks like a good start, working in the right direction.  A couple of suggestions that may save time later:

1. You may want to base your changes on the federated-cleanup branch of reactor-c, which will be merged soon, I think.  It includes quite a few changes to the networking code.
2. Our coding style uses spaces only for indentation, no tabs.
3. Our coding style puts opening brackets on the same line and always encloses if statements in brackets when a new line is involved.  E.g., instead of:

```
int  netdrv_open(netdrv_t *drv)
{
	if (!drv)
		return -1;
	return drv->open(drv);
}
```
it should be
```
int  netdrv_open(netdrv_t *drv) {
    if (!drv) {
        return -1;
    }
    return drv->open(drv);
}
```
or
```
int  netdrv_open(netdrv_t *drv) {
    if (!drv) return -1;
    return drv->open(drv);
}
```

4. Header files should always document functions using JavaDoc style documentation.

Jakio815:
@edwardalee Thank you for having a look. I'll keep in my what you mentioned. Especially for the indents and coding style, it was just the auto formatted. I'll fix it. Also I'll add up the documentation for header files. Thank you.

Jakio815:
> The reason for the error in [handle_address_query](https://github.com/lf-lang/reactor-c/blob/5ead7cc5055a42b20875069618dbe7fe24613362/core/federated/RTI/rti_remote.c#L593) is that [`get_port`](https://github.com/lf-lang/reactor-c/blob/5ead7cc5055a42b20875069618dbe7fe24613362/core/federated/RTI/rti_remote.c#L616) invokes
> 
> https://github.com/lf-lang/reactor-c/blob/9d2c3d574e27cd72d96aac8a32ae7636e87bb173/core/federated/network/lf_socket_support.c#L29
> 
> . So I made [`get_port`](https://github.com/lf-lang/reactor-c/blob/5ead7cc5055a42b20875069618dbe7fe24613362/core/federated/RTI/rti_remote.c#L616) returns `-1` if the socket is not initialized.
> Yet, the below line reports a segmentation fault now. @Jakio815 could you modify this part?
> 
> https://github.com/lf-lang/reactor-c/blob/5ead7cc5055a42b20875069618dbe7fe24613362/core/federated/RTI/rti_remote.c#L617

Thanks for your help. It seems that now the weird accept() failing does not appear. Still don't know why it failed at that point, but really thanks for your help!

chanijjani:
@Jakio815 I updated the reactor-c code so that the `lfc-dev` build with `SimpleFederated.lf` and the `RTI` build are done successfully. I would appreciate it if you could check it when you have time.
When implementing the `comm-type` option in the `SimpleFederated.lf` finishes, I'll update the netdriver branch in the lingua-franca repository.

include/core/federated/federate.h
```diff
@@ -345,7 +340,7 @@ int lf_send_message(int message_type, unsigned short port, unsigned short federa
  * information is needed for the RTI to perform the centralized coordination.
  * @see MSG_TYPE_NEIGHBOR_STRUCTURE in net_common.h
  */
-void lf_send_neighbor_structure_to_RTI(int);
+void lf_send_neighbor_structure_to_RTI(netdrv_t* netdrv);
```
Jakio815, line 343: Check.

PR #331, opened 2024-01-19
Unit tests for the RTI and fix of bug revealed by test

byeong-gil:
This PR adds unit tests for the RTI. For now, these tests only cover the function `update_min_delays_upstream` in the common RTI. More tests will be added in the future PR.

The PR also includes a bug fix revealed by the unit tests. The order of addition is rearranged when calculating the minimum delay between scheduling nodes.


lhstrh:
@byeong-gil this PR is still listed as a draft. Can we merge it?

byeong-gil:
@lhstrh No, I think we need to add sufficient test cases before merging it. I will supplement this PR ASAP.

byeong-gil:
I found a problem in update_min_delays_upstream while adding unit tests. The error is caused by the non-commutative property of tag addition. 

For example, if there is a federated lf program like A --/1 sec/--> B --/0/--> C, the minimum delay between A and C should be `(1 sec, 1)`. However, it's calculated as `(1 sec, 0)` because of the line below. `path_delay` is `(0, 1) + 1 sec = (1 sec, 0)`. Generally, let's call the delay between A and B `Td_AB` and the delay between B and C `Td_BC`. We have to calculate `Td_AB + Td_BC`, **not** `Td_BC + Td_AB`. https://github.com/lf-lang/reactor-c/blob/85d54a47af1925b1c52e7e3afba2ec66bc505128/core/federated/RTI/rti_common.c#L320

Fortunately, the tag addition is associative, I could fix this bug easily by changing the direction of the calculation in https://github.com/lf-lang/reactor-c/pull/331/commits/97291c9a72588eb78f45ead264d9fc27239e17db.

byeong-gil:
@lhstrh How about merging this branch and expanding the test coverage in a future PR? I think this is a good time to do it because we got a test for the function `update_min_delays_upstream` which disclosed an issue. And the issue is fixed in this PR.

If you agree, I'll mark this PR as ready and create an issue about increasing the coverage.

core/federated/RTI/rti_common.c
```diff
@@ -316,8 +316,15 @@ static void _update_min_delays_upstream(
...
-        tag_t path_delay = lf_delay_tag(delay_from_intermediate_so_far, intermediate->upstream_delay[i]);
+        // Add connection delay to path delay so far. Because the delay addition is not commutative
+        // but associative, we have to convert the delay to a tag and add to the path delay.
+        tag_t connection_delay = (tag_t) {.time = intermediate->upstream_delay[i], .microstep = 0};
+        if (intermediate->upstream_delay[i] == NEVER) {
+            connection_delay = ZERO_TAG;
+        } else if (intermediate->upstream_delay[i] == 0) {
+            connection_delay.microstep++;
+        }
+        tag_t path_delay = lf_tag_add(connection_delay, delay_from_intermediate_so_far);
```
edwardalee, line 327: I'm confused about this change.  Isn't this exactly what `lf_delay_tag` is supposed to do?  Perhaps the bug is actually in the implementation of that function?

byeong-gil: The most important change in this part is the order of operands. In `main`, we calculate `delay_from_intermediate_so_far` + `connection_delay`. However, when a tagged message propagates, the message's tag will be calculated like `intermediate->upstream_delay[i]` + `delay_from_intermediate_so_far`. For example, if we are finding the minimum delay between A and C where the structure looks like A --/1 sec/-->B--/0/-->C and `intermediate` is B, `intermediate->upstream_delay[i]` is `1 sec`, and `delay_from_intermediate_so_far` is `(0, 1)`. The proper minimum delay is `(1 sec, 1) = 1 sec + (0, 1) ≠  (0, 1) + 1 sec`. 

Here, the problem with `lf_delay_tag` is, that it operates `tag_t` + `instant_t` where `connection_delay` is `instant_t` and `delay_from_intermediate_so_far` is `tag_t`. 

Thus, what I did here is, convert `intermediate->upstream_delay[i])` (`instant_t`) to `tag_t` and add `delay_from_intermediate_so_far` to it using `lf_tag_add`. An alternative way to solve this problem would be 
```
tag_t connection_delay = lf_delay_tag(ZERO_TAG, intermediate->upstream_delay[i]);
tag_t path_delay = lf_tag_add(connection_delay, delay_from_intermediate_so_far);
```

edwardalee: I see!  Good catch!  This really highlights the value of unit tests.

lhstrh:
> @lhstrh How about merging this branch and expanding the test coverage in a future PR? I think this is a good time to do it because we got a test for the function `update_min_delays_upstream` which disclosed an issue. And the issue is fixed in this PR.
> 
> If you agree, I'll mark this PR as ready and create an issue about increasing the coverage.

Yes, this looks ready to merge. Please convert to "ready for review."

lhstrh (commented):
Generally, I think these changes are great. What I would prefer to see is some comments to explain each test and argue why the expected outcomes are reasonable. This will help maintenance of the code considerably.

byeong-gil:
> Generally, I think these changes are great. What I would prefer to see is some comments to explain each test and argue why the expected outcomes are reasonable. This will help maintenance of the code considerably.

Thank you for reviewing, @lhstrh! I added comments on the changes in https://github.com/lf-lang/reactor-c/pull/331/commits/4b8b9537a7cc660342a3edaa501666aca6698eb8!

lhstrh (approved):
:rocket:

hokeun (approved):
Very nice and clean code.

PR #335, opened 2024-01-22
Fix compiler warning in adaptive scheduler

petervdonovan:
This fixes an issue that was reported this morning by @edwardalee.

### DIFF
--- a/core/threaded/scheduler_adaptive.c
+++ b/core/threaded/scheduler_adaptive.c
@@ -44,6 +44,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef MAX_REACTION_LEVEL
 #define MAX_REACTION_LEVEL INITIAL_REACT_QUEUE_SIZE
 #endif
+
+void try_advance_level(environment_t* env, volatile size_t* next_reaction_level);
+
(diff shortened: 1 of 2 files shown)


core/threaded/scheduler_adaptive.c
```diff
@@ -44,6 +44,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef MAX_REACTION_LEVEL
 #define MAX_REACTION_LEVEL INITIAL_REACT_QUEUE_SIZE
 #endif
+
+void try_advance_level(environment_t* env, volatile size_t* next_reaction_level);
```
edwardalee, line 48: Why is this not a `#include` instead?  Seems to risk divergence of the function signature.

edwardalee (approved):
Minor question added, but otherwise LGTM.  Thanks!

core/threaded/scheduler_adaptive.c
petervdonovan: It is only a single signature and I did not want to take the time to find out if it still compiles after bringing in the whole header file (although I'm pretty sure it would). And even if it does still compile, dragging in complex trees of header files including other header files is reputed to be a big source of slowness in C/C++ build systems.

PR #339, opened 2024-01-26
Use Zephyr's CMake extension to correctly build and link reactor-c

erlingrj:
See https://github.com/lf-lang/lingua-franca/pull/2167


erlingrj:
@petervdonovan would you mind taking a look at this since I believe you authored the CMake build system? There are two changes:
1. I am changing the library name from `core` to `reactor-c` to avoid a naming collision with an ESP32 library called libcore.
2. I am moving some Zephyr conditional building into this CMake lists. To make it work I would like to know if we are targeting Zephyr **before** we define the target (because for Zephyr we must use their Cmake extension). So thus, I gather all the sources in a list (you previously did this for debug printing) and add it to the target sources at the end. I also make a list for compile defs, currently it is only used for defining a PLATFORM_ZEPHYR. PLATFORM_NRF52 etc.

core/CMakeLists.txt
```diff
@@ -72,38 +81,36 @@ if(DEFINED FEDERATED_AUTHENTICATED)
...
   endif()
 endif()
 
-# Link with thread library, unless if we are targeting the Zephyr RTOS
+# Link with thread library
 if(NOT DEFINED LF_SINGLE_THREADED OR DEFINED LF_TRACE)
-    if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Zephyr")
-        find_package(Threads REQUIRED)
-        target_link_libraries(core PUBLIC Threads::Threads)
-    endif()
```
petervdonovan, line 90: Is this change the reason for the test failure?

petervdonovan (approved):
These changes look good and I have confidence that when the tests are passing this PR will be good enough to merge.

core/CMakeLists.txt
erlingrj: Yes probably.

erlingrj:
Thanks Peter. 

I updated the lf-ref back to `master`, it should work when the corresponding lingua-franca PR is through the merge queue.

PR #340, opened 2024-01-26
Rename semaphore.h/c to lf_semaphore.h/c

edwardalee:
This PR addresses a collision that can occur on Linux where CMake tries to include a built-in Linux semaphore.h file (in /usr/include or /usr/local/include) and instead finds the Lingua Franca semaphore.h file.  This prevents the browser-ui and mitt examples in the playground from compiling on Linux.  To fix this, we rename the LF files lf_semaphore.h/c.


petervdonovan (approved):
LGTM

It has long been on our to-do list to change the include paths so that our own headers are disambiguated better, but I'm in no hurry to do that and this is a good quick fix.

Issue #289, opened 2023-10-07
Why do we prefer CLOCK_MONOTONIC over CLOCK_REALTIME


erlingrj:
According to the manpages (`man clock_gettime`):

CLOCK_MONOTONIC: Counts nanoseconds since boot and is not affected by discontinuous jumps in system time. But it does not count time when the system is suspended (what does that mean?)

CLOCK_REALTIME: Gives the system POSIX time since 1970. Affected by discontinuous jumps.

While discontinuous jumps could be disadvantageous, I do believe that the linuxPTP tools use CLOCK_REALTIME. (I will ask Henrik about this). Also, seems dangerous that the LF clock should remain unchanged during system suspension.

henrikau:
As the name suggests, CLOCK_MONOTONIC is a monotonically increasing clock source. It will never make large jumps but instead adjust the phase to adjust the time as the local clock drifts. CLOCK_MONOTONIC_RAW is the more extreme example of this, which will not even adjust the phase but instead provide a semblance of a clock by scaling the value of a CPU counter.

CLOCK_REALTIME is kept as close to the *real* time in your clock domain as possible. This means that if you have 2 systems, the only clock that makes sense for the two is CLOCK_REALTIME, you should not compare CLOCK_MONOTONIC between two machines!

*However*, when you start setting timeouts or timers that need to fire *very* closely to your target, you must be a bit careful with CLOCK_REALTIME as this clock is free to move backward in time. In netchan we used CLOCK_MONOTONIC for this reason and read the PTP Time directly from the NIC. However, this opens you up to a lot of extra complexity. At the same time, if you have microsecond accuracy as a target, you should disable frequency scaling and avoid moving between c-states (see https://github.com/henrikau/net_chan/blob/main/src/netchan.c#L1012 for an example)

If you have a PTP synchronized domain and the clocks are reasonably stable, using CLOCK_REALTIME will save you a lot of complexity (which translates into fewer sources of errors) at the cost of suddenly overshooting a timer when the base is yanked backward from under you.

erlingrj:
Thanks for this input Henrik. Very insightful. 
- netchan uses CLOCK_MONOTONIC to do configure very accurate timers to wakeup a thread after some duration, but it reads out the PTP clock to calculate the duration. You don't use `phc2sys` at all?
[...]

PR #341, opened 2024-01-28
Fix watchdog termination

erlingrj:
This PR address [this](https://github.com/lf-lang/lingua-franca/issues/2150) issue. It also moves the watchdogs to the environment struct and should be merged with [this](https://github.com/lf-lang/lingua-franca/pull/2172) PR in lingua franca. 

This PR does a couple of things:
- Solves race-condition with the watchdog threads on termination.
- Creates a single, non-terminating thread per watchdog instead of terminating at each timeout
- Registers the watchdog trigger at a logical tag equal to the expiration.
- Brings watchdog support to the Zephyr target, for this to work we must know, compile time, of the number of watchdogs so we can pre-allocate a stack for the watchdog threads.


erlingrj:
Zephyr fails because additional "user threads" must be specified. We have previously used compile defs for this. We might consider adding a LF_NUM_WATCHDOGS. A problem is however that the zephyr implementation does not work well with dynamically creating and destroying threads. Since I did add a condition variable we could probably make it work without terminating the watchdog thread at each timeout

edwardalee:
I think the idea with watchdogs was that on embedded platforms they would be implemented using a timer rather than thread. This will require an addition to the platform.h abstraction, I think, where the default implementation uses a thread.  But with Zephyr, it seems like maybe there is a way to implement it with a timer?

> Zephyr fails because additional "user threads" must be specified. We have previously used compile defs for this. We might consider adding a LF_NUM_WATCHDOGS. A problem is however that the zephyr implementation does not work well with dynamically creating and destroying threads. Since I did add a condition variable we could probably make it work without terminating the watchdog thread at each timeout

erlingrj:
I remember we discussed an algorithm to implement watchdogs without threads, with an ISR. For Zephyr, I think using a thread is OK.

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+    int ret;            
+    for (int i = 0; i < env->watchdogs_size; i++) {
+        watchdog_t *watchdog = env->watchdogs[i];
+        if (watchdog->base->reactor_mutex != NULL) {
+            ret = lf_mutex_init((lf_mutex_t*)(watchdog->base->reactor_mutex));
+            LF_ASSERTN(ret, "lf_mutex_init failed");
+        }
+        ret = lf_cond_init(&watchdog->cond, watchdog->base->reactor_mutex);
+        LF_ASSERTN(ret, "lf_cond_init failed");
+    }
```
edwardalee, line 33:
```suggestion
    for (int i = 0; i < env->watchdogs_size; i++) {
        watchdog_t *watchdog = env->watchdogs[i];
        if (watchdog->base->reactor_mutex != NULL) {
            LF_ASSERTN(lf_mutex_init((lf_mutex_t*)(watchdog->base->reactor_mutex)), "lf_mutex_init failed");
        }
        LF_ASSERTN(lf_cond_init(&watchdog->cond, watchdog->base->reactor_mutex), "lf_cond_init failed");
    }
```

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+}
+
+/**
+ * @brief Terminate all watchdog threads. 
+ */
+void _lf_watchdog_terminate(environment_t *env) {
+    void *thread_return;
+    int ret;
```
edwardalee, line 41: Similar change suggested as above. Local variable ret is not really used.

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+            ret = lf_mutex_unlock(watchdog->base->reactor_mutex);
+            LF_ASSERTN(ret, "lf_mutex_unlock failed");
+            ret = lf_thread_join(watchdog->thread_id, &thread_return);
+            LF_ASSERTN(ret, "lf_thread_join failed");
         }
     }
 }
 
 /**
  * @brief Thread function for watchdog.
```
edwardalee, line 57: An `@brief` tag needs a blank line below it or the rest of the paragraph is interpreted as part of the brief comment by doxygen.

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
- * the watchdog and then invokes the watchdog expiration handler function.
+ * Each watchdog has a thread which sleeps until one out of two scenarios:
+ * 1) The watchdog timeout expires and there has not been a renewal of the watchdog budget.
+ * 2) The watchdog is signaled to wake up and terminate.
  * In normal usage, the expiration time is incremented while the thread is
- * sleeping, so the watchdog never expires and the handler function is never
- * invoked.
- * This function acquires the reaction mutex and releases it while sleeping.
+ * sleeping, so when the thread wakes up, it can go back to sleep again.
+ * If the watchdog does expire. It will execute the watchdog handler and the
```
edwardalee, line 63:
```suggestion
 * If the watchdog does expire, it will execute the watchdog handler and the 
```

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+ * 1) The watchdog timeout expires and there has not been a renewal of the watchdog budget.
+ * 2) The watchdog is signaled to wake up and terminate.
  * In normal usage, the expiration time is incremented while the thread is
- * sleeping, so the watchdog never expires and the handler function is never
- * invoked.
- * This function acquires the reaction mutex and releases it while sleeping.
+ * sleeping, so when the thread wakes up, it can go back to sleep again.
+ * If the watchdog does expire. It will execute the watchdog handler and the 
+ * thread will terminate. To stop the watchdog, another thread will signal the
+ * condition variable, in that case the watchdog thread will terminate directly.
```
edwardalee, line 65:
```suggestion
 * condition variable; in that case the watchdog thread will terminate directly.
```

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+ * The expiration field of the watchdog is used to protect against race conditions.
+ * It is set to NEVER when the watchdog is terminated.
  * 
  * @param arg A pointer to the watchdog struct
  * @return NULL
  */
 void* _lf_run_watchdog(void* arg) {
+    int ret;
```
edwardalee, line 73: Same comment as above. `ret` not needed?

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
         interval_t T = watchdog->expiration - physical_time;
-        lf_mutex_unlock((lf_mutex_t*)base->reactor_mutex);
-        lf_sleep(T);
-        lf_mutex_lock((lf_mutex_t*)(base->reactor_mutex));
+        LF_PRINT_DEBUG("Watchdog %p going to sleep", watchdog);
+        // Wait for expiration, or a signal to terminate
+        if(lf_cond_timedwait(&watchdog->cond, watchdog->expiration) != LF_TIMEOUT) {
+            LF_PRINT_DEBUG("Watchdog %p was cancelled. Terminating", watchdog);
+            goto terminate;
+        }
```
edwardalee, line 89: This looks dangerous to me. `lf_cond_timedwait` uses `pthreads_cond_timedwait` which is, according to the documentation, subject to spurious wake ups.  This could cause the watchdog to randomly appear to have been cancelled when there was no intent to cancel.  I think the cancelling has to depend on more than the return value of `lf_cond_timedwait`.  Would it be sufficient to check that `watchdog->expiration == NEVER`?

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
+    // Here the thread terminates. 
+    watchdog->thread_active = false;
+    ret = lf_mutex_unlock(base->reactor_mutex);
+    LF_ASSERTN(ret, "lf_mutex_unlock failed");
     return NULL;
 }
 
 void lf_watchdog_start(watchdog_t* watchdog, interval_t additional_timeout) {
-    // Assumes reaction mutex is already held.
+    int ret;
```
edwardalee, line 111: Same comment.

core/threaded/watchdog.c
```diff
@@ -2,82 +2,132 @@
...
     if (!watchdog->thread_active) {
-        lf_thread_create(&(watchdog->thread_id), _lf_run_watchdog, watchdog);
+        ret = lf_thread_create(&(watchdog->thread_id), _lf_run_watchdog, watchdog);
+        LF_ASSERTN(ret, "lf_thread_create failed");
         watchdog->thread_active = true;
     } 
 }
 
 void lf_watchdog_stop(watchdog_t* watchdog) {
+    int ret;
```
edwardalee, line 126: Again.

edwardalee (changes requested):
These are very useful changes.  This should fix the long-standing bug but also properly puts the watchdog timer in the environment. However, I'm not sure, but I think there might be a flaw that could cause watchdogs to be randomly cancelled.  Should be an easy fix, though. See detailed comments.

erlingrj:
Thanks Edward. You are right about the spurious wakeups of lf_cond_timedwait. I also just remembered the issue with lf_cond_timedwait needing an absolute wakeup-time wrt CLOCK_REALTIME, however lf_time_physical is wrt CLOCK_MONOTONIC. It is related to our discussion on #289. It means that before going to sleep on a cond_var we need to check the offset between CLOCK_MONOTONIC and CLOCK_REALTIME and adjust for it. 
 
We would make it easier for ourselves if we didn't try to solving the clock synchronization problem within the LF runtime and instead just used CLOCK_REALTIME and require that the underlying system is configured correctly

erlingrj:
In order to properly implement the watchdog termination I need to first get #346 merged.

core/threaded/watchdog.c
erlingrj: Thanks. WIll remove

erlingrj (commented):
THanks edward

core/threaded/watchdog.c
erlingrj: Thanks addressed now

core/threaded/watchdog.c
```diff
@@ -2,83 +2,173 @@
...
     LF_MUTEX_LOCK((lf_mutex_t*)(base->reactor_mutex));
-    instant_t physical_time = lf_time_physical();
-    while (physical_time < watchdog->expiration) {
-        interval_t T = watchdog->expiration - physical_time;
-        LF_MUTEX_UNLOCK((lf_mutex_t*)base->reactor_mutex);
-        lf_sleep(T);
-        LF_MUTEX_LOCK((lf_mutex_t*)(base->reactor_mutex));
-        physical_time = lf_time_physical();
-    }
+    while (true) {
```
edwardalee, line 95:
```suggestion
    while (! watchdog->terminate) {
```

core/threaded/watchdog.c
```diff
@@ -2,83 +2,173 @@
...
 
-    if (watchdog->expiration != NEVER) {
+        // Step 1: Wait for a timeout to start watching for.
+        
+        // Edge case 1: We have already gotten a signal to terminate.
+        if (watchdog->terminate) {
+            goto terminate;
+        }
```
edwardalee, line 102:
```suggestion
```

core/threaded/watchdog.c
```diff
@@ -2,83 +2,173 @@
...
+        if (watchdog->expiration == NEVER) {
+            continue;
+        }
+
+        // Finally. The watchdog actually timed out. Handle it.
+        LF_PRINT_DEBUG("Watchdog %p timed out", (void *) watchdog);
         watchdog_function_t watchdog_func = watchdog->watchdog_function;
         (*watchdog_func)(base);
+        watchdog->expiration = NEVER;
+        watchdog->active = false;
```
edwardalee, line 136:
```suggestion
        if(watchdog->expiration == NEVER) {
            // Watchdog has been stopped.
            // Wait for a signal that we have a timeout to wait for on the cond-var.
            LF_COND_WAIT(&watchdog->cond);
            continue;
        } else {
            // Watchdog has been started.
            watchdog_wait(watchdog);
            
            // At this point we have returned from the watchdog wait. But it could
            // be that it was to terminate the watchdog.
            if (watchdog->terminate) break;

               // It could also be that the watchdog was just stopped
            if (watchdog->expiration == NEVER) continue;

            // If we reach here, the watchdog actually timed out. Handle it.
            LF_PRINT_DEBUG("Watchdog %p timed out", (void *) watchdog);
            watchdog_function_t watchdog_func = watchdog->watchdog_function;
                (*watchdog_func)(base);
            watchdog->expiration = NEVER;
            watchdog->active = false;
        }
```

core/threaded/watchdog.c
```diff
@@ -2,83 +2,173 @@
...
     }
-    watchdog->thread_active = false;
 
-    LF_MUTEX_UNLOCK((lf_mutex_t*)(base->reactor_mutex));
+terminate:
```
edwardalee, line 139:
```suggestion
```

core/threaded/watchdog.c
```diff
@@ -2,83 +2,173 @@
...
-
+    watchdog->terminate = false;
     watchdog->expiration = base->environment->current_tag.time + watchdog->min_expiration + additional_timeout;
 
-    if (!watchdog->thread_active) {
-        lf_thread_create(&(watchdog->thread_id), _lf_run_watchdog, watchdog);
-        watchdog->thread_active = true;
+    // If the watchdog is inactive, signal it to start waiting.
+    if (!watchdog->active) {
+        LF_COND_SIGNAL(&watchdog->cond);
```
edwardalee, line 154: I'm confused about this `active` field.  If `active == false`, then the thread has terminated, right? In that case, there is no point in signaling it.

include/core/threaded/watchdog.h
```diff
@@ -34,15 +34,17 @@ struct watchdog_t {
     instant_t expiration;                   // The expiration instant for the watchdog. (Initialized to NEVER)
     interval_t min_expiration;              // The minimum expiration interval for the watchdog.
     lf_thread_t thread_id;                  // The thread that the watchdog is meant to run on.
-    bool thread_active;                     // Boolean indicating whether or not thread is active.  
+    lf_cond_t cond;                         // Condition variable used for sleeping and termination.
+    bool active;                            // Boolean indicating whether or not thread is active.  
+    bool terminate;                         // Boolean indicating whether or not thread is active.
```
edwardalee, line 39:
```suggestion
    bool terminate;                         // Boolean indicating whether termination of the thread has been requested.  
```

edwardalee (approved):
Looks good!  I've made some code clarity suggestions.  Also, I'm not sure about the `active` field. I inserted a question.

core/threaded/watchdog.c
erlingrj: The watchdog thread has two states:
1. inactive: It is initialized to this. It means that the thread is running, but it is not waiting for a timeout
2. active: it is waiting for a timeout.  Calling lf_watchdog_stop will move it to inactive state. Calling _lf_watchdog_terminate will terminate the thread. A timeout will cause it to move to inactive also.

This is added to avoid having the watchdog thread terminate after a timeout.

erlingrj: So `active == false` means that the thread is running, but it is not waiting on a timeout. This means we should not only set the expiration field, but also signal the condition variable

erlingrj:
@edwardalee thanks for the suggestions improving the readability of the main loop of the watchdog thread. I replied to a comment on the active field somewhere. The active field is only written to by the watchdog thread and it indicates whether it has stopped/timed-out or whether it is actively waiting for a time-out. If it is the former, then a call to `lf_watchdog_start` must also signal the cond-var to wake it up from the inactive state.

If this makes sense, lets get this merged

edwardalee:
Merged this and queued the corresponding merge in lingua-franca.

PR #342, opened 2024-01-30
Plugin API for tracing

petervdonovan:
This PR turns tracing into a separately-compiled library instead of a part of the core runtime.

- [x] pass all tests on platforms other than macOS
- [x] get tracing tools working again
- [x] separate APIs provided by `core` and APIs provided by plugins
- [x] check for consistency in compile arguments at initialization time
- [x] prepare RFC
- [x] define and implement platform API exposed to plugins
- [x] define and implement logging API exposed to plugins
- [ ] allow plugins to be provided via `lingo` (EDIT: This will be a separate effort and should not block this PR)
    - [ ] improve `lingo` test framework
    - [ ] write a failing test
    - [ ] pass the failing test

------------------------

**Update**: I have decided that there is no reason not to review the internal RFC for the tracing plugin before reviewing this PR. The internal RFC is short and probably quick to read and provides explanation about what is going on in this PR.


petervdonovan:
It seems like the build changes made here are not really compatible with Zephyr.

Here is the problem as I understand it. This PR separates out the tracing functionality as a separately-compiled library; the key to getting this to work is binary compatibility. Binary compatibility is the focus here because we want a toolchain-agnostic (and even language-agnostic) interface for plugins, and because, the thinking goes, "everything that we care about is compatible with the C ABI." But this doesn't work well in cross-compilation land when it becomes impossible to ignore that there are many C ABIs depending on various things including ISA.

In particular, it doesn't work well in the case of Zephyr because Zephyr's build system is really not designed with independent static libraries in mind; in fact, I am under the impression that CMake support for Zephyr assumes that you have exactly one build target named `app` that everything goes into; therefore, if you want to use Zephyr to cross-compile a static library, you can't use the CMake support for Zephyr. Instead, judging from [this](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/application_development/external_lib), the best thing you can do is to use the CMake Zephyr support to generate the CFLAGS and CC and other compilation parameters, and then feed those into `make`, which isn't very convenient because you need to add an extra compilation step: first generate CFLAGS etc., then recover all that and feed it into your static library compilation, and then use the resulting static library to feed it into the main compilation. And since this is only marginally a supported use case, it will be hard to get much help with this online (note that the README in the tutorial I linked isn't even complete -- and yet that tutorial is apparently the best resource you'll find on the subject).

Is it acceptable for us to simply say that tracing isn't supported on Zephyr? I'm not sure if tracing on Zephyr even works currently since it requires file system access, so it might be ok to just give up on this. But in that case, it would seem that for other plugin interfaces besides tracing, the approach that we have been talking about will be hard to use while maintaining compatibility with Zephyr.

Tagging @erlingrj in case he knows more about this.

petervdonovan:
`src/federated/EnclaveFederatedRequestStop.lf` has flaked repeatedly with the adaptive scheduler, and it is really improbable that the changes in this PR are causing that. Ignoring the failure.

edwardalee:
> `src/federated/EnclaveFederatedRequestStop.lf` has flaked repeatedly with the adaptive scheduler, and it is really improbable that the changes in this PR are causing that. Ignoring the failure.
> 
> EDIT: The failure was also observed with the NP scheduler, so it probably isn't specific to the adaptive scheduler. Probably we should just move that test to `failing`. We probably should not act as though a test that flakes is actually passing.

Looking at the way this test is designed, for it to fail, it has to be that communication between three processes is taking more than one second.  In a well-behaved system, this test should terminate after just 50 ms, so I think we can safely increase this one second threshold to, say, one minute.  (and the 5s backup in the test to 30s).  The test will only run slowly if the CI infrastructure is running slowly (which seems to be extremely common... I think they must be emulating Macs on 8080 processors...).

Note that it is impossible to test distributed consensus about shutdown time in a non-flaky way.  We can easily prove this.

Note also that this test will be flaky on ANY platform, so the only reason it is flaking on Mac is that the CI implementation of Mac is bad.

I have ordered a Mac to use as a runner so that we can avoid these problems, but keep in mind that these flaky tests are telling us something fundamental about distributed computing: Some things cannot be deterministically tested in finite time.

petervdonovan:
> Note that it is impossible to test distributed consensus about shutdown time in a non-flaky way. We can easily prove this.

OK, but we can simulate a distributed system in a system that is not actually distributed. I don't think such proof will apply in this case.

> In a well-behaved system, this test should terminate after just 50 ms, so I think we can safely increase this one second threshold to, say, one minute. (and the 5s backup in the test to 30s).

OK, this will help, but we have been doing these kinds of things for a while. I perceive them as ad hoc solutions that use up time for us, make our tests complicated, and do not always work. The system may sometimes be a few orders of magnitude slower than we expect; how many orders of magnitude more leeway will we have to give before the flakiness is reduced to zero, and how many more orders of magnitude will we be able to "get back" by switching to new hardware? I think it is reasonable to start to give up on these kinds of approaches and instead devote resources toward a search for something else that is radically different and possibly better.

We have discussed using a simulated physical time instead of the actual physical time, and I consider that to be the way forward.

edwardalee:
Yes, I agree that the right solution is to have a simulated physical time platform implementation and use that in CI. However, we should be careful, because many of the flaky tests do not fundamentally need to be flaky (unlike this one).

I also agree that there is an interesting research problem here. In fact, I have not been finding the problem of testing dull.  Some of the previously flaky tests I have figured out how to make non-flaky even with unbounded variability in physical time. This has actually taught me some things about distributed coordination (what is possible and what is not).  This is a truly deep topic, related to fundamental principles in physics and with some pretty cool math (I've been learning about "sheafs", which I recently learned can be used to model the record merge problem in distributed databases). The CALM theorem (from Hellerstein) comes into play as well, and relates in very interesting ways to the partial order models in process algebras.

erlingrj:
That sounds really exciting, Edward. 

I also think the idea of introducing a third timeline, the simulated physical timeline, is interesting, not only for federated programs. I think this should be part of a user-facing LF test framework. (I.e. not only for testing our runtimes, but tools to let users test their programs). I think particularly it is useful to enable simple and precise testing of various physical time-related fault scenarios. E.g. this reactions is triggered with 10 msec lag, that should trigger a deadline miss. And now with the watchdogs there is another primitive where the user probably would like to test various fault scenarios.

For simulation of federated programs we could consider integrating with something like the NS2 network simulator. This could enable quite realistic simulation of various federated topologies communicating over different media and with different protocols.

petervdonovan:
Oops, I had promised to clean up all the merge conflicts last night but new ones cropped up this morning :upside_down_face:. Taking care of it now.

core/trace.c
```diff
@@ -47,308 +47,70 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-        tag_t* tag,
+        tag_t tag,
         int worker,
         int src_id,
         int dst_id,
         instant_t* physical_time,
         trigger_t* trigger,
         interval_t extra_delay,
-        bool is_interval_start
+        bool is_interval_start  // FIXME: Unused argument. It originally was a micro-optimization which I suspect is not needed, but if there is (statistically significant) empirical evidence that it is needed, then we should use it. Note that without a memory barrier, even its original use may have had no effect (i.e., it could have generated the same assembly code anyway).
```
petervdonovan, line 76: Edward has already explained why we do this. I couldn't find the comment where he explained it, but the idea is that when we are flushing the trace buffer, this can create a large delay that will affect our measurement of the length of a time interval. After further thought, I think I have to acknowledge that we don't need to observe this empirically to know that it will happen. But how important is it to us?

The problem is that if we want to address this, we will have to add an API function that prepares to register a tracepoint. This API function would flush the trace buffer if needed. After calling it we could then get the physical time. Alternatively, we could allow the tracepoint function to get the physical time. But the bottom line is that if we want to deal with this, the API will be more complicated. I do not think it is worth it but if anyone disagrees I will happily do what is needed.

petervdonovan:
This PR is failing because it is waiting on the PR in `lingua-franca` that will make `lingua-franca/master` compatible with `reactor-c/main`. After that I will fix the conflicts and make sure this PR is good to merge.

lhstrh:
Which PR is that?

petervdonovan:
https://github.com/lf-lang/lingua-franca/pull/2172

petervdonovan:
`EnclaveFederatedRequestStop` failed again on macOS despite the fact that we are using something that has recently had lingua-franca/main merged into it. This probably means that the measures we took to make `EnclaveFederatedRequestStop` less likely to flake did not work.

lhstrh:
> `EnclaveFederatedRequestStop` failed again on macOS despite the fact that we are using something that has recently had lingua-franca/master merged into it. This probably means that the measures we took to make `EnclaveFederatedRequestStop` less likely to flake did not work.

?? I only see a Zephyr failure?

edwardalee:
> `EnclaveFederatedRequestStop` failed again on macOS despite the fact that we are using something that has recently had lingua-franca/master merged into it. This probably means that the measures we took to make `EnclaveFederatedRequestStop` less likely to flake did not work.

I don't see any mechanism for this test to flake. Do you have a pointer to the run that failed? If this test is failing, then we have a real bug, I think.

petervdonovan:
> > `EnclaveFederatedRequestStop` failed again on macOS despite the fact that we are using something that has recently had lingua-franca/master merged into it. This probably means that the measures we took to make `EnclaveFederatedRequestStop` less likely to flake did not work.
> 
> I don't see any mechanism for this test to flake. Do you have a pointer to the run that failed? If this test is failing, then we have a real bug, I think.

[This is the run where I observed this.](https://github.com/lf-lang/reactor-c/actions/runs/7894868427/job/21546301210). The lingua-franca version (confirmed in the logs for that run) is d739ff6c9a1385d5fdc8cae11dcfa15e30d5ef54. [This is the version of EnclaveFederatedRequestStop at d739ff6c9a1385d5fdc8cae11dcfa15e30d5ef54](https://github.com/lf-lang/lingua-franca/blob/d739ff6c9a1385d5fdc8cae11dcfa15e30d5ef54/test/C/src/federated/EnclaveFederatedRequestStop.lf); you can see that this includes last week's changes to the test, and there are no stop times except the desired one at time 50 ms in that test. The reason for test failure is **timeout**.

petervdonovan:
[SmallDelayDecentralized flaked on macOS as well.](https://github.com/lf-lang/reactor-c/actions/runs/7897015526/job/21551893479) I'm assuming that this is one of those "spurious/expected" failures because the error message is "FATAL ERROR: STP violation at tag (0, 3). This should not happen because the STP offset is large" which goes back to my earlier comment about "how large is large enough".

I think we can ignore that and say the tests are basically passing at this point.

petervdonovan:
[Yet another flaky failure in EnclaveFederatedRequestStop on macOS.](https://github.com/lf-lang/reactor-c/actions/runs/7897015526/job/21582087674)

core/trace.c
lhstrh: @edwardalee OK to go ahead and merge as-is?

edwardalee: Ok with me to merge this.

lhstrh (approved):

core/federated/RTI/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ else()
 endif()
 
 set(IncludeDir ../../../include/core)
+include_directories(../../../include)
```
cmnrd, line 53: There should be an explicit dependency on core, but no need to fix this now.

core/federated/RTI/CMakeLists.txt
```diff
@@ -76,12 +77,22 @@ add_executable(
...
 
 IF(CMAKE_BUILD_TYPE MATCHES DEBUG)
     # Set the LOG_LEVEL to 4 to get DEBUG messages
     message("-- Building RTI with DEBUG messages enabled")
-    target_compile_definitions(RTI PUBLIC LOG_LEVEL=4)
+    set(LOG_LEVEL 4)
 ENDIF(CMAKE_BUILD_TYPE MATCHES DEBUG)
+target_compile_definitions(RTI PUBLIC LOG_LEVEL=${LOG_LEVEL})
+
+include(${CMAKE_CURRENT_LIST_DIR}/../../../plugin-defaults/trace/CMakeLists.txt)
```
cmnrd, line 91: This include is an artifact of the fact that the RTI is treated as a separate cmake project, which it probably shouldn't. This is something we can resolve in a later PR.

core/reactor.c
```diff
@@ -51,6 +51,26 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // Global variable defined in tag.c:
 extern instant_t start_time;
 
+int lf_thread_id() {
+    return 0;
```
cmnrd, line 55: These functions look sus. Why are they needed?

include/plugin-apis/version-abi.h
```diff
@@ -0,0 +1,27 @@
+/**
+ * @file version-abi.h
+ * @author Peter Donovan (peter@xronos.com)
+ * @brief API for runtime plugins to use to sanity-check compatibility with the
```
cmnrd, line 4: Could you explain how this works?

cmnrd (changes requested):
Thanks Peter! I think this is a great start towards modularizing the C runtime and providing clear extension points and APIs.

This is going to be a long comment. Please bear with me. Before I comment on this specific PR, I would like to state my view on the goal of modularizing reactor-c and how the end-product could look.

# Modularization of reactor-c

In the long term, I would like to see a structure like this:
```
reactor-c/
├── CMakeLists.txt
├── federate/
│   ├── CMakeLists.txt
│   ├── include/
│   └── src/
├── logging/
│   ├── CMakeLists.txt
│   ├── include/
│   └── src/
├── platform/
│   ├── CMakeLists.txt
│   ├── include/
│   └── src/
├── rti/
│   ├── CMakeLists.txt
│   ├── include/
│   └── src/
├── runtime/
│   ├── CMakeLists.txt
│   ├── include/
│   └── src/
└──  trace/
    ├── CMakeLists.txt
    ├── include/
    └── src/
```
where each subdirectory is a module and each module is represented by a cmake target (or multiple sub-targets if we need to separate the interface headers from their implementation, like in the case of `trace`). The modules should depend on each other and the dependency graph needs to be acyclic. If we can achieve this, we truly have reusable components. E.g., we can use the trace API and the default implementation in other targets. Or we can use the federate module in the C++ runtime. But without doing this separation and truly thinking about the dependencies, we will always have to link the entire reactor-c runtime if we want to use just a tiny component out of it.

With a structure like this, we would also be prepared to factoring out some components, like the RTI, into a separate repo.

Getting there, however, is difficult. The different components are quite intertwined because the dependencies were not a consideration when adding new functionality. For instance, `core/utils/util.c`, which implements logging functionality and in principle should be a component that can be implemented without dependencies, includes `environment.h` and has some preprocessor logging involving the RTI and federates. In other words, it is not an independent component but has baked-in assumptions about how it is used within the code-base.

Probably this bit should become an RFC.

# This PR

I think Peter has encountered similar problems where the tracing logic was intertwined with other components. However, as far as I can see, he has done a great job identifying an independent tracing API and adjusting the code where needed. We should attempt this also for the other components, but this shouldn't be a goal of this PR (otherwise we will never finish it). But I would like to see this PR to make the right steps towards the goal stated above. I think it does already go in the right direction, but it falls a bit short on a few aspects.

Concretely, I have two main concerns:
- There is a masked cyclic dependency between `core` and `plugin-defaults/trace`. The trace implementation requires the API headers defined in `core`, but compiling `core` also requires the trace implementation. I say this is a masked dependency because it is not managed by cmake, but instead resolved by manually adding to the include path. This works, but I don't think it is that much of an improvement compared to before.
- I am not a fan of the `plugin-apis` and `plugin-defaults` directories. There is nothing wrong with this structure per se, but it seems a bit too detailed and convoluted to me, which in the end supports these hidden dependencies that I mentioned. Also, I would avoid naming things plugin (especially since we don't do real dynamic plugins) on the directory level or adding an `api` prefix to headers. These are the headers visible to the users of a library, and it's quite obvious that the public headers define an API.

How to improve this? I have concrete suggestions:

#### Core API module/target

You have done a fantastic job identifying internal APIs that are relevant for this refactoring (`logging-abi.h` and `platform-abi.h`). Let's make them a standalone module. We could create a directory `core/api` and place both `logging.h` and `platform_mutex.h` there (the name is just a suggestion, as I said above, I am not a fan of the abi/api suffix). We also need to add a new cmake target that we could call `core_api` (or `CoreAPI`). This can be done in a nested `CMakeList.txt` or just in the one of `core` (which would be easier). We just need to do this:
```
add_library(core_api INTERFACE)
target_include_directories(core_api INTERFACE <directory>)
```
Then the `reactor-c` target as well as the trace target can simply depend on `core_api` and will automatically add the headers to their include path. This way, we convert the dependency cycle into a tree.

#### Trace module/target with API

We can use the same approach for defining the trace module. I would use the following structure:
```
reactor-c/
├── ....
├── trace/
│   ├── CMakeLists.txt
│   ├── api/
│   │   └── trace.h    // what is now called trace-abi.h
│   ├── include/
│   │   └── trace_impl.h    // what is now called default-trace.h
│   └── src/
│   │   └── trace_impl.c    // what is now called default-trace.c
├── ....
```
The cmake file would define two targets: `trace_api` and `trace_impl`. `trace_api` is an `INTERFACE` library including the `api` directory. `trace_impl` depends on `trace_api` and in addition adds `include` using `PRIVATE` and also adds the source file. Then core can simply declare a dependency on `trace_impl` if it needs to. And whoever wants to provide an external implementation can depend on `trace_api` to get the API header.


Does this make sense? It still doesn't fully resolve that things are intertwined, but as next steps we could try to actually factor out a logging module and a platform module.

I am also happy to help make those changes, just let me know.

edwardalee:
One small comment:  API stands for application programming interface.  A plugin is not an application.  I think we should reserve the term API for the functions/macros that application programmers use, like lf_set and lf_schedule.

cmnrd:
reactor-c is also not an application. API is an established term for describing the interface of libraries (which the trace module is). Why should we deviate from this? We can certainly think about how we qualify different APIs. This would be an internal API, or plugin API, or just the trace API. This is different from the API that we expose to LF users. Perhaps that should be called LF API? Or reactor API?

edwardalee:
Right. reactor-c is not an application.  An application is a Lingua Franca program.

cmnrd:
Ok, I see your point. But it is really not how the term API is defined and used today (I guess the 'application' in it is a historic artifact).

include/core/federated/network/net_common.h
```diff
@@ -256,7 +256,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * FIXME: This could use the latency estimates that were
  * acquired during initial clock synchronization.
  */
-#define DELAY_START SEC(1)
+#define DELAY_START MSEC(100)
```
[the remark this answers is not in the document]
petervdonovan: It is not. Good catch.

include/plugin-apis/impl-in-plugin/trace-abi.h
```diff
@@ -0,0 +1,56 @@
+#ifndef TRACE_ABI_H
```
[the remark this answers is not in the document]
petervdonovan: Probably I am abusing the term. What I am trying to convey is the idea that these headers are intended to describe what is needed in order to ensure binary compatibility, because I believe that binary compatibility offers more flexibility than compatibility at the level of C language/syntax (in particular, implementations do not have to be written in C or C++).

Concretely, this means that to the greatest extent possible, the important information conveyed these header files is not preprocessor definitions, macros, or typedefs, but instead things that map directly to memory layout and calling convention characteristics which are well-defined in the context of a particular host platform.

Even more concretely, when you write `typedef struct my_struct_t my_struct_t;`, I am saying that that does nothing to describe an ABI because it only says the syntax that you can use (you can write `my_struct_t` instead of `struct my_struct_t`). But when you write `typedef struct my_struct_t { int a; double b; }`, I am saying that that helps to describe an ABI because you are saying that you are interested in something that, on the current host platform, has the memory layout of a C struct with an `int` followed by a `double`. Similarly, the macros that allow logging to be excluded from compilation depending on the log level do not describe ABI, but the declarations of the logging functions do, as does the declaration of the function that, during initialization, checks that the core and the plugins are in agreement about the log level.

I also think that it is sometimes possible to have an abstraction layer that works at the level of the C language (API) but not at the binary level (what I have been calling ABI), AND that it is possible in some cases to fix that. For example, the platform abstraction uses typedefs to say what a mutex is, but this abstraction does not work at the binary level because at the binary level, it is observable and problematic that the `sizeof` the two kinds of mutexes is different. A way to solve this in a way that achieves binary compatibility (what I have been calling ABI) is to change the typedef to describe a POINTER to a mutex instead of the mutex itself. This way, the sizeof the different pointers to mutexes is all the same, and this results in a consistent binary interface because all of the functions that access the mutex can use these pointers while abstracting the fact that they point to things that have different memory layout.

include/plugin-apis/impl-in-core/logging-abi.h
```diff
@@ -0,0 +1,193 @@
+// NON-ABI ************************************************************
```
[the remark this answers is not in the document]
petervdonovan: > it probably should be in a different header

You're right, the fact that there are two sections here indicates that there should be two different files. I'll take care of that. The one that contains the macros can be suffixed by `_macros` or `_api`, as you suggested.

> If it is not part of the API...

It seems like you already understand, but I'll add clarification here just to be extra sure.

It is kind of part of an API, but it doesn't work at the binary level. Maybe this distinction is clearer in the context of the discussion above.

This is messy. For the reasons described above, I want these headers to work as **much** at the binary level as possible and as **little** at the C syntax level as possible. But the fact is that these syntactic preprocessor things are useful, and so if another module does happen to be written in C, I want to have this stuff exposed.

plugin-defaults/trace/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+add_library(DefaultTracing STATIC ${CMAKE_CURRENT_LIST_DIR}/trace/default-trace.c)
+target_include_directories(DefaultTracing PRIVATE ${CMAKE_CURRENT_LIST_DIR}/../../include/plugin-apis)
```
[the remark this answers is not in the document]
petervdonovan: OK, no problem. Originally I was confused when you said there was a cycle because I think of these APIs as being upstream of both the core and tracing, but when you characterize these as being part of `core` I see where you are coming from.

I can make them a dedicated header-only CMake target then.

python/lib/pythontarget.c
```diff
@@ -281,7 +281,9 @@ PyObject* py_main(PyObject* self, PyObject* py_args) {
     // Store a reference to the top-level environment
     int num_environments = _lf_get_environments(&top_level_environment);
     LF_ASSERT(num_environments == 1, "Python target only supports programs with a single environment/enclave");
-
+// #ifndef FEDERATED
+//     lf_tracing_global_init(0, 100); // FIXME: what is the actual max number of threads
```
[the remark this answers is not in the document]
petervdonovan: Just a mistake. Good catch.

core/reactor.c
petervdonovan: What is sus about them? I am trying to avoid race conditions by letting different threads access their own dedicated data structures according to their thread ID. In a single-threaded context, we can simply say that the thread ID is always 0.

include/plugin-apis/version-abi.h
petervdonovan: Just pushed more explanation.

petervdonovan:
> In the long term, I would like to see a structure like this:
 
> If we can achieve this, we truly have reusable components.

> With a structure like this, we would also be prepared to factoring out some components, like the RTI, into a separate repo.

I think that you're right about the utility of this and I'm fine with this, but if this is ever done I would want to only put in the work incrementally as it comes necessary. Ideally this code base would have been architected like this from the start, but since it wasn't, I'm happy to go from where we are one step at a time, and I'd be willing to focus on messier changes that might open up more possibilities for future development per hour of developer time to get there -- compilation of reactor-c programs to dylibs that can be embedded in other programs may be one such possibly-messier-but-more-flexibility-per-time-investment change. I still think it's possible that it will never in the lifetime of this code base be necessary to modularize it completely.

> Probably this bit should become an RFC.

OK, sounds good.

> # This PR
 
> There is a masked cyclic dependency between `core` and `plugin-defaults/trace`. The trace implementation requires the API headers defined in `core`, but compiling `core` also requires the trace implementation. I say this is a masked dependency because it is not managed by cmake, but instead resolved by manually adding to the include path. This works, but I don't think it is that much of an improvement compared to before.

See my reply to your other comment. This is a fair point and one way or another it will be addressed.

> I am not a fan of the `plugin-apis` and `plugin-defaults` directories. There is nothing wrong with this structure per se, but it seems a bit too detailed and convoluted to me, which in the end supports these hidden dependencies that I mentioned.

> How to improve this? I have concrete suggestions:

The point is to separate the API which other implementations might use from this particular implementation. In the end this does not seem very different from the structure that you proposed of having the two in the same subdirectory, with two different CMake targets. I'll make the change if people are on board with it though -- for me this is more of a superficial difference and my goal is going to be to pick whichever layout is going to minimize the amount of discussion and disruption. Maybe we should ask whether @edwardalee and @erlingrj have strong opinions about this; if not, I'll just do what you describe.

> Also, I would avoid naming things plugin (especially since we don't do real dynamic plugins) on the directory level...

Marten commented on this, too -- see the discussion [here](https://github.com/lf-lang/lingua-franca/pull/2192#discussion_r1488968429). With multiple people commenting on this same weirdness I guess that makes it harder for me to ignore.

> or adding an `api` prefix to headers. These are the headers visible to the users of a library, and it's quite obvious that the public headers define an API.

Those suffixes are required in order to avoid naming collisions on Arduino. Yes, I know that the Arduino build system is a mess, and yes, this is a hack, but also adding a suffix is really no big deal and does not much change the architecture.

petervdonovan:
FYI (for the purpose of progress tracking): Progress on this PR will continue when it is more clear that the outcome of the various discussions here is final.

core/reactor.c
cmnrd: I was missing the bit that this only applies to single threaded. Why is that though? There is no obvious preprocessor check here. Where is the implementation for threaded?

cmnrd:
> The point is to separate the API which other implementations might use from this particular implementation. In the end this does not seem very different from the structure that you proposed of having the two in the same subdirectory, with two different CMake targets. I'll make the change if people are on board with it though -- for me this is more of a superficial difference and my goal is going to be to pick whichever layout is going to minimize the amount of discussion and disruption. Maybe we should ask whether @edwardalee and @erlingrj have strong opinions about this; if not, I'll just do what you describe.

Ok. As I mentioned in my comment, there is nothing wrong with this structure per se. But it is rather uncommon and it doesn't align well with the goal of separating concerns between modules. I think it is a good idea to place everything related to the trace mechanism in the trace module and keep the interface close to the implementation.

> FYI (for the purpose of progress tracking): Progress on this PR will continue when it is more clear that the outcome of the various discussions here is final.

I will also not be blocking this PR. If we decide to address some of the points later, that is also fine with me.

include/plugin-apis/impl-in-plugin/trace-abi.h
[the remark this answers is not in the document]
cmnrd: Ok, I think I understand where you are coming from and it is great to see that you are thinking about those aspects. I would say that "ABI" and "binary compatibility" are largely misleading here. True binary compatibility you can only achieve when using the same compiler, on the same OS, on the same architecture. The binary interface that actually gets generated is defined by the compiler based on the ABI of the system.

Let me share my view on what you are trying to do. You are trying to separate a generic interface that can be implemented in many different ways from those functions/tools/types/macros that provide a convenience layer on top. In the case of the logging API, for instance, there are the signatures of the log functions doing the actual work, and there is a convenience layer on top provided by the macros. The requirements for designing a user-facing API and an API that acts as an interface between exchangeable components are very different, and this surfaces in the argument you make. Also both are different artifacts, and it is a good idea to separate them. We can also consider giving these two APIs different names in our terminology. Maybe programmer interface and module interface?

> I also think that it is sometimes possible to have an abstraction layer that works at the level of the C language (API) but not at the binary level (what I have been calling ABI), AND that it is possible in some cases to fix that. For example, the platform abstraction uses typedefs to say what a mutex is, but this abstraction does not work at the binary level because at the binary level, it is observable and problematic that the sizeof the two kinds of mutexes is different. A way to solve this in a way that achieves binary compatibility (what I have been calling ABI) is to change the typedef to describe a POINTER to a mutex instead of the mutex itself. This way, the sizeof the different pointers to mutexes is all the same, and this results in a consistent binary interface because all of the functions that access the mutex can use these pointers while abstracting the fact that they point to things that have different memory layout.

I don't quite follow this specific argument. Let's say there is the core runtime and an extension, both of which use the platform abstraction. The extension will only be compatible with the core runtime if both are compiled using the same platform abstraction (for the same platform). If both are complied for the same platform, both agree about the precise type used and hence `sizeof` returns the same value both in the extension and in the core runtime. In a sense, if you expose a type that depends on compiler settings (i.e. macro definitions) on the module interface, then your interface depends on this setting and you can generate many different interfaces. In the case of the platform, I think this is what we want because we statically select the platform. 

By introducing a pointer indirection, the problem becomes masked but not solved. There is still an implicit contract because the core runtime is compiled using platform X with a specific type A and the extension is compiled using platform Y with a specific type B. If `A != B`, then the core and the extension agree on the size of the pointer type (assuming that we compile for the same ISA), but the program will expose undefined behavior because the implementations don't agree on the size and layout of the object. You would need polymorphism to solve this. Surely, this can be done, but why resolve types dynamically if we are statically compiling for a specific platform anyway? 

The pointer solution that you mention pared with a vtable for polymorphism provides a good module interface when you don't know in advance (at compile time) which type(s) you will be dealing with. But (hopefully) we will not have to deal with unknown platforms at runtime. Whenever we can avoid that indirection, we should.

include/plugin-apis/impl-in-core/logging-abi.h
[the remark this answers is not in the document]
cmnrd: I don't think avoiding C syntax is a good goal. The goal should be to identify the actual (minimal) interface module and separate it as much as possible from convenience features on top.

include/plugin-apis/impl-in-plugin/trace-abi.h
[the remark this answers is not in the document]
petervdonovan: > ...we will not have to deal with unknown platforms at runtime

This is not about trying to deal with scenarios where at compile time we do not yet know the platform that the code will have to work with at runtime. That is not my motivation here.

> If both are complied for the same platform, both agree about the precise type used

...unless they are compiled using different toolchains, which might not even have any shared concept of what a "type" means. Sometimes I try to avoid explicitly using the word "Rust" because I would feel a little silly evangelizing any one particular toolchain, but maybe this pointlessly obscuring my point. Really this is all about being able to implement these interfaces using languages other than C. Being able to write code in a language other than C is my real motivation.

> but the program will expose undefined behavior because the implementations don't agree on the size and layout of the object.

But not if the extension never uses the object directly, but instead only ever passes around the pointer -- as would have to be the case if the layout of the object is scrubbed from the interface, and the extension implementor doesn't do anything devious. The functions that work on the object directly would be provided by the core, and the compilation of the extension would not, even under the hood, have to be parameterized by platform-specific knowledge encoded in the platform abstraction.

> The pointer solution that you mention paired with a vtable for polymorphism provides a good module interface when you don't know in advance (at compile time) which type(s) you will be dealing with

Yes, I guess it is accurate to compare what I am talking about (functions being provided by the core) as being like a vtable. I have been thinking more along the lines of these functions being provided at link time rather than their being provided using a vtable of function pointers, but in the case of a third-party plugin providing these functions at link time may indeed have performance consequences similar to the consequences of a vtable (this is why I talk about LTO in the "tracing plugin" internal RFC).

Note that I have not been aware of performance as a priority in the C target for some time, and I/we have not even set up continuous benchmarking in CI for this reason. So although it would be unreasonable for me to categorically dismiss performance discussions (including concerns about vtable pointer lookup overhead), it is true that any such discussions would be limited by not having the context of a shared holistic view of the actual performance behavior and desired performance targets of the C runtime.

Later you also write

> I don't think avoiding C syntax is a good goal.

We should definitely make sure we agree about what our goals are because if the interface is defined syntactically rather than as "the interface is whatever the calling convention/struct layout would be for this C code on the target platform", then that is not compatible with implementing plugins in languages other than C or maybe C++.

lhstrh:
High-level comment/nitpick: it looks like we are mixing snake and kebab case in the naming of our files and directories now. We should not do this.

erlingrj:
I want to inject two comments here also.  I am all in favor of modularizing `reactor-c`, and I would like to contribute to that. 

1. I think we should align this with the efforts to modularize the network interface. That effort is, last time I checked, borrowing architectural ideas from the Linux driver architecture, this could make sense for the whole reactor-c code-base. It is centered around getting classes and hierarchy through structs with function-pointers. Maybe we should organize a discussion/workshop on how a modular architecture for `reactor-c` should look?

2. I agree with Peter's latest comment, we should reach a common consensus on what the goals of such a modularization of `reactor-c` should be. I am not yet convinced that the overall goal should be to introduce modules written in other languages. From my perspective, the key strength of `reactor-c` is that it is written in C and as such is portable to a vast range of platforms, many of which do not have Rust or Cpp support. I don't know how this is changing and how it will look in the future though.

include/plugin-apis/impl-in-plugin/trace-abi.h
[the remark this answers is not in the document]
cmnrd: My vtable argument was not about performance (but its try that this is an aspect, too), its about complexity. To me this additional layer seems a bit too much, but it also seems like I am still missing something. I wasn't considering the Rust binding before, that of course changes the problem a bit. I think I am still not fully understanding the problem though. Perhaps we should have a call next week and sort this out. It seems important for me to understand those underlying problems.

On the naming: I still think that ABI is misleading as ABIs come into play below the language level and are not at all concerned with types. It seems like what you are after is a simplified interface that is safe to be used between different languages. Not sure how to call this though. Maybe FFI? But that could be misleading too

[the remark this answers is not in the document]
lhstrh: Shall we just carve out time for this during our weekly LF meeting?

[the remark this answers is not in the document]
petervdonovan: Sure, that would work good for me. If anyone prefers a smaller meeting with Erling and Christian, that's good for me too. Whatever works.

petervdonovan:
> Maybe we should organize a discussion/workshop on how a modular architecture for reactor-c should look?

I'm not the person who decides whether we do that, but I would be on board with that idea. I think that the rest of us left you out of a couple of discussions about that already 

> I am not yet convinced that the overall goal should be to introduce modules written in other languages. From my perspective, the key strength of reactor-c is that it is written in C and as such is portable to a vast range of platforms, many of which do not have Rust or Cpp support.

Yes, I agree that there are deeply embedded platforms that can only be programmed using very specific toolchains. It is true that the idea of porting reactor-c to Rust has been brought up, but at least I can speak for myself in saying that porting the runtime to another language was not my motivation when I was advocating for these changes which the team is apparently choosing to make. But regardless of whether anyone wants to port the whole thing, cross-language compatibility is still desirable because if we have it,
- We can re-use modules (such as federated execution) from the C target in other targets. This is currently just speculative.
- We can do rapid prototyping of modules that are intended to be used in the C target using a language that offers better developer productivity. This is more or less what I already did when I was working on runtime instrumentation last semester.
- We can develop customized modules that can be substituted for the built-in C target modules and that are intended for a non-embedded use case. This is what we are doing right now with the tracing module.

In addition, there are other reasons to desire separate compilation, like being able to debug build problems one module at a time and especially being able to contain the area of the code base whose compilation is going to be affected by a given preprocessor definition. If we can see what the interfaces between components are going to compile down to without having to grok a mass of preprocessor metaprogramming, then that can make the code easier to understand, in addition to making it friendlier toward cross-language interoperability.

petervdonovan:
Since this PR now has merge conflicts, to run it in CI without doing something hacky I would probably need to merge in `main` (a rebase would probably be too cumbersome). `main` is broken though, so I guess I'll fix `main` first so that it doesn't become confusing to figure out where the problems came from.

lhstrh:
Sorry, that's my fault. I merged it, not realizing there were errors on Windows and Mac... Should we instate a merge queue and run these checks in the merge queue to avoid this in the future?

petervdonovan:
Maybe. I actually don't see anything in that PR which was merged that could have caused this problem, so it could just be that something about the MSVC used in GH actions changed underneath our feet.

core/reactor.c
petervdonovan: All of `core/reactor.c` is for single-threaded and `core/threaded/reactor-threaded` is the counterpart for threaded. To mitigate duplication we have `core/reactor_common.c` which is used in both cases.

logging/api/logging.h
```diff
@@ -0,0 +1,97 @@
+#include <stdarg.h>
```
petervdonovan, line 1: Note to reviewers: This entire file is copy-pasted from parts of `util.h`.

trace/impl/include/trace_impl.h
```diff
@@ -0,0 +1,46 @@
+#include "trace.h"
+
+// FIXME: Target property should specify the capacity of the trace buffer.
```
petervdonovan, line 3: This FIXME was already there; I did not add it.

trace/impl/include/trace_impl.h
```diff
@@ -0,0 +1,46 @@
+#include "trace.h"
```
petervdonovan, line 1: Note to reviewers: Much of this file was copied untouched from the original `trace.c`.

petervdonovan:
I think that this PR is once again ready for review. Changes since it was last reviewed are in the spirit of Christian's comment above. Because the platform directory had already had to be librarified earlier in the development of this branch, and because the logging API was already separate, I found it to be easiest to put those in their own directories already. I also decided to set a precedent of keeping separate `api` and `impl` directories, with different `CMakeLists`, to emphasize that they were different CMake targets; this is something that @cmnrd and I discussed on Tuesday and seemed to agree upon.

Of course this directory structure does not need to be final. We can keep iterating on it later, ideally using an RFC, when this issue rises back to the top of our priority list.

petervdonovan:
Here's a positive unintended consequence of the refactoring that Christian recommended: it seems like VS Code actually understands our code now because it shows syntax errors, including wrong field names, and documentation on hover.

image

I guess this is what we get for doing things the "right" way.

low_level_platform/api/low_level_platform.h
```diff
@@ -19,7 +19,22 @@ extern "C" {
 
 #include "tag.h"
 #include <assert.h>
-#include "lf_atomic.h"
+#include "platform/lf_atomic.h"
+
+// The following two definitions are duplicated from tag.h to break a dependency
+// cycle. TODO: decide whether this (very small) duplication is the right
```
cmnrd, line 25: Another open TODO. The solution looks fine, but the fact that there is a dependency cycle is an indication that something is off. Perhaps you can document the precise cycle in an issue, and we think about how to resolve it later on?

tag/api/tag.h
cmnrd, line 1: Where is the implementation for `tag.h` located now? We should also move this out to an `impl` subdirectory in the future.

test/Tests.cmake
```diff
@@ -41,10 +42,12 @@ endforeach(FILE ${TEST_FILES})
 if (NOT DEFINED LF_SINGLE_THREADED)
     # Check which system we are running on to select the correct platform support
     # file and assign the file's path to LF_PLATFORM_FILE
+    # FIXME: This is effectively a second build script for the RTI that we have to maintain. This is code duplication.
+    # FIXME: We should not be reaching into the platform directory and bypassing its CMake build.
```
cmnrd, line 46: Agreed. Added it to the list in https://github.com/lf-lang/reactor-c/issues/360

cmnrd (commented):
Great! This is quite a significant refactoring, and I am amazed and glad to see that this has an immediate positive effect on the tooling. 

I left a bunch of comments. Most of them are about known issues that I have documented in https://github.com/lf-lang/reactor-c/issues/360 or that we should still document as issues (but I am lacking the details). I also made a few small requests for changes, which I think should be addressable rather quickly. If not, feel free to ignore them. None of them are crucial.

Personally, I am not a fan of the build scripts that this PR adds, but I'll leave it to the main devs of this repo to decide. I think it is more important to try to educate people on how to interact with cmake than to provide simple scripts. In the case where you have to frequently switch between cmake configurations (like you probably did), I would recommend creating multiple build directories with descriptive names, initializing them once with the configuration that you want. Then you just cd into the build directory with the intended configuration and run `make` or `cmake --build` to (re)compile this configuration. This should work flawlessly, even if you update cmake scripts in your changes.

cmnrd (approved):
Looks like commenting does not override a prior "Request changes", so here is my approval.

core/reactor_common.c
```diff
@@ -1171,23 +1172,52 @@ int process_args(int argc, const char* argv[]) {
...
+    #else
+    LF_ASSERT(version.build_config.single_threaded == TRIBOOL_FALSE || version.build_config.single_threaded == TRIBOOL_DOES_NOT_MATTER, "expected multi-threaded version");
+    #endif
+    #ifdef NDEBUG
+    LF_ASSERT(version.build_config.build_type_is_debug == TRIBOOL_FALSE || version.build_config.build_type_is_debug == TRIBOOL_DOES_NOT_MATTER, "expected release version");
+    #else
+    LF_ASSERT(version.build_config.build_type_is_debug == TRIBOOL_TRUE || version.build_config.build_type_is_debug == TRIBOOL_DOES_NOT_MATTER, "expected debug version");
+    #endif
+    LF_ASSERT(version.build_config.log_level == LOG_LEVEL || version.build_config.log_level == INT_MAX, "expected log level %d", LOG_LEVEL);
+    // assert(!version.core_version_name || strcmp(version.core_version_name, CORE_SHA) == 0); // TODO: provide CORE_SHA
```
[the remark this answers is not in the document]
petervdonovan: Done (opened https://github.com/lf-lang/reactor-c/issues/366)

low_level_platform/api/low_level_platform.h
petervdonovan: Actually this is a mistake on my part, `tag.h` is in fact already `#include`d above. So this is being deleted. It is still true that `tag.c` is dependent on `low_level_platform.h`, so in this sense there is a cycle, but this problem (if it is truly a problem) can be resolved later.

include/core/trace.h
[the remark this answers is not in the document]
petervdonovan: Done

include/plugin-apis/impl-in-plugin/trace-abi.h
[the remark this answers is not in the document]
petervdonovan: I think we have reached a collective understanding of the situation. The term "ABI" has been removed from this PR because I agree that it does not capture the intended constraint that I would like to put on certain interfaces between modules.

Marking this thread as resolved.

core/trace.c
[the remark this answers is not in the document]
petervdonovan: Done

include/plugin-apis/impl-in-core/logging-abi.h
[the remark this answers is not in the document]
petervdonovan: Marking as resolved since at this point we do indeed have two separate header files, `logging.h` and `logging_macros.h`, and because at this point we do seem to have a good shared understanding about what I mean about not working at the C syntax level and why.

petervdonovan:
Looks like this PR has 2 approvals, so I don't think more people are obligated to give another in-depth review.

I'm going to leave this open for at least 24 hours in case anyone else has comments, and also to give time for the companion PR https://github.com/lf-lang/lingua-franca/pull/2192 to get approved, and then I'll merge this.

erlingrj:
Great work Peter! I am curious to understand the need for two platform API. This should not block the merging, I am just interested to hear your thoughts.

 It seems that the "higher-level" platform API does not really provide any functionality, and just calls the lower-level platform API. Is the problem you are solving the fact that we have these `typedef  pthread_t lf_thread_t` etc, which are platform-dependent? And you don't really want the tracing plugin to care about whether we are using C11 or pthreads? 
 
Another solution,which I think is more elegant, could be to define them in `low_level_platform.h` as:
```
typedef struct lf_thread_t {
  int id;
  void * priv
}
```

And so on,

Then each particular platform implementation can use the *priv field for their private data, which would be dynamically created at creation. This would make it so that we cannot just define `lf_mutex_t` and be done, but we always need to create it calling e.g. `lf_mutex_create()` which returns a `lf_mutx_t *` with allocated memory in the `priv` field, the priv field would then contain the pthread_mutex_t. A nice thing is that the pthread-details wont "leak" out of the platform source files. Currently we include pthread stuff in the platform header files and it is thus in scope for all sorts of files, even user-written reaction bodies which needs to include platform.h

This is the sort of inheritance they do in the Linux device driver implementations which I have just started learning about.

petervdonovan:
> It seems that the "higher-level" platform API does not really provide any functionality, and just calls the lower-level platform API.

Correct. This is how I intended the "higher-level" platform API to remain for the foreseeable future -- just a wrapper around the lower-level platform API.

> Is the problem you are solving the fact that we have these `typedef pthread_t lf_thread_t` etc, which are platform-dependent? And you don't really want the tracing plugin to care about whether we are using C11 or pthreads?

Yes, exactly.

> A nice thing is that the pthread-details wont "leak" out of the platform source files. Currently we include pthread stuff in the platform header files and it is thus in scope for all sorts of files, even user-written reaction bodies which needs to include platform.h

Really I was fixated on getting an interface where you know what it will compile to, regardless of compilation options, without having to understand any preprocessor code. This "leaking" is not exactly the problem that I was trying to solve.

However, now that you mention it I think you're right, this is an important added benefit; furthermore, it is solved both by what you describe and by what I have done with `low_level_platform.h` and with `platform.h`. We have already had discussions and taken pains to contain the damage caused by `windows.h`, so this is an issue that I think we already collectively understand and care about, and yes, the same principle applies to containing the areas where pthreads stuff is in scope, too.

> Then each particular platform implementation can use the *priv field for their private data, which would be dynamically created at creation. This would make it so that we cannot just define `lf_mutex_t` and be done, but we always need to create it calling e.g. `lf_mutex_create()` which returns a `lf_mutx_t *` with allocated memory in the `priv` field, the priv field would then contain the pthread_mutex_t.

I see this proposal as being very similar to what I have done with `low_level_platform.h` and `platform.h`. If I understand, the significant difference is that you seem to be proposing that:
1. the mutex and thread representations that use void pointers to abstract the platform differences become the only representations that are made visible to the whole code base. So outside of the platform abstraction, none of the code would be able to see the differences, not even using `sizeof`, unless they dereference the void pointer and make assumptions about what is in the memory that it points to (which would be obviously wrong).
2. we develop a habit of combining these void pointers with other metadata, like thread IDs, to improve code reuse.

Item 2 is obviously a generally good idea to do, in those cases where it makes sense. I would argue that item 2 is compatible with the route that I have already taken and is not dependent on item 1; we can always create structs in `low_level_platform` that have both fields whose size is platform-dependent and fields that we define that are not platform-dependent, and those structs will still overall be like the ones we use already and would have the disadvantages that are addressed by item 1. So item 2 seems like no big deal.

Item 1, however, does represent an important tradeoff. Last Monday I discussed this with Edward and Christian and I promised that my attempts to create this `void*` kind of modularity (where some interfaces don't depend on the preprocessor) would never negatively impact memory footprint at runtime. I promised that there would be an escape hatch.

The tradeoff is as follows. If the preprocessor-independent `void*` approach is the only approach, then the advantage is that the whole code base is brought up into a different and stronger kind of modularity that could force the simplification of a bunch of stuff, as well as making it easier for us to mix and match useful functionalities between the different runtimes, which could potentially break down siloing among LF contributors. In short, it would offer all the advantages of supporting both approaches (as I have done with `low_level_platform` and `platform`), but it would offer them in a way that is even better and cleaner and more strongly enforced. But the disadvantage is that when people are forced to use `void*`, then that can increase the number of allocations that they have to make, and it can increase memory fragmentation -- at least in principle.

Are these disadvantages practically significant? I haven't measured the memory footprint or the performance differences, but I also don't want to have to make those measurements, and I don't want to have to engage in a time-consuming discussion about those measurements. If we set the precedent that anything that could conceivably have overhead is provided with some sort of escape hatch that isn't constrained to have the `void*` kind of modularity, then sure, we get a little more mess in our code base, but we also liberate ourselves from having to worry and discuss too much about this issue. The worry and discussion can be contained to the specific cases where we want to actually implement a specific plugin, and we need to decide which platform interface to use.

PR #344, opened 2024-02-01
Rp2040 multithreaded target support

sberkun:
This PR adds multithreaded support from the RP2040 platform. Builds off of the work in https://github.com/lf-lang/reactor-c/pull/291.

Corresponding PR in lingua-franca: https://github.com/lf-lang/lingua-franca/pull/2178.

Some important notes on this implementation (main differences from Abhi's implementation):
 - semaphores are used as an analog for condition variables. This works because the condition variable only needs to notify at most one other thread. Will trigger a spurious wakeup later if the other thread is not waiting on the condition variable.
 - only one other thread (in addition to the main thread) is allowed. This will be a tricky constraint for the runtime to work with (in particular, scheduling enclaves and watchdog timers will need some work), but will allow for the runtime to work with no context switches.
 
 Edit: this PR is meant as part 1 of a series. This PR is meant to be minimal changes to get the threaded runtime working. The next few PRs will be:
  - modify the runtime so that the main thread also runs a worker
  - merge atomics implementation with Zephyr
  - lf_initialize_clock return value


core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+}
+
+int lf_cond_timedwait(lf_cond_t* cond, instant_t absolute_time_ns) {
+    absolute_time_t a = from_us_since_boot(absolute_time_ns / 1000);
+    bool acquired_permit = sem_acquire_block_until(&(cond->sema), a);
+    return acquired_permit ? 0 : LF_TIMEOUT;
+}
+
+
+// Atomics
```
sberkun, line 323: The atomics implementations are pretty much copy-pasted from Zephyr support. In a future PR, I would like to merge the two somehow. Not going to do it in this PR in case I end up accidentally breaking Zephyr support.

core/CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ endif()
 
 # Link with thread library, unless we are on the Zephyr platform.
 if(NOT DEFINED LF_SINGLE_THREADED OR DEFINED LF_TRACE)
-    if (NOT PLATFORM_ZEPHYR)
+    if (NOT (PLATFORM_ZEPHYR OR ${CMAKE_SYSTEM_NAME} STREQUAL "Rp2040"))
```
sberkun, line 96: I tried `if (NOT (PLATFORM_ZEPHYR OR PLATFORM_RP2040))` but it didn't work. Not sure how `PLATFORM_ZEPHYR` works here.

sberkun: todo: fix, same as https://github.com/lf-lang/reactor-c/blob/e8a0b97917c8ef146cd074e67aa69df43f344bb6/core/platform/CMakeLists.txt#L24C1-L24C30

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+    multicore_launch_core1(core1_entry);
+    *thread = MAGIC_THREAD1_ID;
+    return 0;
+}
+
+int lf_thread_join(lf_thread_t thread, void** thread_return) {
+    if (thread != MAGIC_THREAD1_ID) {
+        return 1;
+    }
+    sem_acquire_blocking(&thread_1_done);
```
gundralaa, line 270: Iirc since the main thread in reactor c just blocks until join completes, this would block core0 from executing anything

gundralaa (commented):
Pretty cool someone’s picking this up :) The only major comment I have is enabling more than 1 worker thread. With only one, it defeats the purpose of having a threaded implementation since the main reactor-c thread does not execute reactions.

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+
+#define MAGIC_THREAD1_ID 314159
+
+void core1_entry() {
+    thread_1_return = thread_1(thread_1_args);
+    sem_release(&thread_1_done);
+
+    // infinite loop; core1 should never exit
+    while (1){
+        tight_loop_contents();
```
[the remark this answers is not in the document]
sberkun: Oops, some of the [multicore examples](https://github.com/raspberrypi/pico-examples/blob/master/multicore/hello_multicore/multicore.c#L24-L25) did this so I thought that maybe returning from core1 was bad. But you're right, when core1_entry returns it goes into a [low power state](https://github.com/raspberrypi/pico-bootrom/blob/ef22cd8ede5bc007f81d7f2416b48db90f313434/bootrom/bootrom_rt0.S#L371). I'll remove this, thanks!

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+static void* thread_1_args;
+static int num_create_threads_called = 0;
+static semaphore_t thread_1_done;
+static void* thread_1_return;
+
+#define MAGIC_THREAD1_ID 314159
+
+void core1_entry() {
+    thread_1_return = thread_1(thread_1_args);
+    sem_release(&thread_1_done);
```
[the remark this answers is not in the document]
sberkun: The [semaphore implementation](https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_sync/sem.c) seems to be based on spinlocks, so should be fine? Is there a reason why a non-multicore compatible semaphore would exist?

core/platform/lf_rp2040_support.c
sberkun: Yup. I should have mentioned this in the PR description, but I plan to start another PR once this one gets merged. Basically:
 - this PR: basic multicore support, only supports one worker thread. As you said, this is kind of pointless, but it's relatively straightforward.
 - next PR: modify the runtime so that the main thread also runs a worker, instead of just waiting for the worker threads. This will allow 2 workers (one main thread, one additional thread) and finally make the multicore support useful. However, this change may potentially affect all platforms, so I want it to be a seperate PR.

core/platform/lf_rp2040_support.c
```diff
@@ -63,13 +64,14 @@ static uint32_t _lf_num_nested_crit_sec = 0;
 
 /**
  * Initialize basic runtime infrastructure and 
- * synchronization structs for an single-threaded runtime.
+ * synchronization structs for a single-threaded runtime.
  */
 void _lf_initialize_clock(void) {
     // init stdio lib
     stdio_init_all();
```
erlingrj, line 71: Should probably check the return value. Also, `_lf_initialze_clock` should return a status value. Because it could fail. But this would require updating all the platform implementations. But it would be an improvement

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
     sem_release(&_lf_sem_irq_event);
```
erlingrj, line 216: Check return value

core/platform/lf_rp2040_support.c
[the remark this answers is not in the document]
erlingrj: Should probably check the return value here

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+    *thread = MAGIC_THREAD1_ID;
+    return 0;
+}
+
+int lf_thread_join(lf_thread_t thread, void** thread_return) {
+    if (thread != MAGIC_THREAD1_ID) {
+        return 1;
+    }
+    sem_acquire_blocking(&thread_1_done);
+    sem_release(&thread_1_done); // in case join is called again
```
erlingrj, line 266: return value

core/platform/lf_rp2040_support.c
```diff
@@ -214,6 +216,170 @@ int _lf_single_threaded_notify_of_event() {
...
+    if (thread_return) {
+        *thread_return = thread_1_return;
+    }
+    return 0;
+}
+
+int lf_mutex_init(lf_mutex_t* mutex) {
+    mutex_init(mutex);
```
erlingrj, line 274: If I remember correctly we need a recursive mutex (Pico SDK has one also). But please look closer into this.

core/platform/lf_rp2040_support.c
erlingrj: Would be good to merge them yes. But another thought is that a scheduler using them extensively will be expensive on Zephyr/Pico and cheap on x86. Maybe we need a new scheduler for 32bit platforms without atomics? Or maybe not :)

erlingrj (changes requested):
This looks good. Just some nitpicks about checking the return values from the Pico SDK. Also look into whether we need a recursive mutex. When getting threaded support for FlexPRET we ran into that issue and had to create a FlexPRET lock implementation that supported recursive calls.

core/platform/lf_rp2040_support.c
sberkun: Definitely should test how much this affects performance once I start benchmarking on the rp2040. A new scheduler would be exciting :)

core/platform/lf_rp2040_support.c
gundralaa: Neat! Sounds like a plan :)

core/platform/lf_rp2040_support.c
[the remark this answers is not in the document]
gundralaa: Maybe just test it out? I think this one was meant for irq notification.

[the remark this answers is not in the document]
sberkun: Did some ad hoc testing on the pico (raw C code, not LF) and semaphores seem to work with both cores. LF programs seem to be working as well

core/platform/lf_rp2040_support.c
sberkun: Right now, `stdio_init_all()` will always return true because both UART and USB are enabled, and UART initialization is infallible. We may want to support LF programs that disable them in the future (in which case this would return false), so I left a comment about that.

core/platform/lf_rp2040_support.c
sberkun: looked at mutex usages and am not sure where recursive mutex is used. Changed to recursive mutex just to be safe

gundralaa (approved):
looks good 👍

sberkun:
@erlingrj just rebased onto the most recent changes in main, would be great if we could get this merged in

lhstrh:
Whoops, @sberkun, it looks like we forgot to merge this, and it has fallen a bit behind. Sorry! Could you once again fix the merge conflicts?

lhstrh:
We should get this merged, but there are some conflicts. Maybe @petervdonovan or @erlingrj could help?

sberkun:
@lhstrh @erlingrj Sorry for the delay! This is once again ready for review.

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -202,9 +201,125 @@ int lf_enable_interrupts_nested() {
...
+static void* thread_1_return;
+
+#define MAGIC_THREAD1_ID 314159
+
+void core1_entry() {
+  thread_1_return = thread_1(thread_1_args);
+  sem_reset(&thread_1_done, 1);
+}
+
+int lf_thread_create(lf_thread_t* thread, void *(*lf_thread) (void *), void* arguments) {
```
edwardalee, line 234: It seems like this could be easily generalized to support NUM_CORES rather than just 2.  It's not essential to do that now, so if you don't have time, just mark this resolved and open an issue.  We would need a platform to test it on anyway.

edwardalee (approved):
LGTM!  One comment included that we should not take as blocking the merge.

low_level_platform/impl/src/lf_rp2040_support.c
sberkun: Yup, that would be a good thing to do. This would be very relevant to an stm32 platform if one were added in the future. I'll open an issue for it

sberkun:
@erlingrj looks like this is blocked because of requested changes; could you take a final look at this and unblock it?

erlingrj (approved):
Looks mostly good. But I have some questions about the cond bar implementation. I think the spurious wake-up problem in pthreads is not something we need to copy.

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -202,9 +199,120 @@ int lf_enable_interrupts_nested() {
...
+  thread_1_args = arguments;
+  num_create_threads_called += 1;
+  sem_init(&thread_1_done, 0, 1);
+  multicore_launch_core1(core1_entry);
+  *thread = MAGIC_THREAD1_ID;
+  return 0;
+}
+
+int lf_thread_join(lf_thread_t thread, void** thread_return) {
+  if (thread != MAGIC_THREAD1_ID) {
```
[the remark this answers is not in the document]
sberkun: Correct, but there's no thread handle for thread0 so nothing can join on thread0 regardless.

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -202,9 +199,120 @@ int lf_enable_interrupts_nested() {
...
+
+int lf_cond_broadcast(lf_cond_t* cond) {
+  for (int i = 0; i < NUM_CORES; i++) {
+    sem_reset(&(cond->notifs[i]), 1);
+  }
+  return 0;
+}
+
+int lf_cond_signal(lf_cond_t* cond) {
+  return lf_cond_broadcast(cond); // spurious wakeups, but that's ok
```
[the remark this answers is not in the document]
sberkun: Yes, spurious wakeups would be good to avoid, but it would complicate the implementation significantly to try to avoid them (i'm imagining an array to keep track of which threads are waiting on the condition variable). I took a look at https://github.com/raspberrypi/pico-sdk/pull/1101 which seems like an ideal implementation, but it's very complex.

Typical usage is like
```
do {
  lf_cond_wait(cond);
} while (condition);
```
So spurious wakeups only cause an extra iteration of this loop, which doesn't have much affect performance-wise.

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -202,9 +199,120 @@ int lf_enable_interrupts_nested() {
...
+  for (int i = 0; i < NUM_CORES; i++) {
+    sem_init(&(cond->notifs[i]), 0, 1);
+  }
+  cond->mutex = mutex;
+  return 0;
+}
+
+int lf_cond_broadcast(lf_cond_t* cond) {
+  for (int i = 0; i < NUM_CORES; i++) {
+    sem_reset(&(cond->notifs[i]), 1);
```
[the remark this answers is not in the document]
sberkun: Yup, see other comment about spurious wakeups.

low_level_platform/impl/src/lf_rp2040_support.c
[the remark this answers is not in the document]
erlingrj: The PR you linked to looked complicated yes, we solved it a lot easier in FlexPRET here https://github.com/pretis/flexpret/blob/master/sdk/lib/src/cond.c

Issue #332, opened 2024-01-19
Excessive lag


edwardalee:
At least on macOS, we get a lag (physical time minus logical time) that is always larger than what I would expect.  @lsk567 and I looked at this yesterday with the following extremely simple LF program:

<img width="624" alt="image" src="https://github.com/lf-lang/reactor-c/assets/8513334/097ff4fc-ffbf-4cdb-bdda-51ec7a8809d8">

In this program, the runtime system will call `lf_cond_timedwait()` to wait for physical time to match each multiple of 1ms before invoking the reaction. `lf_cond_timedwait()` calls the POSIX function `pthread_cond_timedwait()` on macOS (and presumably on Linux as well). We inserted the following code at the point in the runtime right after it returns from `lf_cond_timedwait()`.
```
            instant_t now = lf_time_physical();
            tracepoint(env->trace, user_event, NULL, NULL, 0, 0, 0, &now, NULL, 0, false);
```
This inserts a "user event" into the trace file. We then converted the trace file to CSV using `trace_to_csv` and looked at the times at which things are occurring.  The trace looks something like this:

<img width="830" alt="image" src="https://github.com/lf-lang/reactor-c/assets/8513334/45bff8f7-0b19-4a43-8d8a-7878c93b84e9">

Here, the last column is physical time as reported by `lf_time_physical_elapsed()`. The above section of the trace is when the advancing from 2ms to 3ms in logical time is occurring.

What we found was pretty interesting.

1. `pthread_cond_timedwait()` consistently returns later than you ask it to. In the above trace, it is 291us late.
2. The time taken between when `pthread_cond_timedwait()` returns and the reaction gets invoked is pretty small. In the above trace, it is 5us, although it is often as low as 1us. 1us is the smallest time increment we can measure with `lf_time_physical()` because the macOS clock has a resolution of 1us.
3. `lf_cond_timedwait()` sometimes gets invoked many times in rapid succession. It gets called several dozen times rapidly. Why?

The lag we are seeing is mostly due to `pthread_cond_timedwait()`, not to our runtime system.  There are some obvious questions:

1. Is `pthread_cond_timedwait()` using the same clock as `lf_time_physical()`? If not, this could account for what we are seeing.
2. Could we learn at runtime what the offset is between `pthread_cond_timedwait()` and  `lf_time_physical()` and correct for it by pre-subtracting the 200us from the argument given to  `pthread_cond_timedwait()`?

erlingrj:
[...]

PR #346, opened 2024-02-03
Move to CLOCK_REALTIME introduce clock.h and lf_atomic.h

erlingrj:
The main purpose of this PR is to move to pure CLOCK_REALTIME on UNIX systems as discussed here #289.  It gained a little weight and now it includes:
- Move everything to CLOCK_REALTIME
- Remove the epoch_offset calculations which were needed for CLOCK_MONOTONIC
- Handle adding/subtracting clock-sync offsets BELOW platform API instead of in tag.c This means that any time a platform should hand off a physical time, it must add the clock sync offset. And anytime it receives a physical time to sleep until it must subtract the clock-offset.
- Moving the clock-sync dealings down there makes the implementation of wait_until a lot simpler. It enables using lf_cond_timedwait with just a normal instant_t without having to think about clock-sync offsets.
- Make an own API for atomics. The macro stuff did not work because for GNU you could pass in any type, but for windows and for Zephyr++ you needed to pick a certain datatype. Now there is a separate API for 32bit data and 64bit data.
- Make a common implementation of the atomics API for 32bit platforms using enabling/disabling interrupts. This requires having this implemented both for threaded and single-threaded platforms. (Previously it was just implemented for single-threaded)
- Add -Werror (treat warnings as errors) to unit tests and RTI/. We should eventually add this to the core lib also.
- Make NSEC(x) and friends return a interval_t not a long long. Now we can print NSEC(x) with PRINTF_TIME
- Make lf_time_physical thread-safe and 32bit-safe using the atomics API.
- Make updating the clock-sync offset 32-bit safe using the atomics API.

---
Older discussion:
After some consideration I decided to move the application of clock sync offset **below** the platform API. Previously this was done in tag.h, the layer above the platform API. The disadvantage was that we had an asymmetry. Because whenever you wanted to use lf_sleep_until or lf_cond_timedwait you had to remove the clock sync offset before calling it. Now the platform implementations has to apply the clock sync offset before handing off a timestamp, and it will also remove any clock_sync offset from absolute timestamps it receives for sleeping. It introduce some repetition but now the timestamps moving to/from the platform API are more consistent.  

I also implemented a thread-safe and 32bit/64bit safe modification of the clock sync offset with a mutex. This is up for discussion. It makes it a lot costlier to read out the current time when we have clock-sync enabled. 

There is still a FIXME with regards to how we best can make lf_time_physical monotonic and thread-safe on 32bit and 64bit systems. I think it is the same problem as the modification of the clock_sync offset, but more critical as it affects all systems.

Todos:
- [x] Fix docs in the code
- [x] Make lf_time_physical monotonicity enforcing thread-safe and 32bit-safe


erlingrj:
@sberkun, I found the need to update the atomics implementations already know. For the 32bit platforms I think we have to do it by disabling interrupts. So I have moved `lf_disable_interrupts_nested` so it is defined for both threaded and single-threaded context. Will this work for the pico also? There was a comment indicating that core2 was disabled, what happens if core2 is the one making the call?

erlingrj:
@edwardalee @petervdonovan I am asking for your feedback on this. This got a little more involved than I had planned. I need a proper pass through the diff to add proper docs of the new functions etc. But would like to hear what you think about the proposed changes here.

core/federated/clock-sync.c
```diff
@@ -554,4 +554,21 @@ int create_clock_sync_thread(lf_thread_t* thread_id) {
...
+
+#if defined (_LF_CLOCK_SYNC_ON)
+void clock_sync_apply_offset(instant_t *t) {
+    // Read out the current clock sync offset. USe atomics to ensure thread-safety
+    // also on 32bit platforms.
+    instant_t clock_sync_offset_local = lf_atomic_add_fetch64(&clock_sync_offset,0)
+    *t += clock_sync_offset_local;
+}
+
+void clock_sync_remove_offset(instan_t *t) {
```
petervdonovan, line 566:
```suggestion
void clock_sync_remove_offset(instant_t *t) {
```

core/platform/CMakeLists.txt
```diff
@@ -14,18 +14,31 @@ set(LF_PLATFORM_FILES
     lf_rp2040_support.c
 )
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
petervdonovan, line 17: A long long time ago we talked about moving as much of this conditional compilation stuff as possible from CMake to preprocessor. But recently it seems like we have changed our minds and decided that CMake is powerful enough for us to use it for everything, in which case I'm happy having this code here.

petervdonovan (commented):
Did not look carefully at the changes yet, but here is my response to your description of the PR:

>    Move everything to CLOCK_REALTIME

>    Remove the epoch_offset calculations which were needed for CLOCK_MONOTONIC

I am OK with this. AFAIK the possible downside is that in relatively rare cases, the physical time could jump forward (causing widely separated tags to be executed in quick succession) or stop increasing for a while (causing the program to grind to a halt while waiting for "true" physical time to catch up). This is probably OK for many use cases, but I think it might be very bad for some other use cases. In other words, I think that relative time will matter more than absolute UTC time in many embedded systems.

Therefore the only caveat that I would add is that I think that we should leave open the possibility that later, we may make the physical time implementation a plugin that users can swap out. This relates to our previous discussions about simulated physical time.

>    Handle adding/subtracting clock-sync offsets BELOW platform API instead of in tag.c This means that any time a platform should hand off a physical time, it must add the clock sync offset. And anytime it receives a physical time to sleep until it must subtract the clock-offset.
>    Moving the clock-sync dealings down there makes the implementation of wait_until a lot simpler. It enables using lf_cond_timedwait with just a normal instant_t without having to think about clock-sync offsets.

Sounds great. I am all in favor of separating concerns in this way.

>    Make an own API for atomics. The macro stuff did not work because for GNU you could pass in any type, but for windows and for Zephyr++ you needed to pick a certain datatype. Now there is a separate API for 32bit data and 64bit data.

Sounds great. Without link-time-optimization this will have performance consequences, but it has been a long time since I was concerned enough about performance for that to matter to me. Besides, we can add LTO later.

>    Make a common implementation of the atomics API for 32bit platforms using enabling/disabling interrupts. This requires having this implemented both for threaded and single-threaded platforms. (Previously it was just implemented for single-threaded)

Not something I know much about, but this sounds good. Of course disabling interrupts is not sufficient on its own to provide thread-safety on threaded platforms, so I'm not sure if threaded 32-bit platforms can have completely the same implementation as single-threaded.

>    Add -Werror (treat warnings as errors) to unit tests and RTI/. We should eventually add this to the core lib also.

Great, this has been a long-standing TODO.

>    Make NSEC(x) and friends return a interval_t not a long long. Now we can print NSEC(x) with PRINTF_TIME

Sounds good

>    Make lf_time_physical thread-safe and 32bit-safe using the atomics API.
>    Make updating the clock-sync offset 32-bit safe using the atomics API.

Sounds good

include/core/platform.h
```diff
@@ -202,85 +205,12 @@ int lf_cond_wait(lf_cond_t* cond);
 
 /**
  * Block current thread on the condition variable until condition variable
- * pointed by "cond" is signaled or time pointed by "absolute_time_ns" in
- * nanoseconds is reached.
+ * pointed by "cond" is signaled or time pointed by wakeup_time is reached.
```
edwardalee, line 208:
```suggestion
 * pointed by "cond" is signaled or time given by wakeup_time is reached.
```

edwardalee (commented):
Switching to `CLOCK_REALTIME` seems like the right thing to do.  This PR includes some other things that I don't completely understand, however, particularly concerning interrupts.

I'm unclear on one big aspect of this:  In Unix platforms, disabling interrupts is not an option. I'm pretty sure it cannot be done in user space.  The reason we introduced `lf_critical_section_enter` and `lf_critical_section_exit` to the platform API was precisely so you could disable interrupts on platforms where that is possible and necessary. On a Unix platform, I think acquiring a mutex should be sufficient, though if you have user-level interrupt service routines (which are really callbacks), they will have to use `lf_critical_section_enter`.

Is the idea that `core/platform/lf_atomic32.c` will be included only for platforms that can/need to disable interrupts? If so, then I don't think `lf_disable_interrupts_nested` needs to be part of the platform API, so it doesn't have to be implemented on Unix-based platforms.

Also, I'm confused why `core/platform/lf_atomic64.c` only support Arduino...

I've made a few small suggestions as well, but I think I need to understand better this big picture issue before I can meaningfully comment on the rest.

sberkun:
> @sberkun, I found the need to update the atomics implementations already now. For the 32bit platforms I think we have to do it by disabling interrupts. So I have moved `lf_disable_interrupts_nested` so it is defined for both threaded and single-threaded context. Will this work for the pico also? There was a comment indicating that core2 was disabled, what happens if core2 is the one making the call?

It does work for the pico; however, the name is now a bit misleading. `lf_disable_interrupts_nested` should really be named something like `lf_critical_section_nested` since it provides exclusion from both interrupts and other threads.

erlingrj:
Thanks for the feedback everyone. I will address the suggested edits. I think I need to clarify the proposed atomics changes because there were several comments on it.

- The Arduino related macro in lf_atomic64.c is to avoid including it in Arduino build. It is needed because arduino-cli compiles every single file source file it finds. For the other targets only one of lf_atomic32.c or lf_atomic64.c is included.
- If we cannot disable interrupts for UNIX and windows then I agree we should remove this from the common platform API and only have it implemented for the embedded targets (which are the ones using lf_atomic32.c)
- Peter made the comment that disabling interrupts is not necessarily enough. I guess you are referring to multicores? It suggests that my division into lf_atomic32 and lf_atomic64 is not the most descriptive. It is more like a division between platforms supporting atomic operations on 64 bit variables and those who don't. For any single-core platform without atomics, I believe disabling interrupts should be enough. Besides the Pico I haven't looked much at embedded multicore platforms and I think we will handle them as they come. Maybe we will find a better way of structuring our atomics API with more examples.

core/platform/CMakeLists.txt
erlingrj: You are right. I could add both files to the build and use macros. Or do `#include "lf_atomic32.c` (or 64) from the different platform implementations.

Personally I would prefer to just stick to CMake. But I understand that there might be use-cases down the line where CMake is not an option... so yeah.

edwardalee:
Digging a bit more, I'm not sure moving everything to `CLOCK_REALTIME` is the right thing across the board.  Suppose you have a realtime task running every 10 ms and the NTP process drops in an increments `CLOCK_REALTIME` by 100 ms.  Your controller will effectively miss 10 cycles.  Suppose the NTP process decrements `CLOCK_REALTIME` by 100 ms.  Then, even with our control for monotonicity, your controller will execute 10 iterations in rapid succession.

Could we instead devise a target property that specifies a clocking policy?  This could even be developed into something that supports a 32-bit time measure, e.g. for Arduino.  The man page on Mac for `clock_gettime` has the following explanation:

```
     CLOCK_REALTIME     the system's real time (i.e. wall time) clock, expressed as the amount of time since the Epoch.  This is the same as the
                        value returned by gettimeofday(2).

     CLOCK_MONOTONIC    clock that increments monotonically, tracking the time since an arbitrary point, and will continue to increment while the
                        system is asleep.

     CLOCK_MONOTONIC_RAW
                        clock that increments monotonically, tracking the time since an arbitrary point like CLOCK_MONOTONIC.  However, this
                        clock is unaffected by frequency or time adjustments.  It should not be compared to other system time sources.

     CLOCK_MONOTONIC_RAW_APPROX
                        like CLOCK_MONOTONIC_RAW, but reads a value cached by the system at context switch.  This can be read faster, but at a
                        loss of accuracy as it may return values that are milliseconds old.

     CLOCK_UPTIME_RAW   clock that increments monotonically, in the same manner as CLOCK_MONOTONIC_RAW, but that does not increment while the
                        system is asleep.  The returned value is identical to the result of mach_absolute_time() after the appropriate
                        mach_timebase conversion is applied.

     CLOCK_UPTIME_RAW_APPROX
                        like CLOCK_UPTIME_RAW, but reads a value cached by the system at context switch.  This can be read faster, but at a loss
                        of accuracy as it may return values that are milliseconds old.

     CLOCK_PROCESS_CPUTIME_ID
                        clock that tracks the amount of CPU (in user- or kernel-mode) used by the calling process.

     CLOCK_THREAD_CPUTIME_ID
                        clock that tracks the amount of CPU (in user- or kernel-mode) used by the calling thread.
```

core/tag.c
```diff
@@ -177,22 +112,37 @@ interval_t lf_time_logical_elapsed(void *env) {
...
+    do {
+        // Atomically fetch the last read value. This is done with
+        // atomics to guarantee that it works on 32bit platforms as well.
+        last_read_local = lf_atomic_fetch_add64(&last_read_physical_time, 0);
+    
+
+        // Ensure monotonicity. Remeber that last_read_local is actually the 
+        if (now <= last_read_local) {
+            now = last_read_local+1;
+        }
```
edwardalee, line 129:
```suggestion
        last_read_local = lf_atomic_fetch_add64(&last_read_physical_time, 1);
    

        // Ensure monotonicity. Remeber that last_read_local is actually the 
        if (now < last_read_local) {
            now = last_read_local;
        }
```

include/core/platform/lf_atomic.h
```diff
@@ -0,0 +1,16 @@
...
+#include <stdbool.h>
+
+int32_t lf_atomic_fetch_add32(int32_t * ptr, int32_t val);
+int64_t lf_atomic_fetch_add64(int64_t * ptr, int64_t val);
+int32_t lf_atomic_add_fetch32(int32_t * ptr, int32_t val);
+int64_t lf_atomic_add_fetch64(int64_t * ptr, int64_t val);
+bool lf_atomic_bool_compare_and_swap32(int32_t* ptr, int32_t value, int32_t newval);
+bool lf_atomic_bool_compare_and_swap64(int64_t* ptr, int64_t value, int64_t newval);
+int32_t lf_atomic_val_compare_and_swap32(int32_t *ptr, int32_t val, int32_t newval);
+int64_t lf_atomic_val_compare_and_swap64(int64_t *ptr, int64_t val, int64_t newval);
```
edwardalee, line 14: These functions all need descriptions.  I think that when constructing the descriptions, it may become clear whether they are correctly named or not.  As it stands, I have to look at the implementation to see what they mean. In particular, `lf_atomic_bool_compare_and_swap...` seem particularly strange because there is no `bool` involved.  What is compared?  What is swapped?  How is this different from `lf_atomic_val_compare_and_swap...`?

edwardalee (commented):
Left a couple more questions/comments.

include/core/platform/lf_atomic.h
erlingrj: Will add docs. They are modeled after the GCC atomics described here: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html

bool vs val is the return type from the compare and swap

erlingrj:
> Digging a bit more, I'm not sure moving everything to `CLOCK_REALTIME` is the right thing across the board. Suppose you have a realtime task running every 10 ms and the NTP process drops in an increments `CLOCK_REALTIME` by 100 ms. Your controller will effectively miss 10 cycles. Suppose the NTP process decrements `CLOCK_REALTIME` by 100 ms. Then, even with our control for monotonicity, your controller will execute 10 iterations in rapid succession.
> 
> Could we instead devise a target property that specifies a clocking policy? This could even be developed into something that supports a 32-bit time measure, e.g. for Arduino. The man page on Mac for `clock_gettime` has the following explanation:
> 
> ```
>      CLOCK_REALTIME     the system's real time (i.e. wall time) clock, expressed as the amount of time since the Epoch.  This is the same as the
>                         value returned by gettimeofday(2).
> 
>      CLOCK_MONOTONIC    clock that increments monotonically, tracking the time since an arbitrary point, and will continue to increment while the
>                         system is asleep.
> 
>      CLOCK_MONOTONIC_RAW
>                         clock that increments monotonically, tracking the time since an arbitrary point like CLOCK_MONOTONIC.  However, this
>                         clock is unaffected by frequency or time adjustments.  It should not be compared to other system time sources.
> 
>      CLOCK_MONOTONIC_RAW_APPROX
>                         like CLOCK_MONOTONIC_RAW, but reads a value cached by the system at context switch.  This can be read faster, but at a
>                         loss of accuracy as it may return values that are milliseconds old.
> 
>      CLOCK_UPTIME_RAW   clock that increments monotonically, in the same manner as CLOCK_MONOTONIC_RAW, but that does not increment while the
>                         system is asleep.  The returned value is identical to the result of mach_absolute_time() after the appropriate
>                         mach_timebase conversion is applied.
> 
>      CLOCK_UPTIME_RAW_APPROX
>                         like CLOCK_UPTIME_RAW, but reads a value cached by the system at context switch.  This can be read faster, but at a loss
>                         of accuracy as it may return values that are milliseconds old.
> 
>      CLOCK_PROCESS_CPUTIME_ID
>                         clock that tracks the amount of CPU (in user- or kernel-mode) used by the calling process.
> 
>      CLOCK_THREAD_CPUTIME_ID
>                         clock that tracks the amount of CPU (in user- or kernel-mode) used by the calling thread.
> ```

The scenario you describe would not be acceptable for a real-time system. There are two ways of seeing that situation. Either NTP is configured incorrectly or the LF runtime is implemented incorrectly. I would argue that it is perfectly reasonable to put some requirements on how the behavior of the system clock. For Linux `chronyd` is the recommended NTP client and it has a rich set of options (https://chrony-project.org/doc/3.4/chrony.conf.html). The default behaviour is that the clock is never stepped, but you can configure it such that it performs a step if the offset is greater than a threshold. You can also say that is only allowed to step when the NTP client is starting up the first time to correct any initial offset. You can also control the maximum slew rate it can apply the the clock. The same is true for `linuxptp`.

So, I would argue that this is more of a philosophical question. Can we, and should we, put any requirements on the system on which we are running? After reading up on NTP clients I am even tempted to say that we shouldn't enforce monotonicity because we can require that CLOCK_MONTONIC is monotonic, which means that NTP and PTP must be configured correctly.

soyerefsane:
I was looking into issue #332 under Raspberry Pi 4b, and Edward thought they might be worthwhile to share it here as well.

There are 2 main differences between Raspberry Pi and MacOS call stack that causes the different behavior we are observing in terms of timing. (1) While MacOS uses CLOCK_MONOTONIC, raspberry pi is actually using a CLOCK_REALTIME. (2) MacOS uses the `lf_POSIX_threads_support.c` for `lf_cond_timedwait` function call, which calls `pthread_cond_timedwait` function call underneath. Raspberry Pi is using `lf_C11_threads_support.c` which calls `cnd_timedwait` as the underlying function call. 

I observe during the timer triggers, in many cases the Raspberry Pi implementation produces negative lag, although this negative lag is usually less than 10 microseconds, by inspection of the eye. I also noticed that some runs, although not frequent, generate lags that are around 400 microseconds. I'm working on generating a histogram with different runs of the system to see what the delay pattern looks like. 

Upon inspecting Edward and I came across this [Stackoverflow post](https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic) that seems relevant to this issue. 

Overall, I think I agree with the concerns that Edward mentioned above regarding the switch to CLOCK_REALTIME. And two different clocks have their own set of advantages and disadvantages. In most cases, we are more interested in the accuracy of the periods between the triggers and not necessarily the real physical time that is observed outside of the system. However, at the same time, there are some systems that are by design required to work with a real-time clock. I was looking more into this issue to take it as an agenda item for Thursday's Timing Precision meeting as well.

erlingrj:
@edwardalee I mark this as ready for review. I think a quite important and subtle point here is that this PR moves the responsibility of translating back and forth between the physical clock and the synchronized physical clock introduced by the clock synchronization that we support. This means that whenever you do `_lf_clock_now` (which is what `lf_time_physical` you get an `instant_t` that is corrected for clock synchronization. So all `instant_t`s in the runtime should be corrected for any clock sync. This means that the platform implementations must do two things:
1. Apply any clock synchronization offset to the raw time read from their physical clock.
2. Remove clock synchronization offsets if they need a raw absolute time to schedule a future wakeup.

This is a little subtle because some platforms do sleeping by calculating a sleep duration and then calling some lower-level API. In that case you should NOT subtract any clock sync offset since you will compute the duration relative `_lf_clock_now` which includes the clock-sync offset. So this does introduce some new challenges for implementing a new platform. And it is quite easy to make a mistake, I had made several mistakes related to applying/removing clock sync offsets between the raw physical clock and the `instant_t`s...

There is a way to avoid this, which is to add a layer between the platform API and the user-facing code. For `_lf_clock_now` we already do this, `lf_time_physical` is exactly this layer, and we could apply clock-sync offsets there. However for `lf_cond_timedwait` and `lf_sleep_until` we do not have such a layer. It would only affect the platform API which deals with time....

core/platform/lf_rp2040_support.c
```diff
@@ -92,6 +92,7 @@ int _lf_clock_now(instant_t* t) {
     now = get_absolute_time();
     ns_from_boot = to_us_since_boot(now) * 1000;
     *t = (instant_t) ns_from_boot;
+    clock_sync_apply_offset(t);
```
edwardalee: Need to add to the function comment that the clock sync offset is added.

core/platform/lf_rp2040_support.c
```diff
@@ -203,6 +198,16 @@ int lf_enable_interrupts_nested() {
     return 0;
 }
 
+#if defined(LF_SINGLE_THREADED)
+/**
+ * The single thread RP2040 platform support treats second core
+ * routines similar to external interrupt routine threads.
+ * 
+ * Second core activity is disabled at the same times as
+ * when interrupts are disabled.
```
edwardalee, line 203: What is this comment associated with?  Should this be a file-level comment at the top of the file?

core/platform/lf_unix_clock_support.c
```diff
@@ -28,69 +18,31 @@ struct timespec convert_ns_to_timespec(instant_t t) {
...
 }
 
 /**
- * Fetch the value of _LF_CLOCK (see lf_linux_support.h) and store it in tp. The
- * timestamp value in 't' will always be epoch time, which is the number of
- * nanoseconds since January 1st, 1970.
- *
- * @return 0 for success, or -1 for failure. In case of failure, errno will be
- *  set appropriately (see `man 2 clock_gettime`).
+ * Fetch the value of CLOCK_REALTIME and store it in t. The
```
edwardalee, line 32:
```suggestion
 * Fetch the value of CLOCK_REALTIME, add the clock sync offset to it, and store the result in t.
```

core/platform/lf_windows_support.c
```diff
@@ -107,6 +104,7 @@ int _lf_clock_now(instant_t* t) {
         windows_time.QuadPart |= f.dwLowDateTime;
     }
     *t = (instant_t)((double)windows_time.QuadPart / _lf_frequency_to_ns);
+    clock_sync_apply_offset(t);
```
edwardalee: Need to add to the function comment that the clock sync offset is added.

core/platform/lf_windows_support.c
```diff
@@ -147,7 +145,7 @@ int lf_sleep(interval_t sleep_duration) {
 int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_time) {
     interval_t sleep_duration = wakeup_time - lf_time_physical();
 
-    if (sleep_duration < LF_MIN_SLEEP_NS) {
+    if (sleep_duration < 0) {
```
edwardalee, line 147:
```suggestion
    if (sleep_duration <= 0) {
```

core/platform/lf_zephyr_clock_counter.c
```diff
@@ -142,6 +142,8 @@ int _lf_clock_now(instant_t* t) {
     }
 
     *t = now_nsec;
+    clock_sync_apply_offset(t);
```
edwardalee: Need to add to the function comment that the clock sync offset is added. Actually, this function comment seems incomplete overall.

core/platform/lf_zephyr_clock_kernel.c
```diff
@@ -65,6 +65,7 @@ int _lf_clock_now(instant_t* t) {
         last_epoch_nsec += epoch_duration_nsec;
     }
     *t = (SECOND(1)/CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC)*now_cycles + last_epoch_nsec;
+    clock_sync_apply_offset(t);
```
edwardalee: Incomplete/unclear function comment.  Need to add to the function comment that the clock sync offset is added.

include/core/platform.h
```diff
@@ -202,85 +201,14 @@ int lf_cond_wait(lf_cond_t* cond);
 
 /**
  * Block current thread on the condition variable until condition variable
- * pointed by "cond" is signaled or time pointed by "absolute_time_ns" in
- * nanoseconds is reached.
+ * pointed by "cond" is signaled or time given by wakeup_time is reached. The
+ * wakeup time will include any clock sync offset (which is added in
+ * _lf_clock_now).
```
edwardalee:
```suggestion
 * wakeup time is assumed to include any clock sync offset (which is automatically added in
 * _lf_clock_now and hence also in lf_time_physical).
```

include/core/platform.h
```diff
@@ -307,9 +235,11 @@ int _lf_clock_now(instant_t* t);
...
 
 /**
- * @brief Sleep until the given wakeup time. Assumes the lock for the
- * given environment is held
- * 
+ * @brief Sleep until the given wakeup time. The wakeup time will include any
+ * clock sync offsets (which are added in _lf_clocK_now).
+ *
+ * Assumes the lock for the given environment is held
+ *
```
edwardalee:
```suggestion
 * @brief Sleep until the given wakeup time.
 
 * The wakeup time is assumed to include any clock sync offsets (which are added in _lf_clocK_now and
 * hence also by lf_time_physical).
 *
 * This assumes the lock for the given environment is held.
 *
```

include/core/platform/lf_atomic.h
```diff
@@ -0,0 +1,103 @@
...
+ */
+#ifndef LF_ATOMICS_H
+#define LF_ATOMICS_H
+
+#include <stdint.h>
+#include <stdbool.h>
+
+/**
+ * @brief Atomically fetch a 32bit integer from memory and add a value to it.
+ * Returns the value that was previously in memory.
```
edwardalee, line 15:
```suggestion
 * Return the value that was previously in memory.
```

include/core/platform/lf_atomic.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param ptr A pointer to the memory location.
+ * @param val The value to be added.
+ * @return The value previously in memory.
+ */
+int32_t lf_atomic_fetch_add32(int32_t * ptr, int32_t val);
+
+/**
+ * @brief Atomically fetch 64-bit integer from memory and add a value to it.
+ * Returns the value that was previously in memory.
```
edwardalee, line 25:
```suggestion
 * Return the value that was previously in memory.
```

include/core/platform/lf_atomic.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param ptr A pointer to the memory location.
+ * @param val The value to be added.
+ * @return The value previously in memory.
+ */
+int64_t lf_atomic_fetch_add64(int64_t * ptr, int64_t val);
+
+/**
+ * @brief Atomically fetch 32-bit integer from memory and add a value to it.
+ * Returns the new value of the memory.
```
edwardalee, line 35:
```suggestion
 * @brief Atomically fetch a 32-bit integer from memory and add a value to it.
 * Return the new value of the memory.
```

include/core/platform/lf_atomic.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param ptr A pointer to the memory location.
+ * @param val The value to be added.
+ * @return The new value in memory.
+ */
+int32_t lf_atomic_add_fetch32(int32_t * ptr, int32_t val);
+
+/**
+ * @brief Atomically fetch 64-bit integer from memory and add a value to it.
+ * Returns the new value of the memory.
```
edwardalee, line 45:
```suggestion
 * @brief Atomically fetch a 64-bit integer from memory and add a value to it.
 * Return the new value of the memory.
```

include/core/tag.h
```diff
@@ -75,6 +75,21 @@ typedef struct {
...
+#if defined(_LF_CLOCK_SYNC_ON)
+extern void clock_sync_apply_offset(instant_t *t);
+extern void clock_sync_remove_offset(instant_t *t);
+extern void clock_sync_set_constant_bias(interval_t offset);
+#else
+// Without clock synchronization enabled we can just optimize these away.
+#define clock_sync_apply_offset(x)
+#define clock_sync_remove_offset(x)
+#define clock_sync_set_constant_bias(x)
+#endif
```
edwardalee: I don't see why we wouldn't allow these functions even if clock sync is not on.  Could allow the user to adjust the clock sync.

edwardalee (approved):
This looks excellent to me.  I've added a few small suggestions.

erlingrj:
Thanks for the thorough review @edwardalee. I am sorry that I have made some additional substantial changes which I need some approval for. As I discussed in the meeting I introduced now a layer between the timing APIs of `platform.h` and the rest of the runtime. I called it `clock.h` and it fetches physical clock readings from the platform API and adds/removes clock sync offsets and also ensures monotonicity. 

I think this is an improvement and makes thing more clear.

We can definitely support dropping in your own clock sync protocol at some point. For now I had to hide it with macros because `clock-sync.h` is tied with threaded platform and we cannot include it with an unthreaded platform. The user could also get his own clock sync by directly modifying the physical clock in an async thread also...

include/core/tag.h
erlingrj: See final comment

include/core/federated/clock-sync.h
```diff
@@ -198,4 +198,33 @@ void* listen_to_rti_UDP_thread(void* args);
  */
 int create_clock_sync_thread(lf_thread_t* thread_id);
 
+
+// If clock sync is enabled. Then expose an API for applying and remove the
+// offset in a thread safe manner.
+#if defined(_LF_CLOCK_SYNC_ON)
```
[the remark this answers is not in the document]
erlingrj: See final comment

include/core/clock.h
```diff
@@ -0,0 +1,50 @@
...
+
+ * @return 0 on success or -1 if interrupted.
+ */
+int lf_clock_interruptable_sleep_until_locked(environment_t *env, instant_t wakeup_time);
+
+
+/**
+ * Retrives the current physical time from the platform API. It adjusts for
+ * clock synchronization and guaranteed monotonicity.
+ * @param now
```
edwardalee, line 32:
```suggestion
 * Retrieve the current physical time from the platform API. This adds any clock synchronization offset
 * and guarantees monotonicity. Specifically, each returned value will be at least one microsecond larger
 * than any previously returned time.
 * @param now A pointer to the location in which to store the result.
```

include/core/clock.h
```diff
@@ -0,0 +1,50 @@
...
+ * @return 0 on success or -1 if interrupted.
+ */
+int lf_clock_interruptable_sleep_until_locked(environment_t *env, instant_t wakeup_time);
+
+
+/**
+ * Retrives the current physical time from the platform API. It adjusts for
+ * clock synchronization and guaranteed monotonicity.
+ * @param now 
+ * @return 0 on success.
```
edwardalee, line 33:
```suggestion
 * @return 0 on success, -1 on failure to read the platform clock.
```

include/core/clock.h
```diff
@@ -0,0 +1,50 @@
...
+ * @return 0 on success. 
+ */
+int lf_clock_gettime(instant_t *now);
+
+#if !defined(LF_SINGLE_THREADED)
+/**
+ * This will block the calling thread on the condition variable until it is
+ * signaled or until wakeup_time is reached. Before calling the appropriate
+ * function in the platform API, the wakeup_time will be translated into the
+ * correct timescale by removing any clock synchronization offset.
```
edwardalee, line 42:
```suggestion
 * Block the calling thread on the condition variable until it is
 * signaled or until wakeup_time is reached. Before calling the appropriate
 * function in the platform API, the wakeup_time will be translated into the
 * correct timescale by removing any clock synchronization offset.
```

include/core/platform.h
```diff
@@ -202,85 +201,14 @@ int lf_cond_wait(lf_cond_t* cond);
 
 /**
  * Block current thread on the condition variable until condition variable
```
edwardalee, line 203:
```suggestion
 * Block the current thread on the condition variable until the condition variable
```

include/core/platform.h
```diff
@@ -202,85 +201,14 @@ int lf_cond_wait(lf_cond_t* cond);
...
- * pointed by "cond" is signaled or time pointed by "absolute_time_ns" in
- * nanoseconds is reached.
+ * pointed by "cond" is signaled or time given by wakeup_time is reached. Should
+ * not be used directly as it does not account for clock synchronization offsets.
+ * Use `lf_clock_cond_timedwait` from clock.h instead.
```
edwardalee, line 206:
```suggestion
 * pointed by "cond" is signaled or the time given by wakeup_time is reached. This should
 * not be used directly as it does not account for clock synchronization offsets.
 * Use `lf_clock_cond_timedwait` from clock.h instead.
```

include/core/platform.h
```diff
@@ -289,15 +217,16 @@ int lf_cond_timedwait(lf_cond_t* cond, instant_t absolute_time_ns);
...
- * Fetch the value of an internal (and platform-specific) physical clock and
- * store it in `t`.
- *
- * Ideally, the underlying platform clock should be monotonic. However, the
- * core lib tries to enforce monotonicity at higher level APIs (see tag.h).
+ * Fetch the value of an internal (and platform-specific) physical clock.
+ * Ideally, the underlying platform clock should be monotonic. However, the core
+ * lib enforces monotonicity at higher level APIs (see clock.h).
+ * 
+ * Should not be used directly as it does not apply clock synchronization
```
edwardalee, line 224:
```suggestion
 * This should not be used directly as it does not apply clock synchronization
```

include/core/platform.h
```diff
@@ -307,9 +236,11 @@ int _lf_clock_now(instant_t* t);
 int lf_sleep(interval_t sleep_duration);
 
 /**
- * @brief Sleep until the given wakeup time. Assumes the lock for the
- * given environment is held
- * 
+ * @brief Sleep until the given wakeup time. Should not be used directly as it
+ * does not account for clock synchronization offsets.
+ *
+ * Assumes the lock for the given environment is held
```
edwardalee, line 242:
```suggestion
 * @brief Sleep until the given wakeup time. This should not be used directly as it
 * does not account for clock synchronization offsets. See clock.h.
 *
 * This assumes the lock for the given environment is held.
```

core/tag.c
```diff
@@ -177,22 +111,22 @@ interval_t lf_time_logical_elapsed(void *env) {
     return lf_time_logical(env) - start_time;
 }
 
-instant_t lf_time_physical(void) {
-    return _lf_physical_time();
+instant_t lf_time_physical() {
```
edwardalee, line 114:
```suggestion
instant_t lf_time_physical(void) {
```

edwardalee (approved):
This turned into a pretty big effort!  It looks great to me. I've appended a few more small suggestions.

edwardalee:
I talked to Kevin Stanton, who recently retired from Intel after leading for many years their PTP efforts.  He told me that some Intel platforms are already enforcing monotonicity for CLOCK_REALTIME, and that it is providing the PTP time.  He says he thinks this is the trend and that all Linux systems will eventually do this.  This reinforces that this PR is really the right approach for the long term.

erlingrj:
> I talked to Kevin Stanton, who recently retired from Intel after leading for many years their PTP efforts. He told me that some Intel platforms are already enforcing monotonicity for CLOCK_REALTIME, and that it is providing the PTP time. He says he thinks this is the trend and that all Linux systems will eventually do this. This reinforces that this PR is really the right approach for the long term.

Great to hear! Would be quite interesting to hear more from him on how we can make the LF runtime as relevant as possible for the applications they were targeting at his group at Intel. Also what hardware we should target for our future PTP/TSN testbeds.

PR #348, opened 2024-02-04
First step towards versioned federation protocol

edwardalee:
This PR adds a new signal type `MSG_TYPE_FED_VERSION` to the federated protocol and deprecates `MSG_TYPE_FED_IDS`.  This signal is sent by a federate to the RTI when it first joins the federation.  The signal `MSG_TYPE_FED_VERSION` adds three bytes to indicate major, minor, and patch version numbers for the protocol.  The current version number in this PR is `0.1.0`.  If the `MSG_TYPE_FED_IDS` signal is used by the federates joining the federation, then the RTI will interpret this as version `0.0.0` and will operate in a backward compatible way.  This means that you can install the RTI from `main` or from the nightly build, and it will still work if you compile federates using the latest release instead of `main` or the nightly build.  Thus, one installation of the RTI can be used for development and for demos/etc. with the latest release.

### DIFF
--- a/core/federated/RTI/rti_remote.c
+++ b/core/federated/RTI/rti_remote.c
@@ -1211,7 +1211,12 @@ void *federate_info_thread_TCP(void *fed) {
     return NULL;
 }
 
-void send_reject(int *socket_id, unsigned char error_code) {
+/**
+ * Send a MSG_TYPE_REJECT message to the specified socket and close the socket.
+ * @param socket_id Pointer to the socket ID.
+ * @param error_code An error code.
+ */
+static void send_reject(int *socket_id, unsigned char error_code) {
     LF_PRINT_DEBUG("RTI sending MSG_TYPE_REJECT.");
     unsigned char response[2];
     response[0] = MSG_TYPE_REJECT;
@@ -1239,11 +1244,11 @@ void send_reject(int *socket_id, unsigned char error_code) {
  */
 static int32_t receive_and_check_fed_id_message(int *socket_id, struct sockaddr_in *client_fd) {
     // Buffer for message ID, federate ID, and federation ID length.
-    size_t length = 1 + sizeof(uint16_t) + 1; // Message ID, federate ID, length of fedration ID.
+    size_t length = 4 + sizeof(uint16_t) + 1; // Message ID, version, federate ID, length of federation ID.
     unsigned char buffer[length];
 
-    // Read bytes from the socket. We need 4 bytes.
-    if (read_from_socket_close_on_error(socket_id, length, buffer)) {
+    // Read the first byte from the socket.
+    if (read_from_socket_close_on_error(socket_id, 1, buffer)) {
         lf_print_error("RTI failed to read from accepted socket.");
         return -1;
     }
@@ -1251,15 +1256,14 @@ static int32_t receive_and_check_fed_id_message(int *socket_id, struct sockaddr_
     uint16_t fed_id = rti_remote->base.number_of_scheduling_nodes; // Initialize to an invalid value.
 
     // First byte received is the message type.
-    if (buffer[0] != MSG_TYPE_FED_IDS) {
+    if (buffer[0] != MSG_TYPE_FED_IDS && buffer[0] != MSG_TYPE_FED_VERSION) {
         if (rti_remote->base.tracing_enabled) {
             tracepoint_rti_to_federate(rti_remote->base.trace, send_REJECT, fed_id, NULL);
         }
         if (buffer[0] == MSG_TYPE_P2P_SENDING_FED_ID || buffer[0] == MSG_TYPE_P2P_TAGGED_MESSAGE) {
             // The federate is trying to connect to a peer, not to the RTI.
             // It has connected to the RTI instead.
-            // FIXME: This should not happen, but apparently has been observed.
-            // It should not happen because the peers get the port and IP address
+            // NOTE: This should not happen because the peers get the port and IP address
             // of the peer they want to connect to from the RTI.
             // If the connection is a peer-to-peer connection between two
             // federates, reject the connection with the WRONG_SERVER error.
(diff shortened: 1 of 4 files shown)


lhstrh:
It seems to me that we're versioning the protocol, not the implementation. That makes sense, particularly if there are multiple implementations (are there?) but then the protocol has to be specified somewhere and be versioned. Without a versioned specification of the protocol, the numbers won't mean much.

While versioning of the protocol is useful, we also need to version the implementation. A well-designed protocol doesn't change much over time, but implementations do because they include bug fixes. It's important to know which version of the implementation people are using when they file bug reports, for instance.

core/federated/RTI/rti_remote.c
```diff
@@ -1287,7 +1309,36 @@ static int32_t receive_and_check_fed_id_message(int *socket_id, struct sockaddr_
...
+            }
+            rti_remote->protocol_version = 0;
+            LF_PRINT_DEBUG("RTI: Legacy protocol version 0 will be used.");
+        } else {
+            // Extract the proposed protocol version.
+            int32_t proposed_version = 
+                    ((int32_t)buffer[1]) << 16 
+                    | ((int32_t)buffer[2]) << 8 
+                    | ((int32_t)buffer[3]);
+            if (rti_remote->protocol_version > 0 && proposed_version != rti_remote->protocol_version) {
```
lhstrh, line 1329: It's good that we use semantic versions, but then we shouldn't require exact matches. According to semver, any minor or patch increments should not be include breaking changes. We haven't done a good job with this in the `lingua-franca` repo, but we should really try to adhere to the semver rules here...

petervdonovan:
This is still marked as a draft, so I will just leave high-level comments instead of a detailed review:
- I am not sure that now is the right time to start worrying about supporting old versions of the protocol. If we tell anyone that the RTI supports multiple versions of the protocol, then this will lead to its code becoming more complex, and in order to keep any such promise we will have to run more tests (I am taking for granted that "best effort" is not an acceptable standard for backwards compatibility in this case). The tests that we use are already both expensive and insufficient to guarantee a high standard of correctness even for just one version of the protocol.
    - Although it may be easy to support an older protocol version now, I do not like the precedent that this sets because in the future, when it becomes harder to support the latest two protocol versions, I don't want anyone to feel like they are "losing" anything by choosing to make the RTI compatible only with the latest one.
- I agree that it is reasonable to start versioning the protocol, however. The easiest thing is to just make any protocol mismatch a fatal error.
- (And as a less important point: I don't think there is any need to skimp on bytes for this. It is highly unlikely that any part of the version number will ever be larger than 255, but for a message that is sent only once, there is no harm at all in sending (say) 12 bytes instead of 3.)

include/core/federated/network/net_common.h
```diff
@@ -179,6 +155,26 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * The first federate to join may either not propose a version number (in which case
+ * the RTI will use version 0). If the RTI does not support the version proposed by the
+ * first federate, then it will send MSG_TYPE_REJECT. Otherwise, it sends MSG_TYPE_ACK.
+ * Subsequent federates will have to propose matching versions or else they too will be
+ * rejected.  If the federate itself supports multiple versions, then it can attempt to
+ * join again with a lower proposed version number.  A federate uses the MSG_TYPE_FED_VERSION
+ * to propose a protocol version (this message will also carry the federation ID).
+ * A legacy federate using version 0 attempts to join with MSG_TYPE_FED_IDS instead.
+ */
+#define LF_FED_VERSION 0x000100
```
lhstrh, line 176: Do we really want to encode it this way? If we want to create a script that automatically updates the version number in the code base, that script will have to translate a semver string into a hex code... We need to store the string somewhere anyway because we need to print it when the `-version` flag is used.

edwardalee: There is a tradeoff here.  Either store the binary data that will be transmitted and compared, store a string that has to be parsed at runtime, or store both and risk them being different.  Are there any other alternatives?  I opted for the first option, but don't mind if you want to change it.

lhstrh: I think we should parse the string at runtime if you prefer to use a binary format for the communication. I think it would be bad to leak implementation details about the implementation into other tools and have them take care of something as low-level as turning a human-readable string into a binary format. It also obfuscates the source code. My point is that in order to print the version number, you need the the human-readable string, anyway, so you might as well store it as such.

edwardalee: OK, we can do that. However, I don't really get this argument. Synthesizing a string from binary data is pretty easy using printf. So is parsing a string using scanf.

lhstrh: It would be nice if 1) the define in the code was human readable, so that we can look at it and see instantly what version it denotes; 2) be able to use a `sed` command to update version numbers in the code base without having to convert the desired version number to binary. Neither are true if the define is a binary hex code.

edwardalee: I'm still not convinced.  This version number will have to be sent over the network and compared, so it will be needed in binary form.  So if we give it as a string, then in C code, we will have to parse the string during startup and store the binary form.  It will have to have special parsing logic to handle "alpha", "beta", "snapshot", etc., whereas in binary form these could just be encoded in the final byte.  For the sed script to update the version number, is it more difficult to replace "#define LF_REACTOR_C_VERSION 0x000600" (for 0.6.0) with "#define LF_REACTOR_C_VERSION 0x000700" than to replace strings?

Also, there is no command-line interface for reactor-c, so it's not clear that this version number will ever have to converted to a string.  One of my longer-term goals is to have a pico version of reactor-c that uses no printf (or scanf) anywhere, since these are quite expensive in program memory.

lhstrh: We were discussing versioning of the rti and now we're talking about a specific variant of reactor-c for tiny embedded system. I'm a bit lost...

What I think would be reasonable requirements are:
 - a human-readable version number in the source code
 - ability to have an simple script that uses `sed` to update the version number based on a string argument, which we'll use with every release
 - ability to run **the rti** with a `--version` flag to print the version number

What I'm saying nothing about is:
 - what restrictions to impose on the version number (e.g., dot-separated numerals only)
 - how to send the version number across the network
 - how to extract a version number from a program that uses `reactor-c`
 - anything about trimmed-down runtime implementations that won't print anything in the first place

If you don't think the stated requirements are reasonable, please let me know why you think they aren't.

edwardalee: You are right.  This needs more thought.  I'm going to table this for now.  Other things have higher priority.

lhstrh: > This needs more thought.

Does it, really? I would just stick a `#define VERSION "0.6.0"` in `rti_common.h` and add a function in `rti_remote.c` to turn it into a binary number before using it in the protocol. Seems simple. Am I missing something?

elgeeko1: I've seen CMake generate a `version.h` headerfile that can be included by any target, defining such a `VERSION` macro. This way CMake can intelligently produce major or minor revisions and targets can use as they wish.

edwardalee: I've shelved this PR for now because I have higher priority items to work on.  Happy to yield if someone else can take over.

edwardalee:
I am closing this PR because it's not at all clear that this is the right approach.  There could be major advantages to compiling the RTI together with the federates and sticking in the bin directory with the federate binaries. This will ensure compatibility and enable more significant optimizations, including creating more decentralized versions of the RTI functionality.  That approach comes at price, however, in that the build system for other targets that use the same RTI code will have include reactor-c as a submodule and augment their build system to also build the RTI.

Issue #264, opened 2023-08-22
Optimizing Centralized Federated Execution


byeong-gil:
This issue is inherited from [this discussion](https://github.com/lf-lang/lingua-franca/discussions/1626), mostly from @edwardalee's description. I'm writing this issue to try to make a more readable document as that discussion is very long. 

# Motivation
<img width="592" alt="image" src="https://user-images.githubusercontent.com/8513334/223158587-664148b6-eeb6-4ce4-a5d3-41df5961f24e.png">
Suppose that the Sender, which triggers at 100ms intervals, only occasionally sends an output message, say, on average, every few seconds. Currently, Sender sends LTC, ABS, and NET messages every 100 msec, even with no interesting information. In the below visualization (from @ChadliaJerad), the RTI is on the left, the Sender in the middle, and the Receiver on the right. Every message is redundant because Receiver has nothing to do with those messages (its next event tag is 2 sec, which is the timeout value). So if we can eliminate those messages, network overhead can be reduced.

![diagram](https://github.com/lf-lang/reactor-c/assets/78055940/29e65070-9282-453e-a6d7-e8c6a7bd3c83)

# Solution
A new message type, **Next Downstream Event Tag (NDET)** is proposed to resolve this inefficiency. When the RTI receives a `NET` from a downstream federate, it should forward that message to all upstream federates of the downstream. Those federates should maintain a _barrier_ tag _b_ that is the least such tag they have received (they will need a queue of such tags because the least will drop off when the federate sends an `LTC` matching or exceeding the tag).

### RTI Side
* When the RTI receives a NET (Td), it should forward NET using NDET to those upstream federates. 
Condition: Assuming that a downstream federate D sends NET (Td) and it has n number of upstream federate U<sub>1</sub>, U<sub>2</sub>…U<sub>n</sub>. Each upstream federate’s last LTC is T<sub>U1</sub>, T<sub>U2</sub>, … and so on. Then, an upstream federate Uk (1≤k≤n) should receive NDET(Td) if T<sub>Uk</sub> < Td.

### Federate Side
* When an upstream federate receives an `NDET(Td)`, it should
1. Store the tag `Td`. (If I remember correctly, @lhstrh suggested that an array with a length of the number of its downstream federates is enough to store those tags.)
3. If `Td` <= Current Tag, send a proper `NET` or `LTC` so that the RTI can give a grant to downstream federates.
4. Update the barrier tag b. It should be the minimum tag among the stored tags.

[...]

PR #349, opened 2024-02-05
Downstream next event tag (DNET),  a new signal for more efficient centralized federated execution

byeonggiljun:
This PR replaces https://github.com/lf-lang/reactor-c/pull/176 and https://github.com/lf-lang/reactor-c/pull/337. 

Relevant discussion and issue: https://github.com/lf-lang/lingua-franca/discussions/1626, https://github.com/lf-lang/reactor-c/issues/264

A companion PR in lingua-franca: https://github.com/lf-lang/lingua-franca/pull/2400
A companion PR in reactor-ts: https://github.com/lf-lang/reactor-ts/pull/296

In this PR, the new message type, downstream next event tag (DNET), is introduced to reduce unnecessary NET signals. This signal is sent from the RTI to a federate to notify it that NET tags below some threshold are unnecessary, and not needed by its downstream federates. 

Suppose the RTI is computing the DNET tag for a federate `A` where `A` has one downstream federate `B`. The RTI 
 uses the minimum delay between the federates and `B`'s NET value to compute the DNET tag. For example, suppose the delay from `A` to `B` is `NEVER` and `A` has an event every `10 ms` (`0`, `10 ms`, ...). Also, assume B's next event is at `(100 ms, 0)`. The RTI knows B's next earliest event tag by finding the minimum tag of the most recent NET_B and the head of the in-transit message queue of `B` is `(100 ms, 0)`. 

In this case, the RTI can send `DNET (100 ms, 0)` to `A`. Based on the DNET signal, if A is not producing outputs, A does not need to send any NET signals with tags earlier than or equal to (100 ms, 0) (NET with 10 ms, 20 ms, ..., 100 ms) because the RTI **cannot** send `TAG (100 ms, 0)` using those NET signals. The RTI can grant `TAG (100 ms, 0)` to `B` after receiving `NET (110 ms, 0)` from `A`.

Additionally, in this PR, the RTI computes the tag of the signal tag advance grant (TAG) with the earliest incoming message tag (EIMT) to prevent blocking a federate that is eligible to advance its tag. Currently, if a federate has its next event scheduled at `(100 ms, 0)` and its EIMT is `(200 ms, 1)`, the RTI sends `TAG (100 ms, 0)`. However, this incurs unnecessary NET signals if the federate has an event at the tag (200 ms, 0). So in this PR, the RTI sends `TAG (200 ms, 0)` (the latest earliest tag of EIMT) so that the federate can execute any potential events with tags earlier than `(200 ms, 1)` as well as the event at `(100 ms, 0)`.

This PR also contains some refactoring for the data structure used by the RTI to store delays between federates. A two-dimensional matrix stores the delays now.


core/federated/RTI/main.c
```diff
@@ -316,9 +319,14 @@ int main(int argc, const char* argv[]) {
   assert(rti.base.number_of_scheduling_nodes < UINT16_MAX);
 
   // Allocate memory for the federates
-  rti.base.scheduling_nodes =
-      (scheduling_node_t**)calloc(rti.base.number_of_scheduling_nodes, sizeof(scheduling_node_t*));
-  for (uint16_t i = 0; i < rti.base.number_of_scheduling_nodes; i++) {
+  int n = rti.base.number_of_scheduling_nodes;
+  rti.base.scheduling_nodes = (scheduling_node_t**)calloc(n, sizeof(scheduling_node_t*));
+  // Allocate memory for the array of min_delays.
+  rti.base.min_delays = (tag_t*)calloc((n * n), sizeof(tag_t));
```
edwardalee, line 325: Should this allocation occur only if `rti.base.dnet_enabled` is true?
The for loop below would also have to be made conditional.

core/federated/RTI/rti_common.c
```diff
@@ -18,6 +18,20 @@ static rti_common_t* rti_common = NULL;
 // Global variables defined in tag.c:
 extern instant_t start_time;
 
+static tag_t latest_earlier_tag(tag_t tag) {
+  if (tag.time == NEVER || tag.time == FOREVER) {
+    return tag;
+  } else if (tag.time == 0 && tag.microstep == 0) {
```
edwardalee, line 24: Is this condition really necessary?  Some platforms might use time == 0 for the start time (they may not have UTC clocks).  Even with UTC clocks, a tag of (0,0) is a legitimate tag.  The time refers to midnight on January 1, 1970.

core/federated/RTI/rti_common.c
```diff
@@ -18,6 +18,20 @@ static rti_common_t* rti_common = NULL;
 // Global variables defined in tag.c:
 extern instant_t start_time;
 
+static tag_t latest_earlier_tag(tag_t tag) {
+  if (tag.time == NEVER || tag.time == FOREVER) {
```
edwardalee, line 22: This is a nit, but strictly speaking, I think these should be using `lf_tag_compare` to compare with `NEVER_TAG` and `FOREVER_TAG`.

core/federated/RTI/rti_common.c
```diff
@@ -18,6 +18,20 @@ static rti_common_t* rti_common = NULL;
 // Global variables defined in tag.c:
 extern instant_t start_time;
 
+static tag_t latest_earlier_tag(tag_t tag) {
```
edwardalee, line 21: It seems like this function could go in tag.h/c, in which case it should be called `lf_tag_latest_earlier` and should be documented something like this:
```suggestion
/**
 * @brief Return the greatest tag earlier than the given tag.
 *
 * If the given tag is `FOREVER_TAG` or `NEVER_TAG`, however, just return the given tag.
 * @param tag The tag.
 */
static tag_t latest_earlier_tag(tag_t tag) {
```

core/federated/RTI/rti_common.c
```diff
@@ -34,10 +48,18 @@ void initialize_rti_common(rti_common_t* _rti_common) {
 #define IS_IN_CYCLE 2
 
 void invalidate_min_delays_upstream(scheduling_node_t* node) {
-  if (node->min_delays != NULL)
-    free(node->min_delays);
-  node->min_delays = NULL;
-  node->num_min_delays = 0;
+  if (node->all_upstreams != NULL) {
+    free(node->all_upstreams);
+    for (int i = 0; i < rti_common->number_of_scheduling_nodes; i++) {
```
edwardalee, line 53: Why is this for loop inside the test `node->all_upstreams != NULL`?  Maybe it should be inside a test for `min_delays` being NULL (particularly if you adopt my suggestion above).

core/federated/RTI/rti_common.h
```diff
@@ -53,17 +53,20 @@ typedef struct scheduling_node_t {
...
+                                         // Here, NEVER encodes no delay. 0LL is a microstep delay.
+  uint16_t num_immediate_upstreams;      // Size of the array of immediate upstream scheduling nodes and delays.
+  uint16_t* immediate_downstreams;       // Array of immediate downstream scheduling node ids.
+  uint16_t num_immediate_downstreams;    // Size of the array of immediate downstream scheduling nodes.
+  execution_mode_t mode;                 // FAST or REALTIME.
+  uint16_t* all_upstreams;               // Array of all upstream scheduling node ids.
+  uint16_t num_all_upstreams;            // Size of the array of all upstream scheduling nodes and delays.
+  uint16_t* all_downstreams;             // Array of all downstream scheduling node ids.
+  uint16_t num_all_downstreams;          // Size of the array of all downstream scheduling nodes.
+  int flags;                             // Or of IS_IN_ZERO_DELAY_CYCLE, IS_IN_CYCLE
```
edwardalee, line 69:
```suggestion
  int flags;                             // One of IS_IN_ZERO_DELAY_CYCLE, IS_IN_CYCLE
```

core/federated/RTI/rti_common.h
```diff
@@ -76,7 +79,12 @@ typedef struct rti_common_t {
   scheduling_node_t** scheduling_nodes;
 
   // Number of scheduling nodes
-  int32_t number_of_scheduling_nodes;
+  uint16_t number_of_scheduling_nodes;
+
+  // Matrix of minimum delays between each nodes
+  // Rows represent upstreams and Columns represent downstreams.
+  // FOREVER_TAG means there is no path, ZERO_TAG means there is no delay.
```
edwardalee, line 86:
```suggestion
  // Matrix of minimum delays between pairs of nodes.
  // Rows represent upstream nodes and Columns represent downstream nodes.
  // FOREVER_TAG means there is no path, and ZERO_TAG means there is no delay.
  // This could be NULL if the matrix is not being used, so accesses should test for NULL first.
```

core/federated/RTI/rti_common.h
```diff
@@ -246,25 +257,69 @@ tag_t earliest_future_incoming_message_tag(scheduling_node_t* e);
 tag_t eimt_strict(scheduling_node_t* e);
 
 /**
- * Return true if the node is in a zero-delay cycle.
+ * For the given scheduling node (enclave or federate), if necessary, update the `min_delays`,
+ * `all_upstremas`, `num_all_upstreams`, and the fields that indicate cycles.  These fields will be
```
edwardalee, line 261:
```suggestion
 * `all_upstreams`, `num_all_upstreams`, and the fields that indicate cycles.  These fields will be
```

core/federated/RTI/rti_common.h
```diff
@@ -246,25 +257,69 @@ tag_t earliest_future_incoming_message_tag(scheduling_node_t* e);
...
  * @param node The node.
  */
-bool is_in_zero_delay_cycle(scheduling_node_t* node);
+void update_min_delays_upstream(scheduling_node_t* node);
 
 /**
- * Return true if the node is in a cycle (possibly a zero-delay cycle).
+ * For the given scheduling node (enclave or federate), if necessary, update the `all_downstreams`,
+ * `num_all_downstreams`.  These fields will be updated only if they have not been previously updated
+ * or if invalidate_min_delays_upstream has been called since they were last updated.
```
edwardalee, line 271:
```suggestion
 * For the given scheduling node (enclave or federate), if necessary, update the `all_downstreams` and
 * `num_all_downstreams` fields.  These fields will be updated only if they have not been previously updated
 * or if invalidate_min_delays_upstream has been called since they were last updated.
```
This is a little confusing.  Should `invalidate_min_delays_upstream` be renamed `invalidate_min_delays`?  From this comment, it seems it also invalidates the downstream delays.

edwardalee (changes requested):
I need more time to review this, but here is a starting point.  This is looking very good so far. I have a few suggestions, some of which are just nits and you should feel free to ignore them.  I will follow up with a more complete review as soon as I can.

core/federated/RTI/main.c
byeonggiljun: Not really. `min_delays` is where the delays between each federate are stored. For example, if there are two federate A and B with one connection, A -/10 ms/-> B, the matrix `min_delays` looks like the below.

I note that `scheduling_node_t` no more contains [delays (from upstream)](https://github.com/lf-lang/reactor-c/blob/7db4930deaa0e4100b017b1d274e24aecba11e0d/core/federated/RTI/rti_common.h#L64) but stores **ID**s of upstream or downstream only. Currently, if `A` and `B` are not connected directly, only `B` knows that `min_delay` from `A` is `10 ms` and this is used for computing `EIMT` of `B`. With this matrix, `A` can also know that the delay to `B` is `10 ms` and this is needed for computing `DNET` of `A`.
|                |       A          |       B     |
| ---------| ---------- | ------- |
| A             | NEVER      |   10 ms  |
| B             | FOREVER  |   NEVER|

core/federated/RTI/rti_common.c
byeonggiljun: This function returns the greatest tag earlier than the given tag. And as far as I know, only `NEVER_TAG` is the tag earlier than `(0, 0)` in Lingua Franca. That's why I added this condition. 

Do we need to return `(-1, M_MAX)`, where M_MAX is the largest microstep, instead?

core/federated/RTI/rti_common.h
byeonggiljun: You're right. Thank you for pointing out this! I'll work on it along with the related comments you provided.

core/federated/RTI/rti_common.c
byeonggiljun: I renamed `invalidate_min_delays_upstream` to `invalidate_min_delays` and made that function reset delay information of all nodes in [4507043](https://github.com/lf-lang/reactor-c/pull/349/commits/4507043ca782ed8e19901fc58fd5cf5ea06422d9). Also resolved this comment, too.

core/federated/RTI/rti_common.h
byeonggiljun: Resolved in [4507043](https://github.com/lf-lang/reactor-c/pull/349/commits/4507043ca782ed8e19901fc58fd5cf5ea06422d9).

core/federated/RTI/rti_common.c
byeonggiljun: I updated the commend and moved it to `tag.h/c` in [2857ab8](https://github.com/lf-lang/reactor-c/pull/349/commits/2857ab87ff374b9a818eedae119327299a9e45fa).

core/federated/RTI/rti_common.c
edwardalee: Your other if branches will automatically return (-1, M_MAX), so if this is the right answer, there is no need for this branch.

You are right that we don’t use negative time values normally, but there is no fundamental reason we couldn’t.  In fact, we don’t normally use time values less than today…

core/tag.c
```diff
@@ -124,6 +124,26 @@ tag_t lf_delay_strict(tag_t tag, interval_t interval) {
   return result;
 }
 
+/**
+ * @brief Return the greatest tag earlier than the given tag.
+ *
+ * If the given tag is `FOREVER_TAG` or `NEVER_TAG`, however, just return the given tag.
+ * @param tag The tag.
+ */
```
edwardalee, line 132: I suggest removing the function comment from the .c file. It's in the .h file, and it's better to not have copies that will inevitably become inconsistent.

include/core/federated/network/net_common.h
```diff
@@ -670,6 +670,18 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
+/**
+ * Byte identifying a downstream next event tag (DNET) message sent
+ * from the RTI in centralized coordination.
+ * The next eight bytes will be the timestamp.
+ * The next four bytes will be the microstep.
+ * This signal from the RTI tells the destination federate the latest tag that
+ * the federate can safely skip sending a next event tag (NET) signal.
+ * In other words, the federate doesn't have to send NET signals with tags
+ * earlier than or equal to this tag unless it cannot advance to its next event tag.
```
edwardalee, line 681: I think this could be clearer:
```suggestion
 * This signal from the RTI tells the destination federate that downstream federates do not need
 * for it to send any next event tag (NET) signal with a tag _g_ less than the specified tag. Thus, it should only send
 * those signals if needs permission from the RTI to advance to _g_.
```

byeonggiljun: I applied this clearer and fancy description in [896e152](https://github.com/lf-lang/reactor-c/pull/349/commits/896e15293bd21958a124d7f73b02c455340f81d7).

core/tag.c
byeonggiljun: Resolved in [896e152](https://github.com/lf-lang/reactor-c/pull/349/commits/896e15293bd21958a124d7f73b02c455340f81d7).

core/federated/RTI/rti_common.c
byeonggiljun: Resolved in [896e152](https://github.com/lf-lang/reactor-c/pull/349/commits/896e15293bd21958a124d7f73b02c455340f81d7).

core/federated/RTI/rti_common.c
byeonggiljun: I got it. I deleted that `if` statement in [896e152](https://github.com/lf-lang/reactor-c/pull/349/commits/896e15293bd21958a124d7f73b02c455340f81d7).

edwardalee:
The example given in the PR comment is confusing.  It says:

For example, if the delay from `A` to `B` is `NEVER` and `B`'s next event is scheduled at `(100 ms, 0)`, the RTI can send `DNET (100 ms, 0)` to `A`; if `A` sends `NET (100 ms, 0)` to the RTI, the RTI **cannot** send `TAG (100 ms, 0)`, and if `A` sends `NET (100 ms, 1)`, the RTI can grant `TAG (100 ms, 0)` to `B`.

What does it mean "`B`'s next event is scheduled at `(100 ms, 0)`"?  I think you mean something like "The most recently received NET from `B` is `(100 ms, 0)`"?  Also, doesn't this have to be *transitive* NET?

The rest of the example is irrelevant to this PR.  The statement it makes is true without this PR and has always been true.  Can you instead give an example that actually uses this PR?

core/federated/RTI/rti_common.h
```diff
@@ -53,17 +53,20 @@ typedef struct scheduling_node_t {
...
+  interval_t* immediate_upstream_delays; // Minimum delay on connections from immdediate upstream scheduling nodes.
+                                         // Here, NEVER encodes no delay. 0LL is a microstep delay.
+  uint16_t num_immediate_upstreams;      // Size of the array of immediate upstream scheduling nodes and delays.
+  uint16_t* immediate_downstreams;       // Array of immediate downstream scheduling node ids.
+  uint16_t num_immediate_downstreams;    // Size of the array of immediate downstream scheduling nodes.
+  execution_mode_t mode;                 // FAST or REALTIME.
+  uint16_t* all_upstreams;               // Array of all upstream scheduling node ids.
+  uint16_t num_all_upstreams;            // Size of the array of all upstream scheduling nodes and delays.
+  uint16_t* all_downstreams;             // Array of all downstream scheduling node ids.
+  uint16_t num_all_downstreams;          // Size of the array of all downstream scheduling nodes.
```
edwardalee, line 68:
```suggestion
  uint16_t* immediate_upstream;    // Array of immediate upstream scheduling node ids.
  interval_t* immediate_upstream_delays; // Minimum delay on connections from immdediate upstream scheduling nodes.
                                         // Here, NEVER encodes no delay. 0LL is a microstep delay.
  uint16_t num_immediate_upstream;      // Size of the array of immediate upstream scheduling nodes and delays.
  uint16_t* immediate_downstream;       // Array of immediate downstream scheduling node ids.
  uint16_t num_immediate_downstream;    // Size of the array of immediate downstream scheduling nodes.
  execution_mode_t mode;                 // FAST or REALTIME.
  uint16_t* all_upstream;               // Array of all upstream scheduling node ids.
  uint16_t num_all_upstream;            // Size of the array of all upstream scheduling nodes and delays.
  uint16_t* all_downstream;             // Array of all downstream scheduling node ids.
  uint16_t num_all_downstream;          // Size of the array of all downstream scheduling nodes.
```

edwardalee: Now I'm thinking we don't really need the `all_upstream` and `all_downstream` fields at all.  Once you have constructed the `min_delays` array, you can iterate over all upstream from ID `i` as follows (pseudo code, where `n` is the number of nodes):
```
  for (j from 0 to n-1) {
    if (min_delays[i][j] != FOREVER_TAG && i != j) {
      ... found a downstream node
    }
  }
```
To iterate over all upstream nodes, just reverse `i` and `j` in the `if` statement above.

edwardalee (changes requested):
I'm worried about the n^2 complexity and memory cost.  I've added some suggestions, but I think the `all_upstream`, `all_downstream`, and `min_delay` arrays all have n^2 complexity in the worst case and they seem redundant.  I'm suggesting consolidating to just use the `min_delay` array for all these purposes.

byeonggiljun:
Hello, @edwardalee. 

> What does it mean "`B`'s next event is scheduled at `(100 ms, 0)`"? I think you mean something like "The most recently received NET from `B` is `(100 ms, 0)`"? Also, doesn't this have to be _transitive_ NET?

What I meant was "the minimum tag of the most recent NET_B and the head of the in-transit message queue of `B` is `(100 ms, 0)`" as the RTI uses **in-transit message queue** as well as NET from B to predict B's next earliest event. I'll modify the comment to clarify it.

> Also, doesn't this have to be _transitive_ NET?

I don't get what _transitive_ NET is. Could you please elaborate on it?

> The rest of the example is irrelevant to this PR. The statement it makes is true without this PR and has always been true. Can you instead give an example that actually uses this PR?

I tried to explain how a DNET signal is used in the example. Could you read it and share your thoughts?

edwardalee:
OK, the description is much better now.  I would suggest a slight enhancement. Instead of:

"Based on the DNET signal, A does not send any NET signals with tags earlier than or equal to (100 ms, 0) (NET with 10 ms, 20 ms, ..., 100 ms)"

you could say:

"Based on the DNET signal, if A is not producing outputs, A does not need to send any NET signals with tags earlier than or equal to (100 ms, 0) (NET with 10 ms, 20 ms, ..., 100 ms)"

core/federated/RTI/rti_common.h
byeonggiljun: Thanks for the suggestion, @edwardalee! I remove the fields `all_upstream` and `all_downstream` in [adc0a1d](https://github.com/lf-lang/reactor-c/pull/349/commits/adc0a1d4250bba7b29a39cb4dce635b0db7e5e0c). I resolved all other related comments, too.

core/federated/RTI/main.c
```diff
@@ -136,6 +136,7 @@ void usage(int argc, const char* argv[]) {
   lf_print("          clock sync attempt (default is 10). Applies to 'init' and 'on'.\n");
   lf_print("  -a, --auth Turn on HMAC authentication options.\n");
   lf_print("  -t, --tracing Turn on tracing.\n");
+  lf_print("  -d, --dnet Turn on DNET signals for reducing network messages.\n");
```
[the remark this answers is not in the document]
byeonggiljun: Ah, you're right. I make the option to turn off the feature in [c2f5316](https://github.com/lf-lang/reactor-c/pull/349/commits/c2f5316458bac61074285ca34d92a87c8511eaf9).

edwardalee (approved):

PR #351, opened 2024-02-06
Put a single event on the recycle queue initially to avoid runtime malloc

erlingrj:
An LF program with n timers need n+1 events. The last event should initially reside in the recycle queue to avoid any runtime memory allocations. This PR puts on there if the program has timers.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -550,9 +550,8 @@ void _lf_initialize_timers(environment_t* env) {
         }
     }
     
-    // Create an extra event and put it on the recycle queue. 
-    // An LF program with n timers require n+1 events. By putting the
-    // last event on the recycle queue we avoid allocating it at runtime.
+    // To avoid runtime memory allocations for timer-driven programs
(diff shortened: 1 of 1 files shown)


core/reactor_common.c
```diff
@@ -549,6 +549,14 @@ void _lf_initialize_timers(environment_t* env) {
             _lf_initialize_timer(env, env->timer_triggers[i]);
         }
     }
+    
+    // Create an extra event and put it on the recycle queue. 
+    // An LF program with n timers require n+1 events and the last event 
+    // needs to be initialized to the event recycle queue.
```
edwardalee, line 555: Unclear what this last clause means.  Perhaps just state in the comment that the reason is avoid runtime memory allocation for timer-driven programs?

edwardalee (approved):
Looks good to me.  The comment needs clarification, but otherwise, this is good to merge.

core/reactor_common.c
erlingrj: Thanks, updated

PR #354, opened 2024-02-12
Further cleanup

edwardalee:
This cleanup pass moves a renames many functions to separate user-facing header files (e.g. reactor.h, schedule.h) from internal header files (e.g. reactor_common.h, reactor_threaded.h). A major goal is as a first step towards getting Doxygen docs to give us useful pages, making function names more uniform, cleaning up header files, and making internal functions that are not any part of an API static. This also removes some obsolete function declarations that had no implementations.

There is a [companion PR in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/2210).

It may be easier to view the resulting header files rather than the diff when reviewing.

This work is not complete and there are still some inconsistent function names. The main focus of this PR was on reactor.h.  The functions defined there can be viewed as complete.


lib/schedule.c
```diff
@@ -8,132 +8,76 @@
...
+    // NOTE: This doesn't acquire the mutex lock in the multithreaded version
+    // until schedule_value is called. This should be OK because the element_size
+    // does not change dynamically.
+    if (template->type.element_size != sizeof(int)) {
+        lf_print_error("Action type is not an integer. element_size is %zu", template->type.element_size);
+        return -1;
+    }
+    int* container = (int*)malloc(sizeof(int));
+    *container = value;
+    return lf_schedule_value(action, extra_delay, container, 1);
```
petervdonovan, line 33: Note to other reviewers -- visually checked that this is the same as the original implementation, just in a different place.

(Anyone can go ahead and mark these as resolved when they are no longer helpful)

lib/schedule.c
```diff
@@ -8,132 +8,76 @@
...
-    return _lf_schedule_token((lf_action_base_t*)action, extra_delay, token);
+trigger_handle_t lf_schedule_token(lf_action_base_t* action, interval_t extra_delay, lf_token_t* token) {
+    environment_t* env = action->parent->environment;
+    
+    LF_CRITICAL_SECTION_ENTER(env);
+    int return_value = lf_schedule_trigger(env, action->trigger, extra_delay, token);
+    // Notify the main thread in case it is waiting for physical time to elapse.
+    lf_notify_of_event(env);
+    LF_CRITICAL_SECTION_EXIT(env);
+    return return_value;
```
petervdonovan, line 44: Note to other reviewers: this is the same as the original implementation

lib/schedule.c
```diff
@@ -8,132 +8,76 @@
...
+    // Initialize token with an array size of length and a reference count of 0.
+    lf_token_t* token = _lf_initialize_token(template, length);
+    // Copy the value into the newly allocated memory.
+    memcpy(token->value, value, template->type.element_size * length);
+    // The schedule function will increment the reference count.
+    trigger_handle_t result = lf_schedule_trigger(env, action->trigger, offset, token);
+    // Notify the main thread in case it is waiting for physical time to elapse.
+    lf_notify_of_event(env);
+    LF_CRITICAL_SECTION_EXIT(env);
+    return result;
```
petervdonovan, line 77: Note to other reviewers: same as original implementation

lib/schedule.c
```diff
@@ -142,7 +86,15 @@ trigger_handle_t lf_schedule_value(void* action, interval_t extra_delay, void* v
...
-    return _lf_schedule_value((lf_action_base_t*)action, extra_delay, value, (size_t)length);
+    token_template_t* template = (token_template_t*)action;
+    environment_t* env = action->parent->environment;
+    LF_CRITICAL_SECTION_ENTER(env);
+    lf_token_t* token = _lf_initialize_token_with_value(template, value, length);
+    int return_value = lf_schedule_trigger(env, action->trigger, extra_delay, token);
+    // Notify the main thread in case it is waiting for physical time to elapse.
+    lf_notify_of_event(env);
+    LF_CRITICAL_SECTION_EXIT(env);
+    return return_value;
```
petervdonovan, line 97: Note to other reviewers: this part is the same as the original implementation of `_lf_schedule_value`

lib/schedule.c
```diff
@@ -166,3 +118,230 @@ bool lf_check_deadline(void* self, bool invoke_deadline_handler) {
     }
     return false;
 }
+
+trigger_handle_t lf_schedule_trigger(environment_t *env, trigger_t* trigger, interval_t extra_delay, lf_token_t* token) {
```
petervdonovan, line 122: Note to other reviewers: Based on the original code from `reactor_common.c` that got deleted, this is what the diff should actually look like (relative to the target of this PR):
```
diff --git a/v0.txt b/v1.txt
index d459bde..7f7c928 100644
--- a/v0.txt
+++ b/v1.txt
@@ -1,5 +1,5 @@
-    assert(env != GLOBAL_ENVIRONMENT);
-    if (_lf_is_tag_after_stop_tag(env, env->current_tag)) {
+assert(env != GLOBAL_ENVIRONMENT);
+    if (lf_is_tag_after_stop_tag(env, env->current_tag)) {
         // If schedule is called after stop_tag
         // This is a critical condition.
         _lf_done_using(token);
@@ -12,13 +12,13 @@
         extra_delay = 0LL;
     }
 
-    LF_PRINT_DEBUG("_lf_schedule: scheduling trigger %p with delay " PRINTF_TIME " and token %p.",
+    LF_PRINT_DEBUG("lf_schedule_trigger: scheduling trigger %p with delay " PRINTF_TIME " and token %p.",
             trigger, extra_delay, token);
 
     // Increment the reference count of the token.
     if (token != NULL) {
         token->ref_count++;
-        LF_PRINT_DEBUG("_lf_schedule: Incremented ref_count of %p to %zu.",
+        LF_PRINT_DEBUG("lf_schedule_trigger: Incremented ref_count of %p to %zu.",
                 token, token->ref_count);
     }
 
@@ -41,11 +41,11 @@
     }
     tag_t intended_tag = (tag_t){.time = env->current_tag.time + delay, .microstep = 0};
 
-    LF_PRINT_DEBUG("_lf_schedule: env->current_tag.time = " PRINTF_TIME ". Total logical delay = " PRINTF_TIME "",
+    LF_PRINT_DEBUG("lf_schedule_trigger: env->current_tag.time = " PRINTF_TIME ". Total logical delay = " PRINTF_TIME "",
             env->current_tag.time, delay);
     interval_t min_spacing = trigger->period;
 
-    event_t* e = _lf_get_new_event(env);
+    event_t* e = lf_get_new_event(env);
 
     // Initialize the next pointer.
     e->next = NULL;
@@ -101,11 +101,11 @@
                 found = found->next;
                 intended_tag.microstep++;
             }
-            if (_lf_is_tag_after_stop_tag(env, intended_tag)) {
+            if (lf_is_tag_after_stop_tag(env, intended_tag)) {
                 LF_PRINT_DEBUG("Attempt to schedule an event after stop_tag was rejected.");
                 // Scheduling an event will incur a microstep
                 // after the stop tag.
-                _lf_recycle_event(env, e);
+                lf_recycle_event(env, e);
                 return 0;
             }
             // Hook the event into the list.
@@ -134,14 +134,14 @@
                     // Recycle the new event and decrement the
                     // reference count of the token.
                     _lf_done_using(token);
-                    _lf_recycle_event(env, e);
+                    lf_recycle_event(env, e);
                     return(0);
                 case replace:
                     LF_PRINT_DEBUG("Policy is replace. Replacing the previous event.");
                     // If the event with the previous time is still on the event
                     // queue, then replace the token.  To find this event, we have
                     // to construct a dummy event_t struct.
-                    event_t* dummy = _lf_get_new_event(env);
+                    event_t* dummy = lf_get_new_event(env);
                     dummy->next = NULL;
                     dummy->trigger = trigger;
                     dummy->time = trigger->last_tag.time;
@@ -150,13 +150,13 @@
                     if (found != NULL) {
                         // Recycle the existing token and the new event
                         // and update the token of the existing event.
-                        _lf_replace_token(found, token);
-                        _lf_recycle_event(env, e);
-                        _lf_recycle_event(env, dummy);
+                        lf_replace_token(found, token);
+                        lf_recycle_event(env, e);
+                        lf_recycle_event(env, dummy);
                         // Leave the last_tag the same.
                         return(0);
                     }
-                    _lf_recycle_event(env, dummy);
+                    lf_recycle_event(env, dummy);
 
                     // If the preceding event _has_ been handled, then adjust
                     // the tag to defer the event.
@@ -188,11 +188,11 @@
 
     // Do not schedule events if if the event time is past the stop time
     // (current microsteps are checked earlier).
-    LF_PRINT_DEBUG("Comparing event with elapsed time " PRINTF_TIME " against stop time " PRINTF_TIME ".", e->time - start_time, env->stop_tag.time - start_time);
+    LF_PRINT_DEBUG("Comparing event with elapsed time " PRINTF_TIME " against stop time " PRINTF_TIME ".", e->time - lf_time_start(), env->stop_tag.time - lf_time_start());
     if (e->time > env->stop_tag.time) {
-        LF_PRINT_DEBUG("_lf_schedule: event time is past the timeout. Discarding event.");
+        LF_PRINT_DEBUG("lf_schedule_trigger: event time is past the timeout. Discarding event.");
         _lf_done_using(token);
-        _lf_recycle_event(env, e);
+        lf_recycle_event(env, e);
         return(0);
     }
 
@@ -208,7 +208,7 @@
     // rather than the event_q, so anything put in the event_q with this
     // same time will automatically be executed at the next microstep.
     LF_PRINT_LOG("Inserting event in the event queue with elapsed time " PRINTF_TIME ".",
-            e->time - start_time);
+            e->time - lf_time_start());
     pqueue_insert(env->event_q, e);
 
     tracepoint_schedule(env->trace, trigger, e->time - env->current_tag.time);
```

petervdonovan (approved):
I found no problems in this PR. The only comments that I left are more high-level comments about the diff and notes to help other reviewers understand the diff.

erlingrj:
What is our convention for selecting the name for a function that "does something" between the following:

1.  `do_something`
2. `lf_do_something`
3. `_lf_do_something`

cmnrd:
I was wondering the same. In particular, I noticed that https://github.com/lf-lang/lingua-franca/pull/2210/files removes the `_` prefix from several functions.

I think the policy used to be that the `_` prefix is used on functions that are visible in reaction bodies, but that are not supposed to be used directly by the user. Does this PR ensure that the renamed functions (`_` prefix removed) are not visible in reaction bodies?

edwardalee:
> What is our convention for selecting the name for a function that "does something" between the following:
> 
> 1. `do_something`
> 2. `lf_do_something`
> 3. `_lf_do_something`

Good question.  I think the original intent was that `do_something` should only be used for static functions, where the scope is limited to the file.  `lf_do_something` should be used for user-facing functions, such as those used in reaction bodies.  And `_lf_do_something` should be used for things that can't be static because they are used in multiple files, but they are functions that a user would never invoke.

The problem I ran into is that it's not always clear which functions a user might want to invoke, and this could change over time. For example, should it be `lf_new_reactor` or `_lf_new_reactor`?  `lf_allocate` or `_lf_allocate`?

I've been opting for doing away with the leading underscore when there is any doubt, but I wonder whether we should do away with it completely?  Whether something is user facing, I think, should be determined by which header file declares it.  If it's declared in `reactor.h`, then it is user facing and is in scope for reactions.  If it is declared in `reactor_common.h`, then it is not in scope for reactions.  If we organize the Doxygen docs well, all of this could be made very clear.

An advantage of doing away with the underscores altogether is that we can change our minds later relatively easily.  Right now, `_lf_schedule_at_tag` is not a user-facing function, but maybe we will want it to be later.  If we remove the leading underscore now, then the change is painless: Just move its declaration to `reactor.h`.

So my proposal is:
> 1. `do_something` for static functions (scope limited to the current file).
> 2. `lf_do_something` for everything else.

This PR does not make this change, but maybe we can do is part of the formatting PR?  Or as a separate PR in rapid succession?

Edward

core/threaded/reactor_threaded.c
```diff
@@ -654,10 +621,20 @@ void _lf_initialize_start_tag(environment_t *env) {
...
     // set the current time back by the STA offset.
-    env->current_tag = (tag_t){.time = start_time - _lf_fed_STA_offset, .microstep = 0u};
+    env->current_tag.time -= lf_fed_STA_offset;
+    LF_PRINT_LOG("Waiting for start time " PRINTF_TIME " plus STA " PRINTF_TIME ".",
+            start_time, lf_fed_STA_offset);
+#else
+    instant_t lf_fed_STA_offset = 0;
+    LF_PRINT_LOG("Waiting for start time " PRINTF_TIME ".",
+            start_time);
+#endif
```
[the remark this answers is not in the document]
edwardalee: Thanks for the comment here.  I've added a clarifying comment should help reasoning about the code:
```
    // For other than federated decentralized execution, there is no lf_fed_STA_offset variable defined.
    // To use uniform code below, we define it here as a local variable.
```

cmnrd:
> So my proposal is:
> 
> > 1. `do_something` for static functions (scope limited to the current file).
> > 2. `lf_do_something` for everything else.
> 
> This PR does not make this change, but maybe we can do is part of the formatting PR? Or as a separate PR in rapid succession?
> 

If we truly ensure that the internal API does not leak into the public API via some hidden dependencies, then yes, I think this would be feasible. Having an additional prefix has the advantage that we clearly label the functions. So, also when the function is used, we know whether it originates from a public or a private API. This is not strictly required though and I am fine with either solution.

The separation between public user-facing API and internal API is a great outcome of this PR. I think we should make it more explicit, by moving all headers that are user-facing in a single location and adding this location to cmake using the `PUPLIC` interface, while we should add all internal header locations using the `PRIVATE` interface. Then, we can be sure that only the public headers are visible in LF code.

cmnrd:
I agree that we can address this in a later PR. I opened https://github.com/lf-lang/reactor-c/issues/360 to keep track of the action items that we are aware of.

erlingrj (approved):
This looks very good. I guess we would like to move a lot more into the `api/` folder? E.g. `lf_time_*` functions, `lf_tag*` functions? Or will we make a difference between functions that are purely user-facing and those that are also used internally by the runtime? (This is a topic for the next PR I guess)

cmnrd:
I think we will need to different api folders, one user facing (in lf code) and one internally. My hope, however, is that we don't need an explicit internal directory. In a sense, the headers defined by the modules should define their API. A separate API module would only be needed if we have circular dependencies between modules, which hopefully we can avoid (but that is unclear).

PR #355, opened 2024-02-13
Extend platform API with functions for setting thread scheduling policy, core affinity and priority + remove C11 as platform

### LINKED ISSUE #290: Extensions to the threaded API, do we really need the C11 threads platform?
Hi.

The need to configure the worker-affinity (e.g. pin LF worker to a specific core) and to set the priority of the workers and the scheduling strategy of the underlying thread library is getting apparent. Especially in the static scheduler work. It doesn't make sense to let the Linux kernel migrate our workers around from core to core, we have our own scheduler that distributes the work.

I am experimenting with adding some new functions to the threaded API currently I am thinking of:

```
int lf_thread_scheduler_init();
int lf_thread_set_cpu(lf_thread_t thread, int cpu_number);
int lf_thread_set_priority(lf_thread_t thread, int priority);
lf_thread_t lf_thread_self();
```

The first function is meant to configure the underlying threads scheduler. On Linux this could be using a real-time scheduler like SCHED_FIFO or SCHED_RR. 

Anyways, there is a straight forward way of implementing this with `pthread`, probably also with the windows API. However, C11 Threads does not have a rich enough API to configure these things. We could still use C11 for the thread synchronization and extend it with pthreads for the scheduling. But is this additional preprocessing complexity worth it? What is actually the benefit of using C11 Threads when we have pthreads support? Are there any platforms with C11 support and not pthread support?

erlingrj:
This PR adds the following functions to the platform API:
- lf_thread_self
- lf_thread_set_priority
- lf_thread_set_cpu
- lf_thread_set_scheduling_policy

It provides implementations for POSIX and Zephyr. It also removes the C11 as a platform because it provides a strict subset of the functionality of the POSIX and Windows platforms which we already support. I have not heard about any platform not supporting POSIX (or windows) that does support C11. 

I view this thread scheduling API as experimental and that it is likely to be changed in the future when we start using it with things like enclaved scheduling. The reason to add it now, before we use it in the runtime, is to enable the user to write real-time programs by setting a high priority and a real-time scheduling policy for their workers.

TODOs:
- [x] Can we merge `lf_thread_self()` and `lf_thread_id()`
- [x] Zephyr should return error if fair scheduler is requested (it only does priority)

Closes #290


erlingrj:
@soyerefsane, this should make your benchmarking easier.

edwardalee:
What is the status of this PR?  The changes look very useful to me.

erlingrj:
I havnt had the time to revisit this draft again. I think it is quite complete. I will need to look through it again. There is a couple of unanswered questions still. I have updated the PR with a checklist of missing items.

edwardalee:
This branch has relatively simple conflicts with main.  Is it OK to just overwrite main so we can get the tests to run?

erlingrj:
I can solve the conflicts. Two secs

erlingrj (commented):
Did a self-review and I believe it looks OK now

low_level_platform/api/low_level_platform.h
```diff
@@ -107,12 +107,16 @@ int lf_mutex_lock(lf_mutex_t* mutex);
  */
 int lf_available_cores();
 
+/**
+ * @brief Returns the lf_thread_t of the calling thread.
```
lhstrh, line 111:
```suggestion
 * @brief Return the lf_thread_t of the calling thread.
```

low_level_platform/api/low_level_platform.h
```diff
@@ -132,6 +136,50 @@ int lf_thread_create(lf_thread_t* thread, void* (*lf_thread)(void*), void* argum
...
+} lf_scheduling_policy_type_t;
+
+typedef struct {
+  lf_scheduling_policy_type_t policy; // The scheduling policy
+  int priority;                       // The priority, if applicable
+  interval_t time_slice;              // The time-slice allocated, if applicable.
+} lf_scheduling_policy_t;
+
+/**
+ * This pins a lf_thread to a specific CPU
```
lhstrh, line 157:
```suggestion
 * @brief Pin a thread to a specific CPU.
```

low_level_platform/api/low_level_platform.h
```diff
@@ -132,6 +136,50 @@ int lf_thread_create(lf_thread_t* thread, void* (*lf_thread)(void*), void* argum
...
+ *
+ * @param thread The thread
+ * @param cpu_number the CPU ID
+ * @return 0 on success, platform-specific error number otherwise.
+ */
+int lf_thread_set_cpu(lf_thread_t thread, int cpu_number);
+
+/**
+ * Sets the priority of a thread. Priority ranges from 0 to 99 where a higher
```
lhstrh, line 166:
```suggestion
 * @brief Set the priority of a thread. 
 * Priority ranges from 0 to 99 where a higher
```

low_level_platform/api/low_level_platform.h
```diff
@@ -132,6 +136,50 @@ int lf_thread_create(lf_thread_t* thread, void* (*lf_thread)(void*), void* argum
...
+ * makes sense if the thread is scheduled with LF_SCHED_TIMESLICE or LF_THREAD_PRIORITY
+ *
+ * @param thread The thread.
+ * @param priority The priority.
+ * @return int 0 on success, platform-specific error otherwise
+ */
+int lf_thread_set_priority(lf_thread_t thread, int priority);
+
+/**
+ * Sets the scheduling policy of a thread.
```
lhstrh, line 177:
```suggestion
 * @brief Set the scheduling policy of a thread.
```

lhstrh (approved):
Only nitpicks regarding comments from my side.

low_level_platform/api/low_level_platform.h
edwardalee, line 177: One thing that would help me, at least, if this explains what it means to set a scheduling policy for a thread.  Doesn't a scheduling policy apply to a group of threads?  What if some threads have one scheduling policy and others have a different scheduling policy?  Maybe add a pointer to this page: https://man7.org/linux/man-pages/man7/sched.7.html. That page seems to explain it pretty well.  Also, we need to explain how the LF scheduling policies map to Linux scheduling policies, at least as an example.  E.g., `LF_SCHED_PRIORITY` to `SCHED_FIFO`.

low_level_platform/api/low_level_platform.h
```diff
@@ -132,6 +136,50 @@ int lf_thread_create(lf_thread_t* thread, void* (*lf_thread)(void*), void* argum
  */
 int lf_thread_join(lf_thread_t thread, void** thread_return);
 
+#define LF_SCHED_MAX_PRIORITY 99
+#define LF_SCHED_MIN_PRIORITY 0
```
edwardalee, line 140: In the case of Linux, shouldn't these be the values returned by `sched_get_priority_max` and `sched_get_priority_min`?

low_level_platform/impl/src/lf_zephyr_support.c
```diff
@@ -154,6 +154,40 @@ void initialize_lf_thread_id() {
...
+  // Update the policy
+  switch (policy->policy) {
+    break;
+  case LF_SCHED_TIMESLICE: {
+    k_thread_priority_set(thread, LF_SCHED_MAX_PRIORITY - policy->priority);
+    k_sched_time_slice_set(0, policy->time_slice / MSEC(1));
+    break;
+  }
+  case LF_SCHED_PRIORITY: {
+    k_thread_priority_set(thread, 99 - policy->priority);
```
edwardalee, line 179: Should this `99` be `LF_SCHED_MAX_PRIORITY`?

edwardalee (approved):
Looks great to me!  I added some small suggestions, mostly about documentation. Let's merge this!

low_level_platform/api/low_level_platform.h
erlingrj: My thinking is that we pick these values for LF, i.e. no matter what platform you run on, setting priority of your `lf_thread_t` to 99 will be the highest.

I have now included logic for translating from this range to the range of the target platform

low_level_platform/impl/src/lf_zephyr_support.c
erlingrj: I have fixed this now

low_level_platform/api/low_level_platform.h
erlingrj: Yes thanks

erlingrj (commented):
Thanks for the feedback now I have improved the implementation to map a priority in the range [LF_SCHED_MIN_PRI, LF_SCHED_MAX_PRI] to the range of the target platform.

lhstrh:
Looks like Arduino is now failing...

erlingrj:
Our Arduino-support is really frustrating as adding source files and header files in reactor-c must be accompanied by changes to our compiler to move/delete files because the `arduino-cli` is so limited...

PR #356, opened 2024-02-14
Extensive platform-specific tests no longer run in PRs

lhstrh:
...unless necessitated by changes to specific files. We still run all platforms in `main`.


core/platform/lf_zephyr_support.c
```diff
@@ -22,7 +22,7 @@ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT
 STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ***************/
-
+// FIXME: Comment added to see if the Zephyr tests will run.
```
edwardalee, line 25: Do we want this in the merge?

edwardalee (commented):
It seems that check-diff.sh is missing from this PR. I can't really evaluate it without that.

core/platform/lf_zephyr_support.c
lhstrh: No, something was wrong and I ended up debugging after I had already requested review. The shell script is in `lingua-franca`. Also see https://github.com/lf-lang/lingua-franca/pull/2194 for a bugfix in the script.

PR #358, opened 2024-02-20
Add support for transient federates for centralized and decentralized coordination.

ChadliaJerad:
This PR replaces https://github.com/lf-lang/reactor-c/pull/192, and supports the latest major refactoring of the RTI. 

It implements the transient feature in the federation execution. Details of the implementation are documented in Discussion https://github.com/lf-lang/lingua-franca/discussions/2212.

This should be merged with lf-lang/lingua-franca#2213

---
## Merged #574

This PR  builds on the centralized transient support from the #358  PR of `transient-fed`  branch, extending the `reactor-c` runtime to support transient federates under decentralized coordination, where all connections are P2P and no RTI message forwarding occurs. It also adds support for physical connections involving transient federates under centralized coordination as well, which are P2P connections.

lf-lang/lingua-franca#2609

# Protocol extensions for decentralized transients

## Connection

- `lf_connect_to_federate()`, called when executing the preamble, is updated   to accept an `is_transient` flag; transient outbounds are queried once (no  retry loop) and connected only when the RTI sends `MSG_TYPE_OUTBOUND_CONNECTED`. Non-transient outbounds still use the original pattern and keep retrying every `ADDRESS_QUERY_RETRY_INTERVAL` until  connected.
 - The RTI keeps track of which transients are outbound of each federate in the `outbound_transients` array and updates `number_of_outbound_transients`  accordingly.
- `MSG_TYPE_OUTBOUND_CONNECTED` and  `MSG_TYPE_OUTBOUND_DISCONNECTED` are new  messages. The RTI sends these to the federate whose downstream peers have connected or disconnected. When an outbound transient connects, the federate   queries its address from the RTI. When it disconnects, the federate skips   sending messages, thus avoiding an error writing to a broken pipe.
- `MSG_TYPE_ADDRESS_QUERY` is extended with an `is_transient` byte so the RTI can register the querying federate's outbound-transient relationships.
- `outbound_p2p_connection_is_transient[NUMBER_OF_FEDERATES]` and `inbound_p2p_connection_is_transient[NUMBER_OF_FEDERATES]` are added to  `federate_instance_t`.

## Adaptations

- `mark_inputs_known_absent()`: for inbound transients, the tag is set to   `env->current_tag` instead of `FOREVER_TAG` when the P2P socket closes.   `FOREVER_TAG` permanently blocks ports from being updated after a transient rejoins, causing spurious __Attempt to update to earlier tag__ errors and outbound STP violations. `env->current_tag` is sufficient to unblock the scheduler while leaving the port open to future updates.
- Deferred P2P connection in `get_start_time_from_rti()`: `MSG_TYPE_OUTBOUND_CONNECTED` arriving during the start-time handshake is now deferred — the federate ID is drained from the socket immediately, but `lf_connect_to_federate()` (which itself reads from the RTI socket) is called only after `MSG_TYPE_TIMESTAMP` is received. This eliminates a race condition where the address-query reply consumed the timestamp bytes, leading to __Unexpected reply of type 2__.
- `notify_federate_disconnected()`: now calls   `send_outbound_disconnected_locked()` in addition to the existing   `send_upstream_disconnected_locked()` calls, so inbound federates of a   departing transient close their outbound P2P sockets.
- RTI shutdown: suppressed the spurious `WARNING: Failed to accept the socket. Invalid argument` that fired because `accept()` returns `EINVAL` when unblocked by the intentional  `shutdown_socket()` call at end of execution.

## Tracing and visualization (`fedsd`)

- `send_OUTBOUND_CONNECTED` / `receive_OUTBOUND_CONNECTED` and  `send_OUTBOUND_DISCONNECTED` / `receive_OUTBOUND_DISCONNECTED` trace events are added to `trace_types.h` and `fedsd.py`.
- `P2P_MSG` arrows now connect sender and receiver correctly in `fedsd`: matching is done by physical-time ordering rather than `partner_id` (which is `-1` on both sides for direct P2P tracepoints).


core/federated/RTI/rti_remote.c
```diff
@@ -833,6 +1018,47 @@ void handle_address_ad(uint16_t federate_id) {
     }
 }
 
+/**
+ * Send to the start time to the federate my_fed.
```
edwardalee, line 1022:
```suggestion
 * @brief Send to the start time to the federate my_fed.
 *
```

edwardalee (dismissed):
This looks like a great start, but I have some doubts about the concurrency handling as indicated in the comments. I will have to finish reviewing later, but I think there is enough to work with to go ahead and submit this provisional review.

include/core/federated/federate.h
```diff
@@ -554,4 +560,37 @@ void lf_synchronize_with_other_federates();
...
+ * 
+ * This function is particularly useful for testing transient federates.
+ */
+void lf_stop();
+    
+/**
+ * @brief Return the directory containing the executables of the individual
+ * federates.
+ */
+char* lf_get_federates_bin_directory();
```
cmnrd, line 577: Exposing information about the file system location at compile time to the program seems like a pretty bad idea to me. Paths can change, binaries get deployed to other systems. If it really is needed, it should be provided as a parameter at runtime.

include/core/federated/federate.h
```diff
@@ -554,4 +560,37 @@ void lf_synchronize_with_other_federates();
...
 
+/**
+ * @brief Stop the execution of a federate.
+ * Every enclave within the federate will stop at one microstep later than its
+ * current tag. Unlike lf_request_stop(), this process does not require any 
+ * involvement from the RTI, nor does it necessitate any consensus.
+ * 
+ * This function is particularly useful for testing transient federates.
+ */
+void lf_stop();
```
cmnrd, line 571: If this is used for testing, should we really make it part of the public API?

include/core/federated/network/net_common.h
```diff
@@ -434,7 +434,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define MSG_TYPE_PROVISIONAL_TAG_ADVANCE_GRANT 8
 
 /** 
- * Byte identifying a latest tag complete (LTC) message sent by a federate
+ * Byte identifying a latest tag complete (LTC) message sent by a federate30
```
cmnrd, line 437: ?

cmnrd (dismissed):

core/federated/RTI/main.c
```diff
@@ -281,6 +298,11 @@ int process_args(int argc, const char* argv[]) {
         usage(argc, argv);
         return 0;
     }
+    if (rti.number_of_transient_federates > rti.base.number_of_scheduling_nodes) {
```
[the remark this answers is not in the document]
ChadliaJerad: Thank you for the catch! We still need to have at least one persistent federate. The condition will become `>=` in the next commit.

cmnrd:
In the "files changed" view, you can add suggestions to batch and commit them all at once. This way, we can avoid having lots of commit messages without a descriptive commit message.

lhstrh:
> In the "files changed" view, you can add suggestions to batch and commit them all at once. This way, we can avoid having lots of commit messages without a descriptive commit message.

You probably want to use an interactive rebase to `squash` or even `fixup` all these commits...

core/federated/RTI/rti_remote.c
```diff
@@ -252,6 +314,36 @@ void notify_provisional_tag_advance_grant(scheduling_node_t *e, tag_t tag) {
         // Need to wait here.
         lf_cond_wait(&sent_start_time);
     }
+
+    // Check if sending the tag advance grant needs to be delayed or not
+    // Delay is needed when a federate has, at least one, absent upstream transient
+    int num_absent_upstram_transients = 0;
```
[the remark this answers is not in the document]
ChadliaJerad: Done.

core/federated/federate.c
```diff
@@ -1966,23 +1994,25 @@ void lf_connect_to_rti(const char* hostname, int port) {
...
-        unsigned char buffer[4];
+        unsigned char buffer[5];
         buffer[0] = MSG_TYPE_FED_IDS;
         // Next send the federate ID.
         if (_lf_my_fed_id > UINT16_MAX) {
             lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX);
         }
         encode_uint16((uint16_t)_lf_my_fed_id, &buffer[1]);
+        // Next send the federate type (persistent or transient)
+        buffer[1 + sizeof(uint16_t)] = _fed.is_transient? 1 : 0;
```
edwardalee, line 2005: This change to the `MSG_TYPE_FED_IDS` needs to be documented in `net_common.h`.

ChadliaJerad: Done in [b33b863](https://github.com/lf-lang/reactor-c/pull/358/commits/b33b86304f9b8efc5b165d9b4765667045ed853b).

core/federated/RTI/rti_remote.c
```diff
@@ -240,7 +247,62 @@ void notify_tag_advance_grant(scheduling_node_t *e, tag_t tag) {
...
+    lf_mutex_lock(&rti_mutex);
+    if (lf_tag_compare(fed->pending_grant, NEVER_TAG) == 0) {
+        // If a tag is issued, then stop any possible provisional tag grant
+        fed->pending_grant = tag;
+        fed->pending_provisional_grant = NEVER_TAG;
+        lf_thread_create(&(fed->pending_grant_thread_id), pending_grant_thread, fed);
+    } else {
+        // If there is already a pending tag grant, then let it be sent first
+        // FIXME: Is this correct?
+    }
```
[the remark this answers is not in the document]
ChadliaJerad: This is outdated and a different approach was used to address delayed grants (that is using a pqueue ordered by tags).

network/api/net_common.h
```diff
@@ -737,55 +762,66 @@
...
+#define MSG_TYPE_OUTBOUND_DISCONNECTED_LENGTH (1 + sizeof(uint16_t))
 
 /**
- * @brief Code sent with a @ref MSG_TYPE_REJECT message indicating that the
- * connected to the wrong server.
- * @ingroup Network
+ * Byte sent by the RTI ordering the federate to stop. Upon receiving the message,
+ * the federate will call lf_stop(), which will make it resign at its current_tag
+ * plus 1 microstep.
+ * The next 8 bytes will be the time at which the federates will stop.
```
ChadliaJerad:
```suggestion
```

trace/api/types/trace_types.h
```diff
@@ -53,6 +53,8 @@ typedef enum {
   send_ADR_QR,
   send_ADR_QR_REP,
   send_DNET,
+  send_UPSTREAM_CONNECTED,
+  send_UPSTREAM_DISCONNECTED,
```
ChadliaJerad:
```suggestion
```

trace/api/types/trace_types.h
```diff
@@ -53,6 +53,8 @@ typedef enum {
...
   // Receiving messages
   receive_ACK,
   receive_FAILED,
```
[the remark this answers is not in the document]
ChadliaJerad: Done in [bf9c3a7](https://github.com/lf-lang/reactor-c/pull/358/commits/bf9c3a76fad0741b088c2d7518dbfa9bd19bce4f)

core/federated/RTI/rti_remote.c
```diff
@@ -1466,6 +2185,181 @@ void lf_connect_to_federates(net_abstraction_t rti_net) {
...
+
+        LF_PRINT_LOG("RTI: Send MSG_TYPE_STOP to old federate %d.", fed_id);
+        send_stop(fed_old);
+        LF_MUTEX_UNLOCK(&rti_mutex);
+
+        // Wait for the old federate to send MSG_TYPE_RESIGN
+        LF_PRINT_LOG("RTI: Waiting for old federate %d to send resign.", fed_id);
+        // FIXME: This is a busy wait!  Need instead a lf_cond_wait on a condition variable.
+        while (!hot_swap_old_resigned) {
+        }
```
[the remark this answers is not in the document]
ChadliaJerad: Done in this commit [bf9c3a7](https://github.com/lf-lang/reactor-c/pull/358/commits/bf9c3a76fad0741b088c2d7518dbfa9bd19bce4f)

network/api/net_common.h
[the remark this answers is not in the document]
ChadliaJerad: Done in this commit: [7cb6639](https://github.com/lf-lang/reactor-c/pull/358/commits/7cb663964c2d6580476f7f8c71d7546991c1481e)

core/federated/federate.c
```diff
@@ -706,7 +767,7 @@ static int handle_port_absent_message(net_abstraction_t net, int fed_id) {
     tracepoint_federate_from_federate(receive_PORT_ABS, _lf_my_fed_id, fed_id, &intended_tag);
   }
   LF_PRINT_LOG("Handling port absent for tag " PRINTF_TAG " for port %hu of fed %d.",
-               intended_tag.time - lf_time_start(), intended_tag.microstep, port_id, fed_id);
+               intended_tag.time - lf_time_start(), intended_tag.microstep, port_id, _lf_my_fed_id);
```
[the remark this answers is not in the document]
ChadliaJerad: This should remain fed_id, because the federate is logging from which federate it is receiving a port-absent message.

core/federated/RTI/rti_remote.c
```diff
@@ -1046,14 +1701,72 @@ void* federate_info_thread_TCP(void* fed) {
...
+
+  // Nothing more to do. Close the net and exit.
+  // Prevent multiple threads from closing the same net at the same time.
+  LF_MUTEX_LOCK(&rti_mutex);
+  shutdown_net(my_fed->net, false);
+  my_fed->net = NULL;
+  // Manual clean, in case of a transient federate
+  if (my_fed->is_transient) {
+    // FIXME: Aren't there transit messages anymore???
+    // free_in_transit_message_q(my_fed->in_transit_message_tags);
```
edwardalee, line 1708: We should not leave such a FIXME and commented out code. What does the FIXME mean?

core/federated/RTI/rti_remote.c
```diff
@@ -692,71 +1103,309 @@ void handle_address_ad(uint16_t federate_id) {
...
+             (rti_remote->base.number_of_scheduling_nodes - rti_remote->number_of_transient_federates)) {
+        lf_cond_wait(&received_start_times);
+      }
+    }
+    // Add an offset to the maximum tag to get everyone starting together.
+    start_time = rti_remote->max_start_time + DELAY_START;
+    // Set the start_time in the RTI trace
+    if (rti_remote->base.tracing_enabled) {
+      lf_tracing_set_start_time(start_time);
+    }
```
edwardalee, line 1258: Shouldn't these be done only in the if branch?  I guess it's harmless, but the tracing and the setting of the start time will be done repeatedly, as many times as there are joining federates.

core/federated/RTI/rti_remote.c
```diff
@@ -692,71 +1103,309 @@ void handle_address_ad(uint16_t federate_id) {
...
+        size_t queue_size = pqueue_tag_size(upstream->in_transit_message_tags);
+        if (queue_size != 0) {
+          tag_t max_tag = pqueue_tag_max_tag(upstream->in_transit_message_tags);
+
+          if (lf_tag_compare(max_tag, my_fed->effective_start_tag) >= 0) {
+            my_fed->effective_start_tag = max_tag;
+            my_fed->effective_start_tag.microstep++;
+          }
+        }
+      }
```
edwardalee, line 1327: Shouldn't this also take into account the tags of any messages that the RTI has received from upstream federates but not forwarded because the transient was not connected?  I.e., I'm not sure it's enough to just look at the `in_transit_message_tags`, which does not contain the tags of messages that were previously received by the RTI but not forwarded.

core/federated/RTI/rti_remote.c
```diff
@@ -1149,18 +1865,62 @@ static int32_t receive_and_check_fed_id_message(net_abstraction_t fed_net) {
...
+    // Set that hot swap is in progress
+    lf_mutex_lock(&rti_mutex);
+    hot_swap_in_progress = true;
+    lf_mutex_unlock(&rti_mutex);
+    hot_swap_old_resigned = false;
+    fed = hot_swap_federate;
+    lf_print_info("RTI: Hot Swap starting for federate %d.", fed_id);
+  } else {
+    fed = fed_twin;
+    fed->is_transient = is_transient;
```
edwardalee, line 1921: Shouldn't this happen for both branches of the if?  I.e., put this line after the if-else?

edwardalee (changes requested):
Still working through the code, but I thought I would register these potential issues now.  Also, I'm working on PR #592 which has a bunch of refactoring and typo fixes.  It also tries to clarify some of the comments.  I'll ask for a review when it's ready.

core/federated/RTI/rti_remote.c
```diff
@@ -1258,6 +2037,54 @@ static int receive_connection_information(net_abstraction_t fed_net, uint16_t fe
...
+        }
+        // hot_swap_federate is freed by the caller on failure.
+        return 0;
+      }
+      // Scratch federate used only to validate an unchanged neighborhood on rejoin.
+      if (scratch != NULL) {
+        free_federate_info(scratch);
+      } else {
+        // Hot swap: keep the validated copy on hot_swap_federate.
+        fed->neighbor_structure_received = true;
```
edwardalee, line 2082: Why is this done only if `scratch == NULL`?

core/federated/RTI/rti_remote.c
```diff
@@ -1474,6 +2314,206 @@ void lf_connect_to_federates(net_abstraction_t rti_net) {
...
+        federate_info_t* fed_old = GET_FED_INFO(fed_id);
+        hot_swap_federate->enclave.completed = fed_old->enclave.completed;
+
+        LF_PRINT_LOG("RTI: Send MSG_TYPE_STOP to old federate %d.", fed_id);
+        send_stop(fed_old);
+        LF_MUTEX_UNLOCK(&rti_mutex);
+
+        // Wait for the old federate to send MSG_TYPE_RESIGN
+        LF_PRINT_LOG("RTI: Waiting for old federate %d to send resign.", fed_id);
+        LF_MUTEX_LOCK(&rti_mutex);
```
edwardalee, line 2383: Why release the mutex here and then immediately reacquire it?

edwardalee (changes requested):
Minor questions in this review.

core/federated/RTI/rti_remote.c
ChadliaJerad: It's an old leftover FIXME. Solved in this [commit](https://github.com/lf-lang/reactor-c/pull/358/commits/5558d7de47a7de14d10aa9305b1cbd28003407d4)

core/federated/RTI/rti_remote.c
ChadliaJerad: Correct. This was executed once per federate joining at startup, though harmless. This [commit](https://github.com/lf-lang/reactor-c/pull/358/commits/2729262564905ea15df234248efeb10ff8290f47) makes it happen only once.

core/federated/RTI/rti_remote.c
ChadliaJerad: That's a very very thorough catch! Many thanks!

At some point we made the decision not to "buffer" messages when a transient is absent, because we're not meant to know when, or whether, it will rejoin, and we didn't want to risk an unbounded queue.

When a transient leaves, the RTI empties its `in_transit_message_tags` (`reset_transient_federate()`). And while it is `NOT_CONNECTED`, nothing gets enqueued for it at all and the message is just dropped.

A subtle corner case is a transient that has already connected but hasn't started executing yet, that's the second clause of the same check, (`lf_tag_compare(lf_delay_tag(intended_tag, delay), fed->effective_start_tag) < 0`).
But even in this case, effective_start_tag has already been computed by that point. 

This being said, if there's no after delay on the connection, the mechanism works fine, with a useless Condition 5 (because logical time chases physical time). But it's true that in the presence of a delay on the connection between a transient and its upstream, this breaks: the tag that matters for `effective_start_tag` purposes isn't the upstream's own tag, it's that tag plus the connection delay (`lf_delay_tag(intended_tag, delay)` adjustment the drop check itself applies). 
Condition 5 can end up with a wrong effective_start_tag by the delay amount.

The proposed fix (implemented [here](https://github.com/lf-lang/reactor-c/pull/358/commits/5c1c42606dd3a1700b73963946c2dabebe6c24d4)) replaces Condition 5's dependence on `upstream->in_transit_message_tags` with a new field on `federate_info_t` (`max_intended_tag`), and:

- it is initialized to `NEVER_TAG`.
- updated in `handle_timed_message()` every time the RTI receives a message addressed to this federate, before the drop/forward branch (with the max). We store the  already-delay-adjusted tag. This update is regardless of whether the message ends up forwarded or dropped, so it captures exactly what the current mechanism misses.
- This filed is not reset by `reset_transient_federate()` on disconnect: it needs to persist across the absence window, which is the whole point.
- Condition 5 is updated to account for the new parameter only.

core/federated/RTI/rti_remote.c
ChadliaJerad: No. The `if` branch handles the hot-swap case, the `else` handles a normal transient connection. The message belongs in the `if` only.

core/federated/RTI/rti_remote.c
ChadliaJerad: Good catch! Solved [here](https://github.com/lf-lang/reactor-c/pull/358/commits/30bf21e830124d681277076dad7adc38f58739ba).

core/federated/RTI/rti_remote.c
ChadliaJerad: This part was refactored. Let I'll double check the logic.

ChadliaJerad: The currently implemented logic in `receive_connection_information()` handles 3 different cases, distinguished by validate and scratch:
- Newly joining federate (persistent or transient): `validate == false` and `scratch == NULL`.
- Hot swap: `validate == true` and `scratch == NULL`.
- Rejoining transient: `validate == true` and `scratch != NULL`.

In the hot-swap case, `fed` points to `hot_swap_federate`, and that's where the incoming connection information gets stored. Since `hot_swap_federate` was malloc'd and initialized the normal way (via `initialize_federate()`), its `neighbor_structure_received` starts out `false`. That's why it has to be explicitly set once validation succeeds. It's this same struct that later gets installed as the RTI's pointer for that federate ID (`rti_remote->base.scheduling_nodes[fed_id] = (scheduling_node_t*)hot_swap_federate`), so if the flag weren't set here, the new instance would take over with `neighbor_structure_received == false`, and any subsequent rejoin or hot swap for that federate ID would incorrectly skip validation against the established reference.

edwardalee (approved):
Finally! LGTM!

PR #364, opened 2024-02-23
No use of C11 threads on Windows/Mac

petervdonovan:
This branch is an attempt to address the failing Windows tests on `main`.

### DIFF
--- a/.github/scripts/check-diff.sh
+++ b/.github/scripts/check-diff.sh
@@ -4,8 +4,8 @@ changes() {
   git diff --name-only HEAD $(git merge-base HEAD origin/main)
 }
 
-if changes | grep "$2" | grep -q -v "^.*md\|txt$"; then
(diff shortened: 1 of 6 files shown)


.github/scripts/check-diff.sh
lhstrh, line 1: Whoops. That was my silly mistake.

petervdonovan:
OK so now this PR is blocked by the fact that the PR that we just merged is rejected by AppleClang. I guess I jumped the gun in marking ready for review.

lhstrh:
> OK so now this PR is blocked by the fact that the PR that we just merged is rejected by AppleClang. I guess I jumped the gun in marking ready for review.

Clang it!

lhstrh (approved):
These fixes look good to me! Apologies for the bugs I left in the check diff logic...

PR #367, opened 2024-02-26
Static tracing

erlingrj:
Traces the logical tag of the EXE instruction.

### DIFF
--- a/core/threaded/scheduler_static.c
+++ b/core/threaded/scheduler_static.c
@@ -230,13 +230,13 @@ void execute_inst_DU(lf_scheduler_t* scheduler, size_t worker_number, operand_t 
  */
 void execute_inst_EXE(lf_scheduler_t* scheduler, size_t worker_number, operand_t op1, operand_t op2, operand_t op3, size_t* pc,
     reaction_t** returned_reaction, bool* exit_loop) {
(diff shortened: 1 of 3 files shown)


lsk567:
Thanks for working on this @erlingrj . I share your concern and agree that using the original tracepoint_reaction_starts is a more robust design.

I like that you observed that `op2` (this is what you mean, right? not `op1`) is always used when executing a reaction because `op2` is an argument (more specifically, a pointer to the reactor self struct) passed into the function pointer (`op1`). But since we also use `EXE` to execute helper functions, and those helper functions may also expect arguments in the future, perhaps checking `op2 == NULL` will not be sufficient in the future.

How about we modify the semantics of `EXE` to use the currently unused `op3`, and make `op3` indicate whether `op1` is a reaction or not? We can pass in a zero when `op1` is a reaction and a non-zero when op1 is a helper function. That way we can use the appropriate tracepoint functions in `execute_inst_EXE`.

lsk567:
Instead of setting `op3` to either 1 or 0, now I went with the route of setting it to a reaction number, which helps the tracing function to identify which reaction is invoked. The tracing tool seems to work properly.

erlingrj:
Thanks for finishing this @lsk567

PR #368, opened 2024-02-26
Fix tagged message length from `int32_t` to `uint32_t`

Jakio815:
The tagged message sends a message length of int32_t. The length should not be negative, so changing it to uint32_t.

It also makes the maximum size of the tagged message from 2GB to 4GB.

By the way, the typescript target already writes in unsigned int 32 type.
```
msg.writeUInt32LE(value.length, 5);
```


edwardalee (approved):
LGTM!  Perhaps the comment in net_common.h documentation for `MSG_TYPE_TAGGED_MESSAGE` should change as follows: It currently reads:
```
 * The four bytes after that will be the length of the message.
```
Perhaps it should read:
```
 * The four bytes after that will be the length of the message (as an unsigned 32-bit int).
```

PR #369, opened 2024-02-26
Add address query reply message

Jakio815:
In decentralized coordination, the federate sends MSG_TYPE_ADDRESS_QUERY to the RTI with the payload being the ID of the federate it wishes to connect to.

Then, the RTI will reply with the remote federate's port number and IP address, using the same message header MSG_TYPE_ADDRESS_QUERY. 

I suggest the reply message header to be changed from MSG_TYPE_ADDRESS_QUERY to MSG_TYPE_ADDRESS_QUERY_REPLY.

I just pushed the MSG_TYPE number down. Since 18 was empty, didn't need to push down all the numbers. Future work will to make it in enums.

Since this changes the protocol after this merge, the RTI must be rebuilt.


include/core/federated/network/net_common.h
```diff
@@ -525,12 +525,20 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * Byte identifying a address query message, sent by a federate to RTI
  * to ask for another federate's address and port number.
  * The next two bytes are the other federate's ID.
- * The reply from the RTI will a port number (an int32_t), which is -1
+ */
+#define MSG_TYPE_ADDRESS_QUERY 13
+
+/**
+ * Byte identifying a address query message reply, sent by a RTI to federate
+ * to reply for requested remote federate's address and port number.
```
edwardalee, line 533:
```suggestion
 * Byte identifying a address query message reply, sent by a RTI to a federate
 * to reply with a remote federate's address and port number.
```

edwardalee (approved):
LGTM!  Minor suggestions on comment wording...

cmnrd:
Is there a plan on how we manage the updating of the RTI? How do we tell this to users? How do we even know if the running RTI is using the right version? Perhaps we shouldn't expect the RTI to be installed in the system until we have a proper versioning system. We can simply compile it alongside the federates and also place it in the `bin` directory.

PR #374, opened 2024-02-29
Save trace file for RTI in abnormal termination

### LINKED ISSUE #373: The RTI tacing file is not saved when exiting a program by `ctrl+c`
When I run [`FeedbackDelay5.lf`](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/FeedbackDelay5.lf) in [`remove_LTC`](https://github.com/lf-lang/reactor-c/pull/372) branch (with `master` of `lingua-franca`), the program gets stuck and I need to press `ctrl+c` to stop the program. In this case, `rti.lft` is empty although the log says the RTI trace file is saved. The RTI file should not be empty as it already sends and receives a lot of messages. @petervdonovan Could you take a look at this problem?

image

petervdonovan:
This is a quick cleanup for a mistake in my recent tracing PR. Thanks to @byeong-gil for noticing this issue so quickly.

I tested this PR locally on FeedbackDelay5 when `Ctrl+C`'ing the program early. It seemed to fix the problem and create a non-empty RTI trace file.

Closes #373.

### DIFF
--- a/core/federated/RTI/main.c
+++ b/core/federated/RTI/main.c
@@ -108,6 +108,7 @@ void termination() {
             send_failed_signal(f);
         }
         if (rti.base.tracing_enabled) {
+            lf_tracing_global_shutdown();
             lf_print("RTI trace file saved.");
         }
         lf_print("RTI is exiting abnormally.");


lhstrh (approved):
:rocket:

byeong-gil:
Thanks for the quick fix, @petervdonovan!

PR #375, opened 2024-02-29
Run one worker on main thread

sberkun:
Followup to #344

On embedded platforms, we have a very limited number of threads (i.e. just two on the rp2040). Therefore, we would like the main thread to run a worker, rather than just sitting idle while other threads run workers. This PR proposes the simplest way of doing so.

One alternate approach is to make a list of threads, and a function that runs them all at once (essentially separating out the fork/join logic). I think it would look cleaner, but would involve a significant amount of extra code to create a "thread list" type. Since this "thread list" type would only be used once, I'm not sure if its worth the tradeoff.

Open to any other approaches as well.

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -1026,17 +1026,6 @@ void lf_print_snapshot(environment_t* env) {
     }
 }
 
-// Start threads in the thread pool.
-void start_threads(environment_t* env) {
-    assert(env != GLOBAL_ENVIRONMENT);
-
-    LF_PRINT_LOG("Starting %u worker threads in environment", env->num_workers);
-    for (unsigned int i = 0; i < env->num_workers; i++) {
-        if (lf_thread_create(&env->thread_ids[i], worker, env) != 0) {
-            lf_print_error_and_exit("Could not start thread-%u", i);
-        }
-    }
-}
 
 /**
  * @brief Determine the number of workers.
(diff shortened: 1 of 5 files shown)


core/threaded/reactor_threaded.c
```diff
@@ -1161,23 +1150,48 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
...
+            if (i == 0 && j == 0) {
+                // main thread worker
+                worker_thread_exit_status = main_thread_exit_status;
+            } else {
+                int failure = lf_thread_join(env->thread_ids[j], &worker_thread_exit_status);
+                if (failure) {
+                    lf_print_error("Failed to join thread listening for incoming messages: %s", strerror(failure));
+                }
+            }
         	if (worker_thread_exit_status != NULL) {
```
edwardalee, line 1193:
```suggestion
            if (worker_thread_exit_status != NULL) {
```

core/threaded/reactor_threaded.c
```diff
@@ -1161,23 +1150,48 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
...
-                lf_print_error("---- Worker %d reports error code %p", i, worker_thread_exit_status);
+                lf_print_error("---- Worker %d reports error code %p", j, worker_thread_exit_status);
                 ret = 1;
         	}
```
edwardalee, line 1196:
```suggestion
            }
```

edwardalee (approved):
This looks good to me.  If all tests pass and conflicts are resolved, I don't see any reason to not merge this.

lhstrh:
@sberkun, it looks like this PR has stalled and is at risk of falling behind (esp. after merging #384). Shall we try to get this merged soon? There are some unresolved issues left to address...

sberkun:
I was going to wait for #344 to get merged first, but since these PRs are relatively independent I'll just rebase these changes.

sberkun:
Replaced by #437

PR #376, opened 2024-03-01
Print more info when assertion fails

erlingrj:
Also print __FILE__, __LINE__ and __func__ so we now exactly where an assertion fails.


edwardalee (approved):
LGTM!

PR #377, opened 2024-03-01
Change ADR_RQ to ADR_QR for correct line drawing in trace_svg.html

chanijjani:
This fixes the problem of ADR_QR lines not being displayed in trace_svg.html.


chanijjani:
Please add a proper label and allow reviewers. I can't configure those (maybe) due to permissions.

lhstrh:
> Hello! I'm Chanhee Lee at ASU. Please add a proper label and allow reviewers. I can't configure those (maybe) due to permissions.

Added you as an active committer to the repo!

edwardalee (approved):
Good catch!!

PR #379, opened 2024-03-02
Fix for Clang error

lhstrh:
This PR fixes a Clang error that currently exists in `main`. 

In addition, it simplifies CI by looking at labels to figure out which tests should be run. If a the PR is not properly labeled and then fails in the merge queue, it will be easy to add the appropriate labels to get the desired test output. Making this change as part of _this_ PR was prompted by the need to trigger the macOS tests to verify that the fix worked.


erlingrj (approved):
The logning macro changes look good. I can't comment on the CI changes yet, but would like to understand how I can control what tests are run in CI for my PR. Is this documented somewhere?

PR #382, opened 2024-03-03
Use a priority queue sorted by tags to manage delayed grants.

ChadliaJerad:
In `transient-fed` branch, each delayed grant is managed by an independent thread that calls `lf_sleep()`. 
This PR uses a priority queue sorted by tags instead.


ChadliaJerad:
Replaced by https://github.com/lf-lang/reactor-c/pull/394

PR #384, opened 2024-03-05
Add clang-format and run it on the code-base

erlingrj:
This PR replaces #310. It copies clang-format setup from reactor-cpp. I have added a simple Makefile target so to format reactor-c run:
```
make format
```

In CI the following check is run which returns an error if it finds anything it would like to change
```
make format-check
```

For any outstanding feature branch I suggest the following strategy:
1. Cherry-pick the commit adding clang-format: `git cherry-pick 91bc46cd17cbd12eaaa83f4684b3568341f79d6d`
2. Run the formatter on the feature branch. This must be done several times until clang-format "stabilizes" on a valid format:
```
make format
make format
```

Then hopefully there will be few conflicts


erlingrj:
Tagging the most frequent contributors to reactor-c here for your information: @petervdonovan @ChadliaJerad @lhstrh @lsk567   Anyone else that is working on major features?

edwardalee (approved):
LGTM.  Let's do this and get the pain over with.

petervdonovan (approved):
Yep, I'm on board with this. This will probably make CONTRIBUTING.md, which was never really used anyway, out of date. We should probably just delete CONTRIBUTING.md.

EDIT: This reminds me, if there is a way to do this that will make it easier to reach partial compliance with a standard like MISRA (that is, compliance with the parts that make sense for us), then that might be worth considering. But I have no expertise on that and I'm sure you've already considered that, so I'm not going to try to make a lot of noise about it.

erlingrj:
@petervdonovan regarding MISRA. I think that goes beyond formatting and is more a topic of linting? There is a static analysis/linting tool called clang-tidy that we could integrate also. I do not know a lot on this topic though.

erlingrj:
Thank you @lhstrh!

lhstrh:
> Thank you @lhstrh!

Sorry it took so long... No need for this to hold you up either. The tests are clearly passing, so it's also fine to bypass the merge queue and merge directly this time.

PR #385, opened 2024-03-05
Set core affinity of statically sheduled programs

erlingrj:
This PR incorporates the changes from this draft PR:  #355 

It will make it more challenging to merge static scheduling into main since it will likely be modified before it is merged:

TODO:
- [x] Always set core affinity of workers in increasing order.
- [x] Always set worker priority to 85


erlingrj:
@lsk567 you can try it out. Now all workers are pinned to cores and thread priority is set to 85. This requires superuser privilege, so you must run the LF program with sudo

PR #387, opened 2024-03-06
Enable all warnings and treat them as errors

erlingrj:
In this PR I enable all warnings and treat them as errors with `-Wall -Wpedantic -Werror`, this is how it is done in reactor-cpp. This triggered some refactoring of the CMake system and addressing a lot of warnings all over the code-base. I expect there to be a lot more warnings when CI run tests on Windows, macOS, Zephyr etc.

The advantage of getting this merged is that it puts more requirements on the code that is checked in. The next step is to add clang-tidy which performs static analysis and catches even more possible undefined behaviour.

Disabled on the following platforms:
- macOS: Due to our use of GNU-specific variadic arguments in the logging macros
- Zephyr: Because there are warnings inside the Zephyr kernel and we cannot avoid including those headers
- 32bit platforms: Due to a possible bug in pqueue implementation, see #388


core/CMakeLists.txt
```diff
@@ -134,13 +116,13 @@ if(DEFINED _LF_CLOCK_SYNC_ON)
...
 
-# Link with thread library, unless we are on the Zephyr platform.
-if(NOT DEFINED LF_SINGLE_THREADED OR DEFINED LF_TRACE)
-    if (NOT PLATFORM_ZEPHYR)
-        find_package(Threads REQUIRED)
-        target_link_libraries(reactor-c PUBLIC Threads::Threads)
-    endif()
-endif()
+
+# FIXME: May want these to be application dependent, hence passed as
```
lhstrh, line 120: I realize this is a copy pasta, but should we turn this into a GitHub issue?

core/threaded/scheduler_NP.c
```diff
@@ -111,12 +111,8 @@ void _lf_sched_notify_workers(lf_scheduler_t* scheduler) {
   // Calculate the number of workers that we need to wake up, which is the
```
lhstrh, line 111: The comment up here is confusing. Revise or remove?

include/core/federated/network/net_common.h
```diff
@@ -483,7 +483,6 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   do {                                                                                                                 \
     buffer[0] = MSG_TYPE_STOP_REQUEST;                                                                                 \
     encode_int64(time, &(buffer[1]));                                                                                  \
-    assert(microstep >= 0);                                                                                            \
```
lhstrh, line 486: Curious about this change. Why safe to remove?

lingua-franca-ref.txt
```diff
@@ -1 +1 @@
-master
+run-zephyr-tests
```
lhstrh, line 1: Is this correct?

lhstrh (commented):
These look like great cleanups to me. Left some minor comments/questions.

core/CMakeLists.txt
erlingrj: I made it such that it is configurable by passing flags to CMake, e.g. `cmake -Bbuild -DINITAL_EVENT_QUEUE_SIZE=15`

include/core/federated/network/net_common.h
erlingrj: Our microstep type is unsigned. I guess there is the possibility that someone pass in an integer though, since it is only a macro. I will add the assertion bacl

lingua-franca-ref.txt
erlingrj: Thanks, that has been merged so I change back to master

core/federated/RTI/rti_remote.c
```diff
@@ -1323,6 +1318,7 @@ static int receive_connection_information(int* socket_id, uint16_t fed_id) {
 
     // Allocate memory for the upstream and downstream pointers
     if (fed->enclave.num_upstream > 0) {
+      // FIXME: This looks wrong. We are casting an array of uint16_ts to an int*?
```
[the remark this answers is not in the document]
erlingrj: I fixed it. What makes me a little weary is that this should have lead to out-of-bounds writes. We essentially allocate only half of the memory we need (we allocate for uint16 but interpret it as ints). So this would lead to writing outside the memory returned by malloc. It is hard for me to believe that this would not cause weird issues...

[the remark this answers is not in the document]
erlingrj: I was able to confirm the memory error with valgrind. So thanks for pointing it out, this was a segmentation fault/memory corruption error waiting to happen

erlingrj:
@lhstrh and @edwardalee, I propose we merge this next so that the other PRs are forced to handle their warnings (and not having me do it afterwards in this PR). 

One change that I would like an extra set of eyes on is to the lf_thread_create in the C11 support. To avoid compiler warnings I had to define a wrapper function with a signature expected by the C11 (needs to return an int).

core/lf_utils.cmake
```diff
@@ -0,0 +1,11 @@
...
+    if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+        target_compile_options(${target} PRIVATE -Wall -Wextra -Wpedantic -Werror)
+    elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+        target_compile_options(${target} PRIVATE /W4)
+    elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+        target_compile_options(${target} PRIVATE -Wall)
+    else()
+        target_compile_options(${target} PRIVATE -Wall)
+    endif()
+endfunction()
```
edwardalee, line 11: This seems to omit some of commands it replaces, which are these:
```
if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
    set(CMAKE_SYSTEM_VERSION 10.0)
    message("Using Windows SDK version ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Nrf52")
    list(APPEND REACTORC_COMPILE_DEFS PLATFORM_NRF52)
elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Zephyr")
    list(APPEND REACTORC_COMPILE_DEFS PLATFORM_ZEPHYR)
    set(PLATFORM_ZEPHYR true)
elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Rp2040")
    list(APPEND REACTORC_COMPILE_DEFS PLATFORM_RP2040)
endif()
```
Just checking to make sure this was deliberate.

core/tracepoint.c
```diff
@@ -61,7 +61,7 @@ int register_user_trace_event(void* self, char* description) {
 }
 
 void call_tracepoint(int event_type, void* reactor, tag_t tag, int worker, int src_id, int dst_id,
-                     instant_t* physical_time, trigger_t* trigger, interval_t extra_delay, bool is_interval_start) {
+                     instant_t* physical_time, trigger_t* trigger, interval_t extra_delay) {
```
edwardalee, line 64: I'm not sure about this change. I thought the `is_interval_start` field was used to be able to determine the execution time of a reaction.  Are you sure we can still do that with it removed?  The thing to check is whether trace_to_chrome still displays the reaction executions as an interval.  I concur, however, that on searching for uses of this field, I can't find any, so this is probably right.

edwardalee (approved):
These changes look great to me!  Only one question in the comments.

core/CMakeLists.txt
```diff
@@ -50,23 +40,18 @@ endif()
...
-# Zephyr Cmake extension to create the library and add the sources.
-if(PLATFORM_ZEPHYR)
-    message("--- Building Zephyr library")
-    zephyr_library_named(reactor-c)
-    zephyr_library_sources(${REACTORC_SOURCES})
-    zephyr_library_link_libraries(kernel)
-else()
-    add_library(reactor-c)
-    target_sources(reactor-c PRIVATE ${REACTORC_SOURCES})
-endif()
```
petervdonovan, line 69: Just curious: where did the Zephyr-specific stuff go? To be clear, I'm happy to see it gone.

core/federated/federate.c
```diff
@@ -88,8 +88,8 @@ federate_instance_t _fed = {.socket_TCP_RTI = -1,
                             .has_upstream = false,
                             .has_downstream = false,
                             .received_stop_request_from_rti = false,
-                            .last_sent_LTC = (tag_t){.time = NEVER, .microstep = 0u},
-                            .last_sent_NET = (tag_t){.time = NEVER, .microstep = 0u},
+                            .last_sent_LTC.time = NEVER,
+                            .last_sent_NET.time = NEVER,
```
petervdonovan, line 92: Nit: Why not set the microstep to 0, just to simplify things? It's probably fine but we might as well liberate ourselves from having to reason about UB from uninitialized memory.

core/tracepoint.c
petervdonovan: Yeah, I recently removed the usage of that field as a simplification, since previously it was only used as an optimization to get extra measurement precision. We should add it back later when we have time and we realize we need it.

petervdonovan (approved):
Took a quick look and don't have any important concerns. The change in the C11 threads support looks good to me.

core/CMakeLists.txt
erlingrj: Thanks to your nice refactoring of all platform specific things into a separate target (low-level-platform) I can now do the zephyr-specific things there :D

core/tracepoint.c
erlingrj: We could keep it and add a `(void) is_interval_start` to avoid the warning

core/federated/federate.c
erlingrj: Yes I will do it. The reason for this change was that the original type of assignment was not allowed by our macOS compiler.

erlingrj: OK, went back to NEVER_TAG since we don't have Werror on macOS for other reasons

erlingrj (commented):
Did a final self-review, merging this now.

Issue #388, opened 2024-03-07
Implicit assuming 64bit architecture in pqueue implementation 


erlingrj:
As I am turning on all warnings and treating them as errors I am seeing some odd things with the pqueue implementation. Basically, there is casting between `unsigned long long` which typically are 64 bit, and `void *` which are 32bit or 64bit depending on the architecture. I am specifically talking about the static functions at the top of pqueue_tag.c

```
/**
 * @brief Callback function to get the priority of an element.
 * Return the pointer argument cast to pqueue_pri_t because the
 * element is also the priority. This function is of type pqueue_get_pri_f.
 * @param element A pointer to a pqueue_tag_element_t, cast to void*.
 */
static pqueue_pri_t pqueue_tag_get_priority(void* element) { return (pqueue_pri_t)element; }

/**
 * @brief Callback comparison function for the tag-based priority queue.
 * Return 0 if the first argument is less than second and 1 otherwise.
 * This function is of type pqueue_cmp_pri_f.
 * @param priority1 A pointer to a pqueue_tag_element_t, cast to pqueue_pri_t.
 * @param priority2 A pointer to a pqueue_tag_element_t, cast to pqueue_pri_t.
 */
static int pqueue_tag_compare(pqueue_pri_t priority1, pqueue_pri_t priority2) {
  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag) > 0);
}
```

E.g. in `pqueue_tag_get_priority` we assume that a void* can hold unsigned long long, this is not true on 32bit archs. Does anyone have a good handle on how the pqueue works? It seems that the elements are void* and that we cannot treat these as 64bit values. It suggests that we either must explicitly have elements which have the desired size (64bit) or we must make do with 32bit priorities for the 32bit platforms.

Any thoughts @lhstrh @edwardalee @petervdonovan ?

petervdonovan:
To me it looks like the priority queue is currently only ever used to hold pointers. I think it's safe to just assume that the element size in the queue is 32 bits or 64 bits according to the architecture. Maybe we should see if everything compiles without errors or warnings once we remove any casts to `unsigned long long` and any unsafe casts from `unsigned long long`.

erlingrj:
[...]

PR #390, opened 2024-03-08
Refactoring of event queue

byeonggiljun:
In this PR, the event queue stores events using `pqueue_tag` to sort them in tagged order. Consequently, events scheduled with a tag with a non-zero microstep are handled the same as other events (`next_q` is deleted).

Some functionalities in `pqueue_tag.c` and `pqueue_base.c` are also changed.

The motivation for this refactoring is:
* to prevent a federate sends NET (0, 1) when the federate has received T_MSG (0, 0). Normally, we can infer that the next event's tag is one microstep later than the current tag if the next event's time is the same as the current time. For example, if a next event's time is `1 sec` and the current tag is `(1 sec, 1)`, that event needs to be executed at `(1 sec, 2)` because every event at `(1 sec, 1)` must be popped before advancing to `(1 sec, 1)`. The code below (in `get_next_event_tag`) shows that inference. However, an exception occurs at the start tag. At the start tag `(0, 0)`, the event queue can contain an event with `(0, 0)`. We still infer that the event's tag is `(0, 1)` and send `NET (0, 1)`, which is wrong. If events are sorted by tag, not time, we can use each event's tag as the target tag.

https://github.com/lf-lang/reactor-c/blob/7427d987d725111141139819c5341921edd0fef1/core/threaded/reactor_threaded.c#L290-L293

* to make the scheduling process simpler. Because the events are sorted by time, not tag, we had to put chained dummy events to implement an event with a positive microstep. After this refactoring, we can schedule events with a tag because events are sorted by their tag. Thus, we can remove those chained dummy events and the next queue.


byeonggiljun:
Convert this to draft because mac-os tests are failing.

petervdonovan:
The test that is failing has both decentralized execution and physical actions. It is very hard to make such a test both useful (in the sense of catching bugs) and non-flaky. In particular, [DistributedLoopedPhysicalActionDecentralized](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/DistributedLoopedPhysicalActionDecentralized.lf) checks that messages from the sender are processed in the correct order, which won't be true if there are STP violations.

I don't have a strong opinion about how to fix this. We have discussed pipe-dream kinds of improvements to the testing framework that could solve this sort of thing, but in the near term, I guess we can choose from a few unappealing options, such as moving the test to `failing`, not checking that messages are received in the right order, or deferring the check until the end after the tardy message has been received. I suppose I would lean toward the second option. The third of these could be rather complex, and it is already difficult to interpret the tests.

byeonggiljun:
Thank you, @petervdonovan! I also agree that the test doesn't have to check the order of messages. I'll make a simple LF PR if it's the reason for the error.

lib/schedule.c
```diff
@@ -298,16 +292,20 @@ trigger_handle_t lf_schedule_trigger(environment_t* env, trigger_t* trigger, int
...
-
-  // Do not schedule events if if the event time is past the stop time
-  // (current microsteps are checked earlier).
-  LF_PRINT_DEBUG("Comparing event with elapsed time " PRINTF_TIME " against stop time " PRINTF_TIME ".",
-                 e->time - lf_time_start(), env->stop_tag.time - lf_time_start());
-  if (e->time > env->stop_tag.time) {
-    LF_PRINT_DEBUG("lf_schedule_trigger: event time is past the timeout. Discarding event.");
+  e->base.tag = intended_tag;
+
+  // Do not schedule events if if the event time is past the stop tag.
```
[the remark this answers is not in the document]
byeonggiljun: Fixed in [0200a84](https://github.com/lf-lang/reactor-c/pull/390/commits/0200a84418299a0eb5ea5c35626e15c0dd6630e2)

core/utils/pqueue_tag.c
```diff
@@ -33,7 +33,7 @@ static pqueue_pri_t pqueue_tag_get_priority(void* element) { return (pqueue_pri_
  * @param priority2 A pointer to a pqueue_tag_element_t, cast to pqueue_pri_t.
  */
 static int pqueue_tag_compare(pqueue_pri_t priority1, pqueue_pri_t priority2) {
-  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag) > 0);
+  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag));
```
edwardalee, line 36: This behavior no longer corresponds to the comment describing the function.

edwardalee (commented):
I need to review this more carefully, but I have one high-level question:
You changed the semantics of the `cmppri` comparison functions.
They used to return 0 for a < b and non-zero otherwise.  You now have them returning -1 for a < b, 0 for a==b, and 1 for a > b.

Why did you make this change?  The pqueue implementation doesn't need this, I think, and this change is what resulted in most of the changes in the base classes.  Is this change necessary?

byeonggiljun:
>Why did you make this change? The pqueue implementation doesn't need this, I think, and this change is what resulted in most of the changes in the base classes. Is this change necessary?

Thank you for starting the review, @edwardalee! I have a reason for that change although I'm not sure that change is the only solution.

Please look at the code below. This is a part of `find_equal_same_priority` in `pqueue_base.c`. This function is needed here ([lib/schedule.c#L201-L205](https://github.com/lf-lang/reactor-c/blob/ec06c930a65d05420d103a6773fca8a442d58124/lib/schedule.c#L201-L205)) to check if there is an already scheduled event with the same trigger and the same tag. 

https://github.com/lf-lang/reactor-c/blob/ec06c930a65d05420d103a6773fca8a442d58124/core/utils/pqueue_base.c#L100

We use the `==` operator to check whether two priorities are identical. This works for the time-ordered queue. Nonetheless, `pqueue_tag`'s priority is a pointer to a tag and the `==` operator does not work. Also, currently the `cmp_pri (thiz, that)` function only checks whether `thiz > that` or not. This is why I make the `cmp_pri` function return 0 if two priorities are the same. As a result, we can check the equality of the two priorities as below.
https://github.com/lf-lang/reactor-c/blob/0200a84418299a0eb5ea5c35626e15c0dd6630e2/core/utils/pqueue_base.c#L133

core/utils/pqueue_tag.c
byeonggiljun: Resolved in [1bd320f](https://github.com/lf-lang/reactor-c/pull/390/commits/1bd320f000f6a9010c2cf90be7bd7cdcb55d5112).

core/utils/pqueue.c
edwardalee, line 1: You should add yourself as an author.  Also, in both this file and in `pqueue.h`, the file comment is no longer accurate.  I suggest replacing with this:
```
 * @brief Priority queue definitions for queues where the priority is a number that can be compared with ordinary numerical comparisons.
 *
 * This is used for the reaction queue.  The event queue uses a `tag_t` struct for its priority, so it cannot use this.

File: core/utils/pqueue.c [hunk 3]
```diff
@@ -25,15 +25,15 @@ int event_matches(void* event1, void* event2) { return (((event_t*)event1)->trig
 
 int reaction_matches(void* a, void* b) { return (a == b); }
 
-pqueue_pri_t get_event_time(void* event) { return (pqueue_pri_t)(((event_t*)event)->time); }
+pqueue_pri_t get_event_time(void* event) { return (pqueue_pri_t)(((event_t*)event)->base.tag.time); }
```
edwardalee, line 28: This function can be removed.  It's only use (in `environment.c`) can be replaced with the `getpri` function used by the event_q, `pqueue_tag_get_priority`.

include/core/utils/pqueue_tag.h
edwardalee, line 1: The `@file` tag at the start of this file does not match the filename. I'm not sure whether this matters, but may as well fix it..

include/core/utils/pqueue_tag.h
```diff
@@ -69,6 +69,17 @@ typedef pqueue_t pqueue_tag_t;
  */
 pqueue_tag_t* pqueue_tag_init(size_t initial_size);
 
+/**
+ * @brief Create a priority queue sorted by tags and has a payload that requires a particular comparison function.
```
edwardalee, line 73:
```suggestion
 * @brief Create a priority queue sorted by tags and has a payload that requires a particular equivalence checker function.
```

include/core/utils/pqueue_tag.h
```diff
@@ -124,6 +135,14 @@ int pqueue_tag_insert_if_no_match(pqueue_tag_t* q, tag_t t);
  */
 pqueue_tag_element_t* pqueue_tag_find_with_tag(pqueue_tag_t* q, tag_t t);
 
+/**
+ * @brief Return the first item with the same tag that matches the supplied element or NULL if there is none.
```
edwardalee, line 139:
```suggestion
 * @brief Return an item with the same tag (`cmppri` returns 0) that matches the supplied element
 * (`eqelem` returns non-zero) or NULL if there is none.
```

lib/schedule.c
```diff
@@ -149,17 +149,15 @@ trigger_handle_t lf_schedule_trigger(environment_t* env, trigger_t* trigger, int
   if (!trigger->is_timer) {
     delay += trigger->offset;
   }
-  tag_t intended_tag = (tag_t){.time = env->current_tag.time + delay, .microstep = 0};
+  tag_t intended_tag = lf_delay_tag(env->current_tag, delay);
+  printf("intended tag = " PRINTF_TAG ".\n", intended_tag.time, intended_tag.microstep);
```
edwardalee, line 153: This looks like a debug statement. It should be deleted or replaced with `LF_PRINT_DEBUG`.

include/core/reactor_common.h
```diff
@@ -198,13 +198,10 @@ void _lf_trigger_shutdown_reactions(environment_t* env);
...
- * @param time The logical time of that event.
- * @param next The event to place after the dummy events.
- * @param offset The number of dummy events to insert.
- * @return A pointer to the first dummy event.
+ * @param tag The tag of that event.
+ * @return A pointer to the dummy event.
  */
-event_t* _lf_create_dummy_events(environment_t* env, trigger_t* trigger, instant_t time, event_t* next,
-                                 microstep_t offset);
+event_t* _lf_create_dummy_events(environment_t* env, trigger_t* trigger, tag_t tag);
```
edwardalee, line 204: Given that these dummy events have a difference use now, I suggest replacing the function description as follows:
```
 * @brief Create a dummy event with the specified tag.
 * 
 *  A dummy event is an event with no triggers that can be put on the event queue to trigger a tag advance to the specified tag.
```

include/core/lf_types.h
```diff
@@ -195,15 +196,13 @@ typedef struct event_t event_t;
...
 struct event_t {
-  instant_t time;     // Time of release.
-  trigger_t* trigger; // Associated trigger, NULL if this is a dummy event.
-  size_t pos;         // Position in the priority queue.
-  lf_token_t* token;  // Pointer to the token wrapping the value.
-  bool is_dummy;      // Flag to indicate whether this event is merely a placeholder or an actual event.
+  pqueue_tag_element_t base; // Elements of pqueue_tag. It contains tag of release and position in the priority queue.
+  trigger_t* trigger;        // Associated trigger, NULL if this is a dummy event.
+  lf_token_t* token;         // Pointer to the token wrapping the value.
+  bool is_dummy;             // Flag to indicate whether this event is merely a placeholder or an actual event.
```
edwardalee, line 202: Maybe this is_dummy is not needed anymore.  Could just use` trigger == NULL` to signal a dummy event.

core/utils/pqueue_tag.c
```diff
@@ -27,13 +27,14 @@ static pqueue_pri_t pqueue_tag_get_priority(void* element) { return (pqueue_pri_
...
- * Return 0 if the first argument is less than second and 1 otherwise.
+ * Return -1 if the first argument is less than second, 0 if the two arguments are the same,
+ * and 1 otherwise.
  * This function is of type pqueue_cmp_pri_f.
  * @param priority1 A pointer to a pqueue_tag_element_t, cast to pqueue_pri_t.
  * @param priority2 A pointer to a pqueue_tag_element_t, cast to pqueue_pri_t.
  */
 static int pqueue_tag_compare(pqueue_pri_t priority1, pqueue_pri_t priority2) {
-  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag) > 0);
+  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag));
```
edwardalee, line 37: I'm confused: Shouldn't a _lower_ tag have _higher_ priority?

core/utils/pqueue_tag.c
```diff
@@ -149,3 +158,5 @@ void pqueue_tag_remove_up_to(pqueue_tag_t* q, tag_t t) {
     head = pqueue_tag_peek_tag(q);
   }
 }
+
+void pqueue_tag_dump(pqueue_tag_t* q) { pqueue_dump((pqueue_t*)q, pqueue_tag_print_element); }
```
edwardalee, line 162: Suggest a trailing newline:
```suggestion
void pqueue_tag_dump(pqueue_tag_t* q) { pqueue_dump((pqueue_t*)q, pqueue_tag_print_element); }

```

edwardalee (approved):
This is a very nice piece of work!  I love that quite a bit complicated code is removed.  I've made some suggestions for further code removal and some comment updates. All of these are optional.

core/utils/pqueue_support.h
```diff
@@ -61,7 +61,7 @@ static int reaction_matches(void* next, void* curr) { return (next == curr); }
  * Report a priority equal to the time of the given event.
  * Used for sorting pointers to event_t structs in the event queue.
  */
-static pqueue_pri_t get_event_time(void* a) { return (pqueue_pri_t)(((event_t*)a)->time); }
+static pqueue_pri_t get_event_time(void* a) { return (pqueue_pri_t)(((event_t*)a)->base.tag.time); }
```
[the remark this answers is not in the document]
byeonggiljun: I totally agree with you. I deleted this file in [3b50275](https://github.com/lf-lang/reactor-c/pull/390/commits/3b5027518529609a20ec4ad0cf3be1cf3c25ed02).

include/core/utils/pqueue_base.h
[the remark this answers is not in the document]
byeonggiljun: Resolved in [3b50275](https://github.com/lf-lang/reactor-c/pull/390/commits/3b5027518529609a20ec4ad0cf3be1cf3c25ed02).

include/core/utils/pqueue_base.h
```diff
@@ -140,23 +140,31 @@ void* pqueue_pop(pqueue_t* q);
 void pqueue_empty_into(pqueue_t** dest, pqueue_t** src);
 
 /**
- * Find the highest-ranking item with the same priority that matches the
- * supplied entry.
+ * Find the highest-ranking item with priority up to and including the given
+ * maximum priority that matches the supplied entry.
```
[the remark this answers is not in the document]
byeonggiljun: I removed this function in [3b50275](https://github.com/lf-lang/reactor-c/pull/390/commits/3b5027518529609a20ec4ad0cf3be1cf3c25ed02).

core/utils/pqueue_base.c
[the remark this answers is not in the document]
byeonggiljun: Done in [3b50275](https://github.com/lf-lang/reactor-c/pull/390/commits/3b5027518529609a20ec4ad0cf3be1cf3c25ed02).

core/utils/pqueue.c
byeonggiljun: To remove some more functions e.g., `set_event_position` and `get_event_position`, I made `pqueue_tag_init_customize` more flexible and used it to initialize `recycle_q` in [4b8d42d](https://github.com/lf-lang/reactor-c/pull/390/commits/4b8d42dd65662087ae0b3db77a5cb844ae115446). Does this change make sense?

core/utils/pqueue.c
byeonggiljun: I changed the comment as you suggested in [4b8d42d](https://github.com/lf-lang/reactor-c/pull/390/commits/4b8d42dd65662087ae0b3db77a5cb844ae115446). However, some functions like `print_event` and `event_matches` are still used in the event queue. So, how about we change the comment to the one below?

```
 * @brief Priority queue definitions for the event queue and reaction queue.
 *
 * This is mostly used for the reaction queue. The event queue uses a `tag_t` struct for its priority, so it cannot use this to compare the element's priority. 
```

include/core/utils/pqueue_tag.h
byeonggiljun: This is resolved in [5c966b5](https://github.com/lf-lang/reactor-c/pull/390/commits/5c966b50a131053c738a93ffd7329b020277aa49).

lib/schedule.c
byeonggiljun: Oh, I made a mistake here! Thank you so much for capturing it. Resolved in [5c966b5](https://github.com/lf-lang/reactor-c/pull/390/commits/5c966b50a131053c738a93ffd7329b020277aa49).

include/core/reactor_common.h
byeonggiljun: I replaced the comment as you suggested in [5c966b5](https://github.com/lf-lang/reactor-c/pull/390/commits/5c966b50a131053c738a93ffd7329b020277aa49).

include/core/lf_types.h
byeonggiljun: I removed the flag and used `trigger == NULL` to check whether an event is a dummy event in [739b696](https://github.com/lf-lang/reactor-c/pull/390/commits/739b696d0728fd4cf79ea3b5749c82984d9aaf57).

core/utils/pqueue_tag.c
byeonggiljun: Yes, a _lower_ tag has to have _higher_ priority. 

In the code below, if a parent node's priority is greater than the moving node's priority, we swap them. If a moving node's index is `i`, the parent node's index is `i >> 1`. In other words, a _lower_ index element (_higher_ priority element) has a _lower_ tag.

https://github.com/lf-lang/reactor-c/blob/739b696d0728fd4cf79ea3b5749c82984d9aaf57/core/utils/pqueue_base.c#L176-L190

Also, this code aligns with the current comparison function. It also returns 1 if thiz > that.
https://github.com/lf-lang/reactor-c/blob/ec06c930a65d05420d103a6773fca8a442d58124/core/utils/pqueue.c#L16

core/utils/pqueue.c
edwardalee: The `print_event` and `event_matches` functions both refer to the `event_t`, but nothing in `pqueue.c` assumes that an entry is an `event_t`.  I would suggest moving these two functions to `environment.c` and making them static.

byeonggiljun: That sounds perfect! I moved those functions to `environment.c` in [d10375d](https://github.com/lf-lang/reactor-c/pull/390/commits/d10375d34f39395ed0a6c760e75826a76cf213fe).

byeonggiljun:
Seems this PR is ready to be merged. Let me put this into the merge queue.

core/utils/pqueue_tag.c
```diff
@@ -75,11 +64,21 @@ static void pqueue_tag_print_element(void* element) {
 //////////////////
 // Functions defined in pqueue_tag.h.
 
+int pqueue_tag_compare(pqueue_pri_t priority1, pqueue_pri_t priority2) {
+  return (lf_tag_compare(((pqueue_tag_element_t*)priority1)->tag, ((pqueue_tag_element_t*)priority2)->tag));
```
erlingrj, line 68: Here an `unsigned long long` is casted to an `event_t *` this will result in warnings on any architecture where `sizeof(void *) != sizeof(unsigned long long)`

erlingrj (commented):
Hi @byeonggiljun and @edwardalee, I have a few questions about this. This seems like a good design, but there is something weird about how the priority queue is implemented where we are forced to cast between points and integers and it is not really portable.

I think we need to have `pqueue_pri_t` as a `void *`, this would work with no problem for the pqueue_tag.c as you have implemented it here, but for the pqueue.c that is used for the reaction queue, it would not work directly. But it should also adapt this design where reactions are sorted based on the values of their fields and never based on the integer values of the pointers.

core/utils/pqueue.c
byeonggiljun: Hi, @erlingrj. the `event_matches` and `print_event` functions were moved to `environment.c` based on this conversation.

`pqueue.c` only has functions for `reaction_q` so I don't think we can put the functions there. Maybe `pqueue_tag.c` is a better place for the functions. Could you share your opinion on this, @edwardalee?

core/environment.c
```diff
@@ -38,6 +38,31 @@
...
+ * @brief Callback function to print information about an event.
+ * This function is used by event queue and recycle.
+ * @param element A pointer to an event.
+ */
+static void print_event(void* event) {
+  event_t* e = (event_t*)event;
+  LF_PRINT_DEBUG("tag: " PRINTF_TAG ", trigger: %p, token: %p", e->base.tag.time, e->base.tag.microstep,
+                 (void*)e->trigger, (void*)e->token);
+}
+
```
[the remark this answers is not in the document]
edwardalee: The rationale for this is to avoid creating unnecessary dependencies.
Neither pqueue.c nor pqueue_tag.c have any dependence on event_t.
They are more generic.  It is in evironment.c that pqueue_tag is adapted to contain items of type event_t.  The way it does that is by providing these two functions as arguments to pqueue_tag_init_customize.
These functions are made static, so this design of the event queue is entirely defined in environment.c.

core/utils/pqueue_tag.c
edwardalee: I agree that pqueue_base should be changed so pqueue_pri_t is a void*.  This, however, requires redesigning the reaction queue, which has a (probably premature) optimization that combines the deadline and the level into a single index that it uses for ordering reactions.  I think this should be done as a separate PR.  Do you want to open an issue?  Probably we should benchmark the performance impact.  There is likely to be a cost.

core/utils/pqueue.c
edwardalee: (Comments crossed) As stated above, pqeueue_tag.c has no dependence on event_t nor any other definitions in lf_types.h.  Putting these functions into pqueue_tag.c would create a dependence, and they don't really belong there because there are many ways to specialize pqueue_tag, and having it contain event_t is just one of those.

byeonggiljun: I perfectly understand. Thank you so much for clarifying this!

core/environment.c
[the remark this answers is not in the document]
erlingrj: OK I see, thanks for the explanation.

core/utils/pqueue_tag.c
erlingrj: I think it is covered by this issue: https://github.com/lf-lang/reactor-c/issues/388

PR #391, opened 2024-03-11
Fix tagged message length from int32_t to uint32_t

Jakio815:
This is a fix of #368.

The tagged message sends a message length of int32_t. The length should not be negative, so changing it to uint32_t.

It is both applied on MSG_TYPE `MSG_TYPE_TAGGED_MESSAGE` and `MSG_TYPE_P2P_TAGGED_MESSAGE`.

It also makes the maximum size of the tagged message from 2GB to 4GB.

By the way, the typescript target already writes in unsigned int 32 type.

`msg.writeUInt32LE(value.length, 5);`


edwardalee (approved):
LGTM.

PR #392, opened 2024-03-11
Add address query reply message

Jakio815:
This PR reopens #369 

In decentralized coordination, the federate sends MSG_TYPE_ADDRESS_QUERY to the RTI with the payload being the ID of the federate it wishes to connect to.

Then, the RTI will reply with the remote federate's port number and IP address, using the same message header MSG_TYPE_ADDRESS_QUERY.

I suggest the reply message header to be changed from MSG_TYPE_ADDRESS_QUERY to MSG_TYPE_ADDRESS_QUERY_REPLY.

I just pushed the MSG_TYPE number down. Since 18 was empty, didn't need to push down all the numbers. Future work will to make it in enums.

Since this changes the protocol after this merge, the RTI must be rebuilt.


edwardalee (approved):
LGTM.  This will create an incompatibility in the RTI, for which we have no versioning, but we will just have to deal with that.

PR #394, opened 2024-03-15
Transient federates: Use a priority queue sorted by tags to manage delayed grants.

ChadliaJerad:
In `transient-fed` branch, each delayed grant is managed by an independent thread that calls `lf_sleep()`. This PR uses instead a priority queue sorted by tags.

This PR replaces [#382](https://github.com/lf-lang/reactor-c/pull/382), and replays its commits after the rebase of `transient-fed` (applying the new formatting).


cmnrd (changes requested):
This pqueue implementation looks extremely dangerous to me. In particular, the cast operations seem to rely on undefined behavior. Could you explain why this wrapper around the tag pqueue is needed and why the casts are safe?

edwardalee:
The casts can be avoided, but only through code duplication.  Is this really what we want?

ChadliaJerad:
In this implementation, one single thread (`lf_delayed_grants_thread()`) manages all the delayed grants, as opposed to using a max of two separate threads *per persistent federate* that has a transient upstream federate.

Instead of re-implementing from scratch, this alternative builds on top of the existing mechanism for using a priority queue sorted by tags `pqueue_tag`. This type of queue is used in `rti_remote` to manage the `in_transit_message_tags` of a federation structure. Delayed grant elements are customized from `pqueue_tag` elements as indicated [here](https://github.com/lf-lang/reactor-c/blob/main/include/core/utils/pqueue_tag.h#L37), which guarantees alignment, and thus the safe casting of the elements.

The functions implementation follows the pattern in [pqueue_tag.c](https://github.com/lf-lang/reactor-c/blob/91bc46cd17cbd12eaaa83f4684b3568341f79d6d/core/utils/pqueue_tag.c), where wrappers and casts are used as well. 

Does this make sense, @cmnrd?

cmnrd:
> The casts can be avoided, but only through code duplication. Is this really what we want?

Usually, there are more options than using code duplication or potentially dangerous casts. To be clear, I am not opposing casts in general, but the cast from `pqueue_delayed_grant_element_t`   to `pqueue_tag_elemen_t` is not safe at first glance. I see now how this works by having a `pqueue_tag_elemen_t` as a first element in `pqueue_delayed_grant_element_t`. I am still not fully convinced that this actually is safe because as far as I know, the compiler is allowed to insert padding and alignment as needed.

A better and more general solution, however, would be to use a generic pqueue implementation that uses `void*`. In fact, this is already implemented and the tag queue already utilizes this base queue. So why build the new pqueue on the tag pqueue and have two layers of casts instead of building on the base pqueue directly?

edwardalee:
Merging this into transient-fed branch so we are working with just one branch and one PR.

PR #399, opened 2024-03-22
Update RTI dockerfile to match file layout change

petervdonovan:
This could harm reproducibility and could increase the size of the RTI image. However, reproducibility is actually not much worse than the status quo which was still copying whole directories. Also, the amount of extra stuff copied is small. In the long run this will keep us from running into further problems as we continue to refactor this repo without testing the dockerized RTI build in CI (I am assuming that it is not worthwhile to add a test for it in CI)

I have manually verified that this does build locally.

### DIFF
--- a/core/federated/RTI/rti.Dockerfile
+++ b/core/federated/RTI/rti.Dockerfile
@@ -1,7 +1,6 @@
 # Docker file for building the image of the rti
 FROM alpine:latest
-COPY core /lingua-franca/core
-COPY include /lingua-franca/include
+COPY . /lingua-franca
 WORKDIR /lingua-franca/core/federated/RTI
 RUN set -ex && apk add --no-cache gcc musl-dev cmake make && \
     mkdir container && \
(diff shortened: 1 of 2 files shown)


erlingrj:
Hm, @lhstrh told me that the target parameter:
```
  docker: {
    rti-image: "rti:local"
  }
``` 
Would trigger lfc to build a docker image, presumably with the Dockerfile, and this is apparently used in CI, I wonder why those tests didn't fail after your refactoring Peter

petervdonovan:
> I wonder why those tests didn't fail after your refactoring Peter

OK, filed a PR (https://github.com/lf-lang/lingua-franca/pull/2249) to fix the error handling.

erlingrj (approved):
This looks correct, ran into the same problem myself.

lhstrh (approved):
This looks good but needs to be merged after https://github.com/lf-lang/lingua-franca/pull/2249, which still reports failures.

PR #400, opened 2024-03-22
Zephyr make kernel clock default + improvements

erlingrj:
We have had some issues with using the Zephyr Counter API as the timing service, it is implemented differently for different boards and did not work for a STM32 board which was reported by a user on Zulip. I had other similar reports also earlier. Therefore, I want to make the Zephyr Kernel time API the default, it gives less precise timing, but it is simple and supported by all boards with Zephyr.

The Counter API can still be used, if the board supports it, but it is not the default.

Docs: https://github.com/lf-lang/lf-lang.github.io/pull/247
Lingua-franca: https://github.com/lf-lang/lingua-franca/pull/2248


lhstrh (approved):
LGTM!

lhstrh:
Oh, looks like macOS is failing :disappointed:

lhstrh:
But I see now that `main` is failing, too: https://github.com/lf-lang/reactor-c/actions/runs/8512792339/job/23315116610

erlingrj:
Yes, this should not affect macOS as we only build and test Zephyr with a Linux host

erlingrj:
Dont know what to make of it. No useful output from the CI and it passed when I re-ran it..

lhstrh:
FYI: I added the `mac` label to trigger the macOS tests that ended up failing in the merge queue.

erlingrj:
The macOS test failures can not come from any changes introduced here, so there is a problem already in `main` or with the CI. Unfortunately I dont get a log output

erlingrj:
Seems to be: `src/federated/DistributedPhysicalActionUpstreamLong.lf` which is timing out

lhstrh:
Could you mark it as failing and create an issue?

PR #403, opened 2024-03-28
Making MIN_SLEEP_DURATION a platform variable and Add busy wait

soyerefsane:
As discussed in #332, I have moved forward to make the following changes for the threaded runtime:

- If the wait duration is less than `MIN_SLEEP_DURATION` the `wait_until` function busy waits until `wait_until_time` is reached. 
- `MIN_SLEEP_DURATION` was defined as an arbitrary constant of 10 usec, I changed that to be a platform variable such that the value of `MIN_SLEEP_DURATION` is equal to underlying system's clock resolution.

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -28,6 +28,8 @@
 #include "reactor_common.h"
 #include "watchdog.h"
 
+#include "platform/lf_unix_clock_support.h"
+
 #ifdef FEDERATED
 #include "federate.h"
 #endif
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -241,6 +241,9 @@ bool wait_until(instant_t logical_time, lf_cond_t* condition) {
     if (wait_duration < MIN_SLEEP_DURATION) {
       LF_PRINT_DEBUG("Wait time " PRINTF_TIME " is less than MIN_SLEEP_DURATION " PRINTF_TIME ". Skipping wait.",
                      wait_duration, MIN_SLEEP_DURATION);
+      while (lf_time_physical() < wait_until_time) {
+        //Busy wait
+      }
       return true;
     }
 
--- a/include/core/reactor_common.h
+++ b/include/core/reactor_common.h
@@ -42,7 +42,7 @@
  * to prevent unnecessary delays caused by simply setting up and
  * performing the wait.
  */
-#define MIN_SLEEP_DURATION USEC(10)
+// #define MIN_SLEEP_DURATION USEC(10)
 
 //////////////////////  Global Variables  //////////////////////
 
--- a/low_level_platform/api/platform/lf_unix_clock_support.h
+++ b/low_level_platform/api/platform/lf_unix_clock_support.h
@@ -1,6 +1,8 @@
 #include <time.h>
 #include <errno.h>
 
+extern instant_t MIN_SLEEP_DURATION;
+
 /**
  * @brief Convert a _lf_time_spec_t ('tp') to an instant_t representation in
  * nanoseconds.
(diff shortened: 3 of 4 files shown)


soyerefsane:
I made the `MIN_SLEEP_DURATION` a platform property but realized the clock resolution value was just for Unix-based platforms, WIP is to update the other support files to add the declaration of `MIN_SLEEP_DURATION` based on their clk resolutions right now.

core/threaded/reactor_threaded.c
```diff
@@ -239,6 +241,9 @@ bool wait_until(instant_t logical_time, lf_cond_t* condition) {
     if (wait_duration < MIN_SLEEP_DURATION) {
       LF_PRINT_DEBUG("Wait time " PRINTF_TIME " is less than MIN_SLEEP_DURATION " PRINTF_TIME ". Skipping wait.",
                      wait_duration, MIN_SLEEP_DURATION);
```
edwardalee, line 243: This debug message is no longer accurate. Please update.

include/core/reactor_common.h
```diff
@@ -42,7 +42,7 @@
  * to prevent unnecessary delays caused by simply setting up and
  * performing the wait.
  */
-#define MIN_SLEEP_DURATION USEC(10)
+// #define MIN_SLEEP_DURATION USEC(10)
```
edwardalee, line 45: We prefer to just remove code rather than commenting it out.

low_level_platform/api/platform/lf_unix_clock_support.h
```diff
@@ -1,6 +1,8 @@
 #include <time.h>
 #include <errno.h>
 
+extern instant_t MIN_SLEEP_DURATION;
```
edwardalee, line 4: Suggest renaming `lf_min_sleep_duration`.  Our convention uses upper case for `#define` constants` and the prefix `lf_` for all variables whose scope extends beyond one file.

low_level_platform/impl/src/lf_unix_clock_support.c
```diff
@@ -6,6 +6,8 @@
 #include "logging.h"
 #include "platform/lf_unix_clock_support.h"
 
+instant_t MIN_SLEEP_DURATION;
```
edwardalee, line 9: Same comment.

low_level_platform/impl/src/lf_unix_clock_support.c
```diff
@@ -23,6 +25,7 @@ void _lf_initialize_clock() {
   }
 
   lf_print("---- System clock resolution: %ld nsec", res.tv_nsec);
+  MIN_SLEEP_DURATION = NSEC(res.tv_nsec);
```
edwardalee, line 28: This implementation does not match the description of the PR.

edwardalee (changes requested):
Good direction.  Suggestions added.

core/threaded/reactor_threaded.c
byeonggiljun: Resolved in [bb552dc](https://github.com/lf-lang/reactor-c/pull/403/commits/bb552dceaf6549a14f93206133d340462fd3d4cb).
https://github.com/lf-lang/reactor-c/blob/bb552dceaf6549a14f93206133d340462fd3d4cb/core/threaded/reactor_threaded.c#L205-L206

include/core/reactor_common.h
byeonggiljun: Resolved in [bb552dc](https://github.com/lf-lang/reactor-c/pull/403/commits/bb552dceaf6549a14f93206133d340462fd3d4cb).

low_level_platform/api/platform/lf_unix_clock_support.h
byeonggiljun: I renamed the variable in [bb552dc](https://github.com/lf-lang/reactor-c/pull/403/commits/bb552dceaf6549a14f93206133d340462fd3d4cb).

PR #411, opened 2024-04-14
Proper handling of negative value passed in as the number of federates

chanijjani:
Invoke error returns when negative values are given as the number of federates and delete unreachable code found by the [code coverage](https://github.com/lf-lang/rust-rti/pull/52#issue-2241887602) with unit test cases in Rust RTI.


lhstrh (approved):
Thanks for the fix!

chanijjani:
> Thanks for the fix!

@lhstrh Thank you for updating the PR title! I'll update similar fixes continually whenever I find them.

PR #414, opened 2024-04-25
Fix clock sync init option

edwardalee:
This PR fixes a bug where clock synchronization was not being done when the clock-sync target option was set to `init`, which is the default.

This also fixes a bug where if the clock-sync-options field collect-stats was set to true, stats would not be collected.

This also fixes an unrelated warning in net_util.c.


erlingrj (approved):
Thanks for fixing this, my bad for not including the `_LF_CLOCK_SYNC_INITIAL` compiler flag when I did that refactoring. 

However, I don't see how this should affect flakyness? If all federations run on the same host, then running clock sync between them  will just introduce clock sync error.

PR #415, opened 2024-04-26
Add fix to thread scheduling API

erlingrj:
Actually there were several mistakes in the original merge. It was not tested well enough and that's on me. I think we really need to add some tests for this API. A problem is that the program needs privileges to change its own scheduling policy and priority. I am not sure how easy it will be to do this in CI. But maybe not that hard.

Fixes:
- We need to initialize the priority to something valid when setting the scheduling policy. To avoid duplicating the logic of translating it from LF priority range to target platform priority range, I initially set it to max, and then call `lf_thread_set_priority`.
- There was a mistake in the function mapping the LF priority range into the target priority range. I am using integers for that, might be better with floats..
- Actually forward the pthread error codes. A problem is that we make several syscalls so you wont know exactly which failed.


lhstrh (approved):
LGTM!

PR #416, opened 2024-04-28
Unit testing of thread scheduling API + additional fixes

erlingrj:
- Add unit test of the scheduling API
- Fix a few more issues in the scheduling API
- Improve unit test CMake system
- Move RTI unit tests to the RTI folder and reuse the original CMake script
- Disable unit tests on Windows. They haven't been running anyways.


erlingrj:
OK, @lhstrh this is finally ready for review

lhstrh (commented):
I reviewed this earlier but forgot to submit it :facepalm:

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,16 @@ jobs:
...
           mkdir build
           cd build
           cmake .. ${{ inputs.cmake-args }}
           cmake --build .
-          make test
+      - name: Run unit tests with sudo. (Needed for changing prio/policy)
+        run: sudo make test
+        if: matrix.platform == 'ubuntu-latest'
+      - name: Run unit tests normal.
+        run: sudo make test
```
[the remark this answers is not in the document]
erlingrj: Sudo needed to use the scheduling API.

trace/impl/lib/lf-trace-impl.a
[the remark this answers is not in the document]
erlingrj: Accident that is now solved :)

test/scheduling/scheduling_api_test.c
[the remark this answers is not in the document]
erlingrj: Thanks, fixed

erlingrj:
@lhstrh the comments should all be addressed now

CMakeLists.txt
```diff
@@ -22,10 +22,6 @@ if(DEFINED LF_SINGLE_THREADED)
     add_compile_definitions(LF_SINGLE_THREADED=1)
 endif()
 
-# Warnings as errors
-add_compile_options(-Werror)
```
lhstrh, line 26: I see that you've added `-Werror` in the CMake config for the RTI, but what about the other modules?

core/federated/RTI/test/rti_common_test.c
```diff
@@ -256,5 +255,4 @@ int main(int argc, char** argv) {
   two_nodes_zero_delay();
   two_nodes_normal_delay();
   multiple_nodes();
-}
-#endif
\ No newline at end of file
+}
```
lhstrh, line 258:
```suggestion
}

```

lhstrh (approved):
A lot of CMake magic here that I have a hard time following, but this generally looks good!

CMakeLists.txt
erlingrj: It is also added in `core/CMakeLists.txt` for when compiling the integration tests and in `test/Tests.cmake` when compiling the tests.

erlingrj:
We appear to have a problem with `EnclaveFederatedRequestStop`, this test is a little strange since it uses enclaves which arent officially supported yet. And the initial implementation does not work with federations. It does, however, compile and run correctly on my machine. But we might want to remove that test

lhstrh (approved):

CMakeLists.txt
lhstrh: :+1:

lhstrh:
It looks like unit tests were disabled for Windows, but we still expect them to pass in order to be pushed onto the merge queue...

.github/workflows/unit-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   run:
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
```
lhstrh, line 14: Are you sure about this? We can make the step that does `sudo make test` conditional and only include `sudo` on macOS and Linux. Wouldn't that be better?

erlingrj: The thing is. The unit tests dont run for Windows on `main` either, it just fails silently instead. But when splitting it out in its own steps it fails with a make error. Since I do not have a windows machine I rather not get into debugging the Cmake+CTest on Windows...

I would say it sufficient to run the unit tests under Linux and mac

PR #417, opened 2024-05-02
Fix race condition in lf_watchdog_stop

erlingrj:
This PR addresses a race condition exposed when `lf_watchdog_start` and `lf_watchdog_stop` was called in very close succession.


edwardalee (approved):
LGTM. Good catch!

PR #420, opened 2024-05-07
TracePluginProperty fixes

erlingrj:
When trying to compile a simple example with an external trace plugin I got linker errors. It seems that we need to link with `dl`. I am not exactly sure why it has worked before without specifying this though.
- [x] Fix problem where the LF_TRACE_PLUGIN Cmake variable is not initialized when using docker
- [x] Fix displaying of the trace-plugin target property in lfc (https://github.com/lf-lang/lingua-franca/issues/2273)


lhstrh:
The fixmes in the description of this PR seem things that need to be addressed on the Lingua Franca end, are they not?

core/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ if (DEFINED LF_TRACE)
     message(STATUS "linking trace plugin library ${LF_TRACE_PLUGIN}")
     target_link_libraries(reactor-c PUBLIC lf::trace-api)
     target_link_libraries(reactor-c PRIVATE "${LF_TRACE_PLUGIN}")
+    target_link_libraries(reactor-c PRIVATE dl)
```
lhstrh, line 65: :magic_wand:

lhstrh (approved):

core/CMakeLists.txt
erlingrj: Added a brief commet

lhstrh: :+1:

lhstrh (approved):
Looks great! Thanks, @erlingrj.

erlingrj:
I am going to double check why this was working in other projects.



ons. 8. mai 2024 kl. 22:13 skrev Marten Lohstroh ***@***.***>:

petervdonovan (approved):
LGTM

Issue #409, opened 2024-04-08
Incoming sockets must be completely emptied if the destination is disconnected.


erlingrj:
If the RTI receives TAGGED_MSG from a federate and its destination is disconnected. The RTI must read out the complete message from the socket before returning. Currently, only the first chunk is read. What happens is that the RTI will search through the remaining chunks for the beginning of a new message. It is likely that it will, at some point, find one of the magic bytes and then decode a completely rubbish signal.

I am working on a fix and will submit soon. Posting it here just-in-case.

Jakio815:
@erlingrj Could you explain why the RTI should read out the message of a disconnected federate?
I don't see the reason why we need to read it, if it is not used. Is is for normal termination or abnormal termination?

edwardalee:
The message that needs to be read out is not from a disconnected federate, but _to_ a disconnected federate.  The sender is still connected.  The problem is that if the RTI does not read the entire message, then that still connected federate cannot, for example, send a message to another still connected federate.  The bytes for that new message will just pile up behind the unread original message.  And as @erlingrj says, the RTI will be reading the socket for further signals or messages from the federate, but will read payload information instead of real messages or signals.

Issue #408, opened 2024-04-08
read_from_socket should always retry if it returns with errno=EAGAIN


erlingrj:
Currently, it will only retry a certain number of times before deeming the socket connection for faulty. This will then close this socket, and in the case of the RTI, it will then deem a federate disconnected. 

I have observed this happening with federates that don't produce any signals/messages for a long time. E.g. the ones that have only intermittent events due to a physical action.

I will address this issue soon in a PR, logging it here just-in-case.

erlingrj:
Actually, bounding the number of retries was done in this PR https://github.com/lf-lang/reactor-c/pull/323, @edwardalee do you remember the reasoning?

What I have seen is that when there is that the call to `read` will return with EAGAIN after a while, so if there is periods with no data arriving from a federate it will just close the connection to it and drop any data destined for it.

edwardalee:
I agree that EAGAIN should always result in a retry.

PR #422, opened 2024-05-10
RTI and federate socket fixes

erlingrj:
Addressing the following two networking issues:
- #408 
- #409

### DIFF
--- a/core/federated/RTI/rti_remote.c
+++ b/core/federated/RTI/rti_remote.c
@@ -386,11 +386,10 @@ void handle_timed_message(federate_info_t* sending_federate, unsigned char* buff
   // issue a TAG before this message has been forwarded.
   LF_MUTEX_LOCK(&rti_mutex);
 
-  // If the destination federate is no longer connected, issue a warning
-  // and return.
+  // If the destination federate is no longer connected, issue a warning,
+  // remove the message from the socket and return. 
   federate_info_t* fed = GET_FED_INFO(federate_id);
   if (fed->enclave.state == NOT_CONNECTED) {
-    LF_MUTEX_UNLOCK(&rti_mutex);
     lf_print_warning("RTI: Destination federate %d is no longer connected. Dropping message.", federate_id);
     LF_PRINT_LOG("Fed status: next_event " PRINTF_TAG ", "
                  "completed " PRINTF_TAG ", "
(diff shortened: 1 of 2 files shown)


core/federated/RTI/rti_remote.c
```diff
@@ -401,6 +400,18 @@ void handle_timed_message(federate_info_t* sending_federate, unsigned char* buff
...
                  fed->enclave.last_provisionally_granted.time - start_time,
                  fed->enclave.last_provisionally_granted.microstep);
+    // If the message was larger than the buffer, we must empty out the remainder also.
+    size_t total_bytes_read = bytes_read;
+    while (total_bytes_read < total_bytes_to_read) {
+      bytes_to_read = total_bytes_to_read - total_bytes_read;
+      if (bytes_to_read > FED_COM_BUFFER_SIZE) {
+        bytes_to_read = FED_COM_BUFFER_SIZE;
+      }
+      read_from_socket_fail_on_error(&sending_federate->socket, bytes_to_read, buffer, NULL,
```
edwardalee, line 410: Should this really be fail on error?  It seems that if the fed is no longer connected, this is very likely to fail, and then the RTI will crash.

edwardalee (changes requested):
LGTM, except that I think the RTI should not exit if it fails to drain the socket connecting it to a federate that is no longer connected?

core/federated/RTI/rti_remote.c
erlingrj: Good point!

erlingrj: We are not receiving data _from_ the disconnected fed. But from another normal fed which wants to send data to the disconnected one. 

But, should the RTI exit with error on socket errors? Probably not, but we have `read_from_socket_fail_on_error` all over the place. I am not sure exactly what it should do if it is not able to read out this entire message, because if TCP is working then it should come eventually, or the other fed or even the network failed between two chunks. 

I think this will require quite careful thought and a pass through the entire RTI and federated code-base to make sure that we can handle different scenarios. So I think for this small change we should keep to current style of the RTI (which is exiting if reading/writing from sockets fail), and then do a more thorough well-thought change later?

edwardalee: Ah, yes, you are right.

PR #425, opened 2024-05-12
Fixed preprocessor directives for clock sync

edwardalee:
The preprocessor directives controlling clock synchronization were erroneously defaulting to having no clock synchronization at all.  Only by setting the clock-sync parameter of the target to on or init could you actually get clock synchronization to work. This PR fixes this so that the default clock synchronization is to perform clock synchronization at initialization and then stop synchronizing.

This will require bumping reactor-c in lingua-franca to have any effect.


erlingrj:
Are you sure that we want clock sync to be the default? For all federated programs running on the same host, this would just introduce a slight timing error into the program. For federated programs where there already is clock sync running from NTP/PTP, this would also possibly introduce errors. If we do the LF initial sync at a time where the hosts are a little "off", then later NTP/PTP would presumably bring the clocks together, however since we did an initial clock sync, we would keep this initial bias.

I think that it instead should be opt-in.

edwardalee:
Yes, initial clock sync should be the default. NTP is not very precise, and any clock sync error will cause behavior that  users won't expect, like waiting a long time before starting. When running on the same host with the RTI, clock sync is skipped.

erlingrj:
I think if the system is running NTP then doing an initial clock sync will only cause problems. Imagine that the two computers for some has a 5 sec offset initially, I think that is unlikely if they are running NTP though which should give accuracy to the millisecond level. If we initially add a constant bias of 5 sec to the system clock then this will turn into a 5sec clock sync error once NTP steers in the clocks correctly. I think this will be much more confusing to a user.

I think our application-level clock sync should only be used if the system is not running any other clock sync clients. So I think it should only be default if we believe that this type of system is the most probable target for distributed LF.

edwardalee:
I'm not sure about this argument. Our clock synchronization is with a clock that is much closer, typically, than what NTP is using. When it is turned on, we are seeing clock sync errors on the order of tens of microseconds, a level not achievable with NTP.  You are right that if NTP is running poorly and then suddenly starts running well why our program is running that this could create problems. But is this likely?

In any case, the default policy is a one-line change in this PR. The PR is still necessary because without it we can't actually set any default policy.

erlingrj (approved):
Looks good, not sure what the CI problem is, I cannot seem to re-run any of the tests either...

edwardalee:
Looks like CI is failing because a Python federate is sending a clock sync signal but the RTI doesn't know to expect it.  Perhaps this is an error in the script launching the federation?

erlingrj:
@edwardalee I believe this should solve our issues. My changes include:
1. `lf_time_add` with saturation and overflow/underflow detection
2. Provide clock_sync_add/subtract empty implementations in clock-sync.c for the scenario with FEDERATED but CLOCK_SYNC_OFF
3. Changes to the code-generator so that it defines LF_CLOCK_SYNC_OFF when we are on the same host as the RTI

tag/api/tag.h
```diff
@@ -104,6 +104,15 @@ tag_t lf_tag(void* env);
  */
 tag_t lf_tag_add(tag_t a, tag_t b);
 
+/**
+ * Adds an interval to an instant, checks for overflows and underflows.
```
edwardalee, line 108:
```suggestion
 * @brief Return the sum of an interval and an instant, saturating on overflow and underflow.
```

edwardalee (commented):
Nice!  Looks ready to me.

core/federated/RTI/rti_remote.c
```diff
@@ -1738,6 +1738,12 @@ void initialize_RTI(rti_remote_t* rti) {
...
 
+// The RTI includes clock.c, which requires the following functions that are defined
+// in clock-sync.c.  But clock-sync.c is not included in the standalone RTI.
+// Provide empty implementations of these functions.
+void clock_sync_add_offset(instant_t* t) { (void)t; }
+void clock_sync_subtract_offset(instant_t* t) { (void)t; }
+
 void free_scheduling_nodes(scheduling_node_t** scheduling_nodes, uint16_t number_of_scheduling_nodes) {
   for (uint16_t i = 0; i < number_of_scheduling_nodes; i++) {
     // FIXME: Gives error freeing memory not allocated!!!!
```
lhstrh, line 1749: Just noticed this FIXME, which looks a bit concerning. Not part of this PR, I know, but we should either confirm the problem and create an issue for it, or just remove it...

edwardalee: I think we are not seeing this error, so this can safely be removed.

core/tag.c
```diff
@@ -39,20 +39,43 @@ tag_t lf_tag(void* env) {
...
+  instant_t res = lf_time_add(a.time, b.time);
+  if (res == FOREVER) {
     return FOREVER_TAG;
-  if (b.time > 0)
+  }
+  if (res == NEVER) {
+    return NEVER_TAG;
+  }
+
+  if (b.time > 0) {
```
lhstrh, line 70: The comment "Ignore microstep of first arg if time of second is > 0" suggests that the code should be:
```suggestion
  if (b.microstep > 0) {
```
But why are we doing this?

core/tag.c
```diff
@@ -39,20 +39,43 @@ tag_t lf_tag(void* env) {
...
     a.microstep = 0; // Ignore microstep of first arg if time of second is > 0.
-  tag_t result = {.time = a.time + b.time, .microstep = a.microstep + b.microstep};
-  if (result.microstep < a.microstep)
-    return FOREVER_TAG;
-  if (result.time < a.time && b.time > 0)
+  }
+  tag_t result = {.time = res, .microstep = a.microstep + b.microstep};
+
+  // If microsteps overflows
+  if (result.microstep < a.microstep) {
```
lhstrh, line 76: Why are we returning `FOREVER_TAG` if the microstep overflows? Is it even a problem for the microstep to overflow? If it is a problem, that must mean there is an event on the queue that has a greater tag while it is supposed to denote an earlier event. But before this happens, wouldn't we likely have run out of memory already? I'm not quite sure what the right answer is, but the current choice strikes me as odd...

erlingrj: It does seem like something that should trigger a runtime error instead. Not sure. Could this happen any other way than a action scheduled with 0 delay 2^32 times?

lhstrh: I don't think it can be achieved in any other way. But if that's what happens, then wrapping around is a fine solution imo, and not an error. Moving time to infinity in response seems like a very weird thing to do.

core/tag.c
lhstrh: Oh, I see. You're right. Let me change my suggestion with a comment that clarifies this.

core/tag.c
erlingrj: If it wraps around it would be an error situation. If you just wrap around then the new tag is less than the previous. This could result in scheduling a logical action into the past, and the event will just be discarded. The other, more reasonable alternative in my opinion is to increment the timestamp. This way, at least the event is not dropped. However semantically it would be incorrect. I think in both cases an error must be printed

erlingrj: I will create an issue for this and reference it here in the code

core/federated/RTI/rti_remote.c
erlingrj: FIxed

core/tag.c
lhstrh:
```suggestion
  if (b.time > 0) {
    // NOTE: The reason for handling this case is to "reset" the microstep counter at each after delay.
```

PR #427, opened 2024-05-13
Dont apply/subtract clock sync offset from NEVER/FOREVER

erlingrj:
This caused an error when a worker tried to sleep FOREVER and and applied a positive offset and the wakeup instant wrapped and got negative.

@edwardalee: Note that this is a PR into your existing clock-sync PR. It might fix the CI issue


erlingrj:
@edwardalee we could consider adding an API for adding/subtracting instant_t which handles the saturation

edwardalee (approved):
Good catch! The `lf_tag_add` function in `tag.h`/`tag.c` performs more complete checks of overflow and underflow.  Maybe we should refactor this function into two, `lf_time_add` and `lf_tag_add` and then use `lf_time_add` for adding this offset.

PR #432, opened 2024-05-21
Additional debug print-outs for watchdogs

erlingrj:
Attempt to debug this issue on macOS: https://github.com/lf-lang/reactor-c/actions/runs/9170148115

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced debugging output for the watchdog functionality, improving traceability and issue diagnosis related to timing and termination.

- **Chores**
	- Changed branch name from `master` to `watchdog-mac`, indicating a shift in development focus.


edwardalee (approved):
LGTM!

erlingrj:
This appears to be an issue with really poor real-time performance on macOS. The original watchdog test should yield a 100msec overrun. However I am seeing that the watchdog thread wakes up around 150 msec after its scheduled timeout. And in most cases this is too late so that the LF runtime has had the chance to replenish its time. 

I now updated the watchdog test to yield a 600 msec overrun which has to be sufficient to get the watchdog thread reliably awoken in time. 

I am not entirely satisfied with this explanation though.

lhstrh:
@edwardalee, you approved this, but there was a test failure. Is there a chance that this would be fixed by the recent watchdog changes you worked on?

Issue #426, opened 2024-05-12
GEDF scheduler violates EDF policy


edwardalee:
The GEDF scheduler is supposed to execute all reactions with smaller (inferred) deadline regardless of level before executing any reactions with a larger deadline or with no deadline.  Unfortunately, it does not do that, as shown by the following example (due to @soyerefsane).  For some reason, instead of using a single reaction queue, the GEDF scheduler uses a separate reaction queue for each level (anybody know why?).  This means that it will only prioritize a reaction with a deadline over another reaction if both reactions are at the same level.

In the following program, at each tag, detetctor2 should always execute before task1. In the current realization, it never executes before task1.

<img width="948" alt="image" src="https://github.com/lf-lang/reactor-c/assets/8513334/0ba86f05-cf0e-4834-af03-ed103cddbb7c">

```
target C {
  timeout: 200ms,
  workers: 1
};

reactor Periodic(period:time = 50ms, name:string = "Unnamed") {
  timer trigger(0, period)
  
  output t: time
  
  reaction(trigger) -> t {=
    instant_t start_time = lf_time_physical_elapsed();
    lf_set(t, start_time);
    lf_print("%s started at physical time: " PRINTF_TIME, self->name, start_time);
  =} 
}

reactor Probe(dl:time = 2ms, name:string = "Unnamed") {
  input i: time
  output t: time
  
  reaction (i) -> t {=
    lf_set(t, lf_time_physical_elapsed());
    lf_print("%s started at physical time: " PRINTF_TIME, self->name, lf_time_physical_elapsed());
  =} deadline (dl) {=
    lf_set(t, lf_time_physical_elapsed());
    lf_print("%s VIOLATED DEADLINE at physical time: " PRINTF_TIME, self->name, lf_time_physical_elapsed());
  =}
}

main reactor {
  task1 = new Periodic(period = 50ms, name = "task1")
  detector1 = new Probe(dl = 50ms, name = "detector1")
  
  task1.t -> detector1.i  
  task2 = new Periodic(period = 50ms, name = "task2")
  detector2 = new Probe(dl = 25ms, name = "detector2")
  
  task2.t -> detector2.i
  
  reaction(task1.t, detector2.t) {=
    if (task1.t->is_present && detector2.t->is_present && task1.t->value < detector2.t->value) {
      lf_print_error_and_exit("EDF policy violated. detector2 should execute before task1 when both are triggered.");
    }
  =}
}
```

Sample output:

```
[...]

PR #433, opened 2024-05-21
Redesign of GEDF scheduler

edwardalee:
This PR has a [companion PR in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/2294).

This a PR addresses issue #426, which points out that the GEDF scheduler was not prioritizing reactions with tighter deadlines unless they also had the same or lower levels compared to reactions with looser deadlines. It also fixes a bug (exposed by the change to GEDF) where deadlines were not propagated upstream to reactions that lexically precede a reaction with a deadline.

This scheduler also achieves slightly better performance than the previous scheduler (390 ns/reaction vs. 450 ns/reaction on a performance test).

This PR also cleans up scheduler_instance.h, removing several items that are relevant only to specific schedulers and are not part of the scheduling API. There are also various other code cleanups.

This PR also removes the chain_ID field of reactions, which never worked.


lhstrh:
Seems to work well on Linux, but not on macOS :thinking:

edwardalee:
> Seems to work well on Linux, but not on macOS 🤔

It didn't run on Linux.  Looks like a problem with propagation of deadlines across federates!  We were masking this problem by having a flawed scheduler...

edwardalee:
> > Seems to work well on Linux, but not on macOS 🤔
> 
> It didn't run on Linux. Looks like a problem with propagation of deadlines across federates! We were masking this problem by having a flawed scheduler...

Actually, the situation looks worse. Deadlines are not propagated anywhere!  ReactorInstance.assignDeadlines is never called anywhere in the code generation tree.

lhstrh:
Oh you're right. I must have looked at the wrong CI output 🫣

core/threaded/scheduler_GEDF_NP.c
```diff
@@ -25,151 +33,34 @@
...
-    LF_PRINT_DEBUG("Scheduler: Worker %zu is trying to acquire the scheduling "
-                   "semaphore.",
-                   worker_number);
-    lf_semaphore_acquire(scheduler->semaphore);
-    LF_PRINT_DEBUG("Scheduler: Worker %zu acquired the scheduling semaphore.", worker_number);
-  }
+inline static void wait_for_other_workers_to_finish(lf_scheduler_t* scheduler, int worker_number) {
+  scheduler->number_of_idle_workers++;
+  tracepoint_worker_wait_starts(scheduler->env, worker_number);
+  LF_COND_WAIT(&scheduler->custom_data->reaction_q_changed);
```
petervdonovan, line 61: The doc comment is correct, but I would argue that the name of this procedure is misleading -- it could exit due to a spurious wakeup, or to a change to the event queue other than all workers finishing the current level. If it isn't possible to make this procedure truly return only when the other workers finish, then maybe its name should be updated and its doc comment should emphasize that it does not guard against spurious wakeups.

I do think that the implementation should work correctly as-is, so this comment should be easy to address.

core/threaded/scheduler_GEDF_NP.c
```diff
@@ -223,91 +105,138 @@ void lf_sched_init(environment_t* env, size_t number_of_workers, sched_params_t*
...
+        pqueue_pop(scheduler->custom_data->reaction_q);
+
+        // If there is another reaction at the current level and an idle thread, then
+        // notify an idle thread.
+        reaction_t* next_reaction = (reaction_t*)pqueue_peek(scheduler->custom_data->reaction_q);
+        if (next_reaction != NULL && LF_LEVEL(next_reaction->index) == scheduler->custom_data->current_level &&
+            scheduler->number_of_idle_workers > 0) {
+          // Notify an idle thread.
+          LF_COND_SIGNAL(&scheduler->custom_data->reaction_q_changed);
+        }
```
petervdonovan, line 140: I do not understand why an individual other worker is being signaled to here, after we are already processing a level. Why doesn't the worker that takes charge of advancing the level broadcast to all other workers then? Then they would all be woken up front, in a relatively cold path, instead of doing it piecemeal, in a relatively hot path.

core/threaded/scheduler_GEDF_NP.c
```diff
@@ -223,91 +105,138 @@ void lf_sched_init(environment_t* env, size_t number_of_workers, sched_params_t*
...
+        scheduler->custom_data->current_level = 0;
+#ifdef FEDERATED
+        // In case there are blocking network input reactions at this level, stall.
+        lf_stall_advance_level_federation_locked(scheduler->custom_data->current_level);
+#endif
+      } else {
+        // Some other workers are still working on reactions on the current level.
+        // Wait for them to finish.
+        wait_for_other_workers_to_finish(scheduler, worker_number);
+      }
```
petervdonovan, line 197: This could be factored out into a separate `static` procedure with a signature like `advance_tag_or_wait(lf_scheduler_t* scheduler, int worker_number)`

core/threaded/scheduler_GEDF_NP.c
```diff
@@ -223,91 +105,138 @@ void lf_sched_init(environment_t* env, size_t number_of_workers, sched_params_t*
...
+#ifdef FEDERATED
+          // In case there are blocking network input reactions at this level, stall.
+          lf_stall_advance_level_federation_locked(scheduler->custom_data->current_level);
+#endif
+        } else {
+          // Some workers are still working on reactions on the current level.
+          // Wait for them to finish.
+          wait_for_other_workers_to_finish(scheduler, worker_number);
+        }
+      }
```
petervdonovan, line 165: This could be factored out into a separate `static` procedure with a signature like `advance_level_or_wait(lf_scheduler_t* scheduler, int worker_number)`

petervdonovan (approved):
I left minor comments. Overall this looks like an improvement to the comprehensibility and scheduling behavior of the code.

core/threaded/scheduler_GEDF_NP.c
edwardalee: Again, I have opted for a somewhat smaller function `advance_level` to keep the decision logic in this main procedure.

PR #436, opened 2024-05-23
Let pthread detect if the cpu number is valid

erlingrj:
Previously we just checked if the cpu_number > NUM_CPUs. However this does not work reliable if we pin the process to a subset of the CPUs.


petervdonovan (approved):
LGTM. I agree that this is simpler and better and maintains compliance with the doc comment of the function.

PR #437, opened 2024-05-25
Run one worker on main thread

sberkun:
Replaces  #375

Runs a worker on the main thread, allowing number of workers to equal number of cores. This is important for the RP2040 and FlexPRET.

As a followup to this, the CMake check for flexpret should be adjusted so that it allows workers=threads rather than workers=threads-1.


edwardalee (approved):
LGTM! Let's merge this!

PR #440, opened 2024-05-31
Fix cmake syntax

edwardalee:
The CMake syntax previously given does not parse when NUMBER_OF_WORKERS is not defined (apparently, it tries to parse the whole thing with NUMBER_OF_WORKERS replaced by nothing), at least with cmake version 3.27.6.  This fixes that problem.

This also fixes lingua-franca-ref.txt to point to master.


cmnrd (approved):
Looks good. Is there a specific reason to do the check in cmake and not in the LF validator?

PR #441, opened 2024-05-31
RP2040 support based on low-level platform API

edwardalee:
This PR removes the use of macros defined in util.h because #including util.h does not work.
To build RP2040 code, we also need the [companion PR](https://github.com/lf-lang/lingua-franca/pull/2303).

### DIFF
--- a/low_level_platform/impl/src/lf_rp2040_support.c
+++ b/low_level_platform/impl/src/lf_rp2040_support.c
@@ -134,16 +134,14 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
   sem_reset(&_lf_sem_irq_event, 0);
   // create us boot wakeup time
   target = from_us_since_boot((uint64_t)(wakeup_time / 1000));
-  // Enable interrupts. NOTE: It would be nice to use the macro LF_CRITICAL_SECTION_EXIT,
-  // but there seems to be no way to #include "util.h" that works in this file.
+  // Enable interrupts.
   lf_critical_section_exit(env);
   // blocked sleep
   // return on timeout or on processor event
   if (sem_acquire_block_until(&_lf_sem_irq_event, target)) {
     ret_code = -1;
   }
-  // Disable interrupts. NOTE: It would be nice to use the macro LF_CRITICAL_SECTION_ENTER,
-  // but there seems to be no way to #include "util.h" that works in this file.
+  // Disable interrupts.
   lf_critical_section_enter(env);
   return ret_code;
 }
--- a/low_level_platform/impl/src/lf_rp2040_support.c
+++ b/low_level_platform/impl/src/lf_rp2040_support.c
@@ -134,15 +134,17 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
   sem_reset(&_lf_sem_irq_event, 0);
   // create us boot wakeup time
   target = from_us_since_boot((uint64_t)(wakeup_time / 1000));
-  // allow interrupts
-  LF_CRITICAL_SECTION_EXIT(env);
+  // Enable interrupts. NOTE: It would be nice to use the macro LF_CRITICAL_SECTION_EXIT,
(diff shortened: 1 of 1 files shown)


lhstrh:
While the change looks good to me, it doesn't seem to match the description of the PR. This seems to just deal with an API change in `reactor-c`.

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -134,15 +134,17 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
   sem_reset(&_lf_sem_irq_event, 0);
   // create us boot wakeup time
   target = from_us_since_boot((uint64_t)(wakeup_time / 1000));
-  // allow interrupts
-  LF_CRITICAL_SECTION_EXIT(env);
+  // Enable interrupts. NOTE: It would be nice to use the macro LF_CRITICAL_SECTION_EXIT,
```
lhstrh, line 137: Why would this be nice? Tagging @petervdonovan who introduced the low-level platform API. My sense is that the code change is correct but the comment isn't...

petervdonovan: I think that this comment is correct. There is a complexity in the build system architecture which prevents this code in the `low_level_platform` module from depending on `util.h`, which is in the `core` module. The `core` module depends on `low_level_platform`, so we prefer that `low_level_platform` does not also depend on `core`.

In my view, this complexity is essential complexity, not accidental complexity, and it is not really specific to CMake.

To make the code easy to understand, we would like for the dependency graph between the different modules to be a DAG; this way it is possible in principle to check the correctness of the most fundamental modules, then check the modules that depend on them, and so forth. If there are cyclic dependencies between modules, then everything in the cycle must be understood in order to understand one part. Therefore, in our preferred approach there is essential complexity in avoiding cyclic dependencies between `core` and `low_level_platform`.

If we want to use that macro in `low_level_platform`, then probably we would need to move the macro out of `core` and into `low_level_platform`. I do not have an opinion about whether such a refactoring is necessary, however.

lhstrh: The comment may be correct in the sense that it describes why the macro cannot be used, but I don't think it's useful to express a desire for using it 1. because there is a good reason this isn't possible and 2. because the code written here is functionally equivalent and isn't any less elegant. I suggest to just strike the comment.
```suggestion
  // Enable interrupts.
```

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -134,15 +134,17 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
...
+  // but there seems to be no way to #include "util.h" that works in this file.
```
lhstrh, line 138:
```suggestion
```

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -134,15 +134,17 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
...
+  lf_critical_section_exit(env);
   // blocked sleep
   // return on timeout or on processor event
   if (sem_acquire_block_until(&_lf_sem_irq_event, target)) {
     ret_code = -1;
   }
-  // remove interrupts
-  LF_CRITICAL_SECTION_ENTER(env);
+  // Disable interrupts. NOTE: It would be nice to use the macro LF_CRITICAL_SECTION_ENTER,
```
lhstrh, line 145:
```suggestion
  // Disable interrupts.
```

low_level_platform/impl/src/lf_rp2040_support.c
```diff
@@ -134,15 +134,17 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
...
+  // but there seems to be no way to #include "util.h" that works in this file.
```
lhstrh, line 146:
```suggestion
```

lhstrh (approved):

PR #442, opened 2024-06-02
Fix support for NRF52

edwardalee:
With the refactoring of platform support, code generation for NRF52 stopped working. This gets it working again, albeit with quite a few limitations.  To program an nRF52 board, we suggest starting with the [lf-nRF52 repository](https://github.com/lf-lang/lf-nRF52).  The limitations are documented there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected system name from "Nrf52" to "nRF52" in various configuration files for consistency.

- **Refactor**
  - Improved variable declarations and reuse in the scheduling function for better efficiency.
  - Updated interrupt handling and critical section logic to handle nesting correctly.

- **Documentation**
  - Updated platform name in documentation comments to "nRF52" for accuracy.
  
- **Enhancements**
  - Added macro `platform_define` to apply preprocessor definitions based on defined variables.


low_level_platform/api/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ target_include_directories(lf-low-level-platform-api INTERFACE ${CMAKE_CURRENT_L
 add_library(lf::low-level-platform-api ALIAS lf-low-level-platform-api)
 target_link_libraries(lf-low-level-platform-api INTERFACE lf::tag-api)
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Nrf52")
+if(${CMAKE_SYSTEM_NAME} STREQUAL "nRF52")
```
lhstrh, line 6: We've been using CamelCase for platforms, I think. If instead we want to capitalize platforms the way they are commonly referred to, we should also change `MacOS` to `macOS`. There might be more such cases...

lhstrh:
Any particular reason this is still marked as draft?

low_level_platform/api/CMakeLists.txt
edwardalee: I'm not sure our arbitrary policy should override what the devices are called.  Also, why isn't the correct camel case NRF52?  The acronym is spelled out when you say it, which is why OS is capitalized in macOS.

lhstrh: Good point. I'm fine with the change, honestly...

lhstrh:
@edwardalee what should we do with this PR? Include it in `0.8.0` or not?

low_level_platform/impl/src/lf_nrf52_support.c
```diff
@@ -258,23 +263,41 @@ int _lf_interruptable_sleep_until_locked(environment_t* env, instant_t wakeup_ti
...
- * @return int
+ * @return 0
  */
-int lf_enable_interrupts_nested() { return sd_nvic_critical_region_enter(&_lf_nested_region); }
+int lf_enable_interrupts_nested() {
+  if (_lf_nested_count == 0)
+    return 1; // Error. Interrupts have not been disabled.
+  _lf_nested_count--;
+  return sd_nvic_critical_region_exit(0);
+  // FIXME: If softdevice is not enabled, do the following instead of above:
```
lhstrh, line 283: Is there an issue for this? If not, we should create it, and reference it here.

lhstrh (approved):

low_level_platform/impl/src/lf_nrf52_support.c
edwardalee: I have created an issue: https://github.com/lf-lang/reactor-c/issues/460

PR #448, opened 2024-06-20
Suppress error: cast from pointer to integer of different size

petervdonovan:
Closes #447.

Summary: I reviewed the `pqueue_tag` code that Edward and Byeong-gil worked on, and I determined that it was safe because I think that in practice, there are essentially no important platforms on which it will not work properly to cast a void pointer to an `unsigned long long` and back. This is because an `unsigned long long` is always at least 64 bits, and pointers in the real world are never larger than 64 bits.

Therefore, I decided to suppress the error using an intermediate cast to `uintptr_t`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved priority calculation issues in the queue by adjusting type casting for consistent handling of elements.

PR #450, opened 2024-06-22
Use uintptr_t for unsigned int that can hold a pointer

edwardalee:
This PR addresses https://github.com/lf-lang/reactor-c/issues/447 by using an unsigned int data type that can go back and forth from a pointer.  To be useful, of course, we will need to point Lingua Franca to this updated reactor-c.

### DIFF
--- a/core/utils/pqueue.c
+++ b/core/utils/pqueue.c
@@ -35,5 +35,5 @@ void set_reaction_position(void* reaction, size_t pos) { ((reaction_t*)reaction)
 
 void print_reaction(void* reaction) {
   reaction_t* r = (reaction_t*)reaction;
-  LF_PRINT_DEBUG("%s: index: %llx, reaction: %p", r->name, r->index, reaction);
+  LF_PRINT_DEBUG("%s: index: %" PRIxPTR ", reaction: %p", r->name, r->index, reaction);
 }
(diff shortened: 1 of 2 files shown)


petervdonovan (approved):
LGTM; we can close #448 in favor of this one. I can confirm that using only 32 bits for the priority on 32-bit platforms will not break the reaction queue, which now seems to fit both the tag and deadline into only 32 bits.

By the way, if deadlines are in nanoseconds, I guess that means that deadlines exceeding 65 microseconds are treated as the same by our scheduler.

edwardalee:
I cannot see any way that the Zephyr failure here has anything to do with these changes, and I can't replicate problem due to this [failure](https://github.com/lf-lang/lf-west-template/issues/8).  Any ideas?

edwardalee:
Closing in favor of #448.

petervdonovan:
Above I wrote:

> I can confirm that using only 32 bits for the priority on 32-bit platforms will not break the reaction queue, which now seems to fit both the tag and deadline into only 32 bits.

After seeing the test failure, I took a second look. For the record, my original assessment was wrong -- based on the implementation of [lf_combine_deadline_and_level](https://github.com/lf-lang/reactor-c/blob/587a8f9498b69bc82f404e432f881d7bb0bcf383/core/reactor_common.c#L1203), it looks like the deadline is in fact represented with 48 bits. Based on the [failing test](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/DeadlineWithBanks.lf), it looks like this explains the test failure on a 32-bit Zephyr platform.

Issue #449, opened 2024-06-21
The type of federate_id is used mixing uint16_t and int, and int32_t.


Jakio815:
The use of federate_id's type is being mixed.

1. Definition of federate_id on the federate side. It is defined as **int**.
```
## util.c
/**
 * The ID of this federate. For a non-federated execution, this will be -1.
 * For a federated execution, it will be assigned in the generated code.
 */
int _lf_my_fed_id = -1;
```

2. The federate sends the federate_id on `MSG_TYPE_FED_IDS`. It is sending it as **unit16_t**.
```
    buffer[0] = MSG_TYPE_FED_IDS;
    // Next send the federate ID.
    if (_lf_my_fed_id > UINT16_MAX) {
      lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX);
    }
    encode_uint16((uint16_t)_lf_my_fed_id, &buffer[1]);
```

3. The RTI receives the federate_id. Receives as **uint16_t**.
```
    // Received federate ID.
    fed_id = extract_uint16(buffer + 1);
```
4. The function which received the federate_id, returns it as **int32_t**.
```
int32_t receive_and_check_fed_id_message(){
    ...
    return (int32_t)fed_id;
}
```
5. RTI mixes up using it. 
```
    // The first message from the federate should contain its ID and the federation ID.
    int32_t fed_id = receive_and_check_fed_id_message(&socket_id, (struct sockaddr_in*)&client_fd);
    if (fed_id >= 0 && socket_id >= 0 && receive_connection_information(&socket_id, (uint16_t)fed_id) &&
        receive_udp_message_and_set_up_clock_sync(&socket_id, (uint16_t)fed_id)) {
```

6. Later on clock synchronization. Federate sends `MSG_TYPE_CLOCK_SYNC_T3` with its federate_id as **int32_t**. However the buffer size is defined as `1 +sizeof(int)` which can be critical for 16-bit systems.
```
  // Reply will have the federate ID as a payload.
  unsigned char reply_buffer[1 + sizeof(int)];
  reply_buffer[0] = MSG_TYPE_CLOCK_SYNC_T3;
  encode_int32(_lf_my_fed_id, &(reply_buffer[1]));

  // Write the reply to the socket.
  LF_PRINT_DEBUG("Sending T3 message to RTI.");
  if (write_to_socket(socket, 1 + sizeof(int), reply_buffer)) {
[...]

PR #453, opened 2024-06-24
Usage of 'fixed-size' integer types in RTI code

Jakio815:
This resolves issue #449.

1. Clarified types of `federate_id`, which was defined as just `int` not `int32_t`, which could make problems on 16-bit systems.

2. Also, change the runtime clock-synchronization after initial clock-sync, to send a 2 byte uint16_t federate_id not int32_t which is 4 bytes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency and data representation in clock synchronization by switching message size calculations and federate ID encoding from `int` to `uint16_t`.
  - Removed obsolete checks and error handling for federate ID limits in connection functions.

- **Refactor**
  - Updated the type of `_lf_my_fed_id` and `lf_fed_id` from `int` to `uint16_t` for improved clarity and performance.


lhstrh:
@edwardalee?

core/federated/RTI/rti_remote.c
```diff
@@ -935,7 +935,7 @@ void* clock_synchronization_thread(void* noargs) {
         // If any errors occur, either discard the message or the clock sync round.
         if (!read_failed) {
           if (buffer[0] == MSG_TYPE_CLOCK_SYNC_T3) {
-            int32_t fed_id_2 = extract_int32(&(buffer[1]));
+            int32_t fed_id_2 = (int32_t)extract_uint16(&(buffer[1]));
```
cmnrd, line 938: Why are we casting a 16 bit unsigned int to a 32 bit signed int here?

core/federated/clock-sync.c
```diff
@@ -275,13 +275,13 @@ int handle_T1_clock_sync_message(unsigned char* buffer, int socket, instant_t t2
   // T3-T2 between receiving the T1 message and replying.
 
   // Reply will have the federate ID as a payload.
-  unsigned char reply_buffer[1 + sizeof(int)];
+  unsigned char reply_buffer[1 + sizeof(uint16_t)];
   reply_buffer[0] = MSG_TYPE_CLOCK_SYNC_T3;
-  encode_int32(_lf_my_fed_id, &(reply_buffer[1]));
+  encode_uint16((uint16_t)_lf_my_fed_id, &(reply_buffer[1]));
```
cmnrd, line 280: Casting a signed 32 bit int to an unsigned 16 bit it. Is this really safe?

include/core/utils/util.h
```diff
@@ -86,12 +86,12 @@ typedef struct lf_stat_ll {
...
  * @see xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CGenerator.xtend.
  */
-extern int _lf_my_fed_id;
+extern int32_t _lf_my_fed_id;
 
 /**
  * Return the federate ID or -1 if this program is not part of a federation.
  */
-int lf_fed_id(void);
+int32_t lf_fed_id(void);
```
cmnrd, line 94:
```suggestion
uint16_t lf_fed_id(void);
```

include/core/utils/util.h
```diff
@@ -86,12 +86,12 @@ typedef struct lf_stat_ll {
  * _lf_initialize_trigger_objects() is called.
  * @see xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CGenerator.xtend.
  */
-extern int _lf_my_fed_id;
+extern int32_t _lf_my_fed_id;
```
cmnrd, line 89:
```suggestion
extern uint16_t _lf_my_fed_id;
```

core/utils/util.c
```diff
@@ -69,7 +69,7 @@ print_message_function_t* print_message_function = NULL;
 /** The level of messages to redirect to print_message_function. */
 int print_message_level = -1;
 
-int lf_fed_id() { return _lf_my_fed_id; }
+int32_t lf_fed_id() { return _lf_my_fed_id; }
```
cmnrd, line 72:
```suggestion
uint16_t lf_fed_id() { return _lf_my_fed_id; }
```

core/utils/util.c
```diff
@@ -58,7 +58,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * The ID of this federate. For a non-federated execution, this will be -1.
  * For a federated execution, it will be assigned in the generated code.
  */
-int _lf_my_fed_id = -1;
+int32_t _lf_my_fed_id = -1;
```
cmnrd, line 61: Why not make this a 16 bit unsigned int?

```suggestion
uint16_t _lf_my_fed_id = -1;
```

core/utils/util.c
Jakio815: `lf_fed_id()` function in `util.c` and `util.h` is actually not used anywhere. `_lf_my_fed_id` is directly use everywhere, so this function can be deleted, however I just left it just in case.

core/utils/util.c
Jakio815: `uint16_t _lf_my_fed_id = -1;` will not work, since it is an uint16_t.

So, it seems to initialize the `_lf_my_fed_id` as -1, to ensure when it is used, it is a proper integer larger than zero. However, I don't exactly see if there are any fail cases that the `_lf_my_fed_id` will be left as -1.

core/federated/clock-sync.c
Jakio815: Two cases that can make a problem.
1. If you cast the signed 32-bit integer -1 to an unsigned 16-bit integer, the result will be 65535.
2. If you cast the signed 32-bit integer 70000 to an unsigned 16-bit integer, the result will be 4464 (70000 mod 65536).

`_lf_my_fed_id` is initialized as -1, and on the generated code, which is generated in `CExtension.java`, does assign a federate id as 
```
    // Set global variable identifying the federate.
    code.pr("_lf_my_fed_id = " + federate.id + ";");
```
So, I believe the federate id will always be a 0 or positive integer, however, I don't see any checks.

For case 2, there is a check on the federate side, such as `if (_lf_my_fed_id > UINT16_MAX) {`. So, it seems there will be no problem, but these features make it hard to find future errors. 

Are there any better ideas to initialize the federate_id?

core/utils/util.c
edwardalee: I think `lf_fed_id()` is meant to be user facing, part of the API a reactor writer can use. I think it is better to have a function here than unrestricted access to a global variable.

core/utils/util.c
edwardalee: I think it would be safe to remove the -1 initialization of `_lf_my_fed_id`. It seems this variable is set in generated code in the function that initializes the federate, so it should never go uninitialized.

edwardalee (changes requested):
This looks good to me, but I agree with @Jakio815 and @cmnrd that should go a step further and make it `uint16_t` throughout.  I don't think the initialization to -1 is needed.  I don't think it matters what it is initialized to because generated code will set it during initialization of the federate.

core/federated/RTI/rti_remote.c
Jakio815, line 938:
```suggestion
            uint16_t fed_id_2 = extract_uint16(&(buffer[1]));
```

core/federated/clock-sync.c
Jakio815, line 280:
```suggestion
  encode_uint16(_lf_my_fed_id, &(reply_buffer[1]));
```

core/utils/util.c
Jakio815, line 61:
```suggestion
uint16_t _lf_my_fed_id;
```

core/utils/util.c
```diff
@@ -112,23 +112,29 @@ void _lf_message_print(const char* prefix, const char* format, va_list args,
     // interleaved between threads.
     // vprintf() is a version that takes an arg list rather than multiple args.
     char* message;
+    if (_lf_my_fed_id < 0) {
```
Jakio815, line 115: This part gives compile time errors, since the uint16_t type will never be a negative integer.

However, if I remove the whole if and else state, `general_utils_pqueue_test_c` fails with a segfault.

It will not go through the removed if statement, and fail at line 129, because the `name` is a NULL pointer. Does anyone have ideas?

I'll leave the call stack below.
image

cmnrd: I think it would be good to replace `-1` with another id that indicates an uninitialized id or no id. Either `0x0000` or 0xffff.

core/utils/util.c
```diff
@@ -58,7 +58,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * The ID of this federate. For a non-federated execution, this will be -1.
  * For a federated execution, it will be assigned in the generated code.
  */
-int _lf_my_fed_id = -1;
+uint16_t _lf_my_fed_id = 0xffff;
```
cmnrd, line 61: Can we introduce a macro (e.g. `LF_INVALID_FED_IC`) instead of a magic number?

cmnrd (approved):
Looks good. My only nitpick is that I would like to suggest using a name macro instead of the magic number `0xffff`.

core/federated/federate.c
```diff
@@ -1792,10 +1792,6 @@ void lf_connect_to_federate(uint16_t remote_federate_id) {
       size_t buffer_length = 1 + sizeof(uint16_t) + 1;
       unsigned char buffer[buffer_length];
       buffer[0] = MSG_TYPE_P2P_SENDING_FED_ID;
-      if (_lf_my_fed_id > UINT16_MAX) {
-        // This error is very unlikely to occur.
-        lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX);
-      }
```
edwardalee, line 1798: I suggest:
```
      if (_lf_my_fed_id == UINT16_MAX) {
        lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX - 1);
      }
```

core/federated/federate.c
```diff
@@ -1915,9 +1911,6 @@ void lf_connect_to_rti(const char* hostname, int port) {
     unsigned char buffer[4];
     buffer[0] = MSG_TYPE_FED_IDS;
     // Next send the federate ID.
-    if (_lf_my_fed_id > UINT16_MAX) {
-      lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX);
-    }
```
edwardalee, line 1920: I suggest:
```
      if (_lf_my_fed_id == UINT16_MAX) {
        lf_print_error_and_exit("Too many federates! More than %d.", UINT16_MAX - 1);
      }
```

core/utils/util.c
edwardalee:
```suggestion
uint16_t _lf_my_fed_id = UINT16_MAX;  // Federate IDs are counted up from 0. If we hit this, we have too many.
```

edwardalee (approved):
LGTM!  I made some minor suggestions for paranoid coding.

core/utils/util.c
cmnrd: I consider UINT16_MAX also as a magic number. Unlike a specifically named variable, it does not capture the intended use. And you don't ever want to change the name of `UINT16_MAX`, but we might want to change the invalid (or maximum) federation ID.

edwardalee: Good point. I suggest a macro like LF_FED_ID_LIMIT to make it clear that it is an upper bound.

Jakio815:
I'm still getting failures on this test, `./gradlew core:compileKotlin` on Windows. Does anyone have ideas about this?

cmnrd:
No, the error is really odd. It seems to be related to some caching mechanism. @lhstrh do you have any idea why this fails? Is there a way to reset the CI cache?

lhstrh:
> No, the error is really odd. It seems to be related to some caching mechanism. @lhstrh do you have any idea why this fails? Is there a way to reset the CI cache?

Nope. Not a Windows user... Remove the Gradle cache and try again? It looks like this PR was almost ready to merge but somehow got stuck. How do we move it forward?

Jakio815:
https://github.com/lf-lang/reactor-c/actions/caches

Figured the problem. If the cache is full, that problem happens. I manually removed all caches, however looks like there is a way to do it using the CI script.

PR #456, opened 2024-06-27
Fix compilation errors in `initialize-master-thread-id`

petervdonovan:
This silences undefined reference linker errors on other platforms that either have no support or that do not currently implement the API.

This is a hack, but I think it is better than the status quo, where we are keeping uninitialized memory in an array that we might used later. As far as our code is concerned, it adds wrong implementations to our code base, but functionally, it replaces uninitialized memory with a null pointer.

This also implements the API for Windows, which should make the API legitimately work on all 3 of the major general-purpose OSes.


cmnrd (commented):
Great! Thanks for fixing this, Peter.

PR #457, opened 2024-06-28
Trace deadline violations

petervdonovan:
This allows deadline violations to be traced using a format that is similar to regular reaction executions. I am not sure if we want to mainline it, but we could do so backwards-compatibly.

### DIFF
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -164,6 +164,7 @@ int _lf_do_step(environment_t* env) {
         // Deadline violation has occurred.
         violation = true;
         // Invoke the local handler, if there is one.
+        tracepoint_reaction_starts(env, reaction, 0);
         reaction_function_t handler = reaction->deadline_violation_handler;
         if (handler != NULL) {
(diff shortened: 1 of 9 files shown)


cmnrd (commented):
Changes look good, but why does this remove the `lf_check_deadline` API function? This doesn't seem backwards compatible, or is it?

petervdonovan:
> Changes look good, but why does this remove the `lf_check_deadline` API function? This doesn't seem backwards compatible, or is it?

Ah, I forgot that this was a user-facing API and mistook it for dead code. I'll fix that.

cmnrd (approved):
Looks good! Should we get this merged?

petervdonovan:
> Looks good! Should we get this merged?

OK. I don't think it strictly needs to be merged, but I think that merging it _probably_ will not break things for anyone (unless they are depending on very specific properties of the tracing output) and will make observability a little simpler to maintain. Let's merge this if we can easily do so.

PR #461, opened 2024-07-01
Better error messages when HMAC authentication is attempted by federates when RTI does not support it

Jakio815:
This gives better error messages in lf-lang/lingua-franca#2326.
 When the federates are using the HMAC authentication as `auth:true`, and the RTI is not executed using the `-a` or `--auth` option, the RTI will not expect the `MSG_TYPE_FED_NONCE` type, and reject it. 
For better errors, add a rejection code `RTI_NOT_EXECUTED_WITH_AUTH` and messages.


edwardalee (approved):
LGTM!

PR #463, opened 2024-07-02
Fix adaptive scheduler

petervdonovan:
As a reminder, the adaptive scheduler is a scheduler that, at least when it was created, either matched or outperformed our other schedulers in our benchmarks. Maybe we have too many schedulers, and maybe we should remove it, but this is the reason why we have it in our code base. See the graph [here](https://github.com/lf-lang/lingua-franca/pull/1207); the green line labeled "C heuristic" is the adaptive scheduler.

This fixes two serious bugs in the adaptive scheduler. In one bug, casting and pointer arithmetic are done in the wrong order. In the other bug, the level is advanced even if, in the process of waiting for the level to be ready to advance, more reactions were injected onto the reaction queue.

It seems like the adaptive scheduler may not have kept up with updates to federated execution in the past year.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability by adjusting the control flow in the termination function to ensure proper cleanup.
  - Enhanced assertions and type handling in the scheduler to prevent runtime errors.

- **Refactor**
  - Refined pointer arithmetic and type casting in scheduler functions for better code maintainability.

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -1166,6 +1166,7 @@ void termination(void) {
       }
     }
   }
+  lf_tracing_global_shutdown();
   // Skip most cleanup on abnormal termination.
   if (_lf_normal_termination) {
     _lf_free_all_tokens(); // Must be done before freeing reactors.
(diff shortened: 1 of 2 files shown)


cmnrd (approved):
Looks good!

PR #464, opened 2024-07-04
RTI dockerfile support multiarchitecture builds. 

elgeeko1:
Add multiarchitecture support and use build stages to reduce image size.

1. Add platform-specific base images to allow multiarchitecture support
2. Use build stage to reduce image size by 20x.
3. Add maintainer documentation to Dockerfile.

`docker build . --file core/federated/RTI/rti.Dockerfile` works as it did before.

To build for multiple architectures:

`docker build . --file core/federated/RTI/rti.Dockerfile --platform=linux/amd64,linux/aarch64,linux/arm/7,linux/riscv64`


lhstrh (changes requested):
This looks great! The problem is that we're not currently testing Docker builds, only native builds (see https://github.com/lf-lang/reactor-c/blob/main/.github/workflows/build-rti.yml). We should add a simple test to make sure that building using Docker succeeds. Could you take care of this?

As a side note, we need to regularly push updates to DockerHub, which is currently done manually, and way too infrequently. We should automate this, but I'm unsure how to considering that neither the RTI nor reactor-c are versioned.

elgeeko1:
I'm not sure I'll have time to create a test for Docker, but this is something we could consider getting some help with. Since it's not part of the development workflow and only used by a small group, could we get by without adding a test?

In a sense RTI is versioned, when building [xronos-inc/lf-dev-docker](https://github.com/xronos-inc/lf-dev-docker), the Dockerfile builds the RTI submodule that is associated with the LF release.

lhstrh:
I could add a simple test that ensure the docker build works correctly.

lhstrh:
I'm actually having trouble building manually:

```
[marten@gen6 reactor-c]$ docker build . --file core/federated/RTI/rti.Dockerfile
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  65.45MB
Step 1/21 : ARG BASEIMAGE=alpine:latest
Step 2/21 : FROM --platform=linux/amd64 ${BASEIMAGE} AS base-amd64
latest: Pulling from library/alpine
Digest: sha256:b89d9c93e9ed3597455c90a0b88a8bbb5cb7188438f70953fede212a0c4394e0
Status: Downloaded newer image for alpine:latest
 ---> a606584aa9aa
Step 3/21 : FROM --platform=linux/arm64 ${BASEIMAGE} AS base-arm64
latest: Pulling from library/alpine
Digest: sha256:b89d9c93e9ed3597455c90a0b88a8bbb5cb7188438f70953fede212a0c4394e0
Status: Downloaded newer image for alpine:latest
 ---> 092561eea88f
Step 4/21 : FROM --platform=linux/arm/v7 ${BASEIMAGE} AS base-arm
latest: Pulling from library/alpine
Digest: sha256:b89d9c93e9ed3597455c90a0b88a8bbb5cb7188438f70953fede212a0c4394e0
Status: Downloaded newer image for alpine:latest
 ---> 7ba4509d8b7b
Step 5/21 : FROM --platform=linux/riscv64 riscv64/${BASEIMAGE} AS base-riscv64
 ---> 04f98b6119b4
Step 6/21 : FROM base-${TARGETARCH} as builder
invalid reference format
```
and

```
[marten@gen6 reactor-c]$ docker build . --file core/federated/RTI/rti.Dockerfile --platform=linux/amd64,linux/aarch64,linux/arm/7,linux/riscv64
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  65.45MB
Error response from daemon: "amd64,linux" is an invalid component of "linux/amd64,linux/aarch64,linux/arm/7,linux/riscv64": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
```

elgeeko1:
Testing for multiarchitecture builds would be more involved.

Running `docker build . ` should continue to work just fine. If you want to build for multiple platforms, you will need to configure docker accordingly. The same goes for any CI/CD pipeline that tests multiple platforms. See https://docs.docker.com/build/building/multi-platform/

lhstrh:
> Testing for multiarchitecture builds would be more involved.
> 
> Running `docker build . ` should continue to work just fine. If you want to build for multiple platforms, you will need to configure docker accordingly. The same goes for any CI/CD pipeline that tests multiple platforms. See https://docs.docker.com/build/building/multi-platform/

But `docker build . --file core/federated/RTI/rti.Dockerfile` error'ed for me (see above).

.github/workflows/build-rti.yml
```diff
@@ -4,16 +4,25 @@ on:
...
         run: .github/scripts/build-rti.sh -DAUTH=ON
+
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out reactor-c repository
+        uses: actions/checkout@v4
+      - name: Perform multi-arch build
+        run: |
+          docker build --file core/federated/RTI/rti.Dockerfile --platform=linux/amd64,linux/aarch64,linux/arm/7,linux/riscv64 .
```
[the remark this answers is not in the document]
elgeeko1: `docker buildx build` is deprecated

elgeeko1:
@lhstrh it looks like you're working with an older build of Docker? See the output

```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
```

Your docker installation shows you're using a deprecated builder that predates BuildKit (and the notation for build stages and multiarchitecure builds).

lhstrh:
Note to self: https://docs.docker.com/build/ci/github-actions/multi-platform/

lhstrh (approved):
This looks great! Also happy that this brought down the size of the image considerably. I added a smoke test to make sure the images are built successfully.

core/federated/RTI/rti.Dockerfile
```diff
@@ -1,13 +1,32 @@
...
+# base image from which all other stages derive
+ARG BASEIMAGE=alpine:latest
+
+# derive from the appropriate base image depending on architecture
+FROM --platform=linux/amd64 ${BASEIMAGE} AS base-amd64
+FROM --platform=linux/arm64 ${BASEIMAGE} AS base-arm64
+FROM --platform=linux/arm/v7 ${BASEIMAGE} AS base-arm
+FROM --platform=linux/riscv64 riscv64/${BASEIMAGE} AS base-riscv64
+
+FROM base-${TARGETARCH} as builder
```
Depetrol, line 10:
```suggestion
FROM alpine:latest
```

elgeeko1: Thanks @Depetrol, I agree the `FROM --platform` lines can be removed. They are a vestige of when there was not a riscv64 base image, but this is now available so no need for target-specific parents.

I suggest keeping line 2 `ARG BASEIMAGE=alpine:latest` as this is common in the Docker community to allow the user to override the base image. Then line 10 would be `FROM ${BASEIMAGE} as builder`

In either case, line 25 for the application stage needs to be updated.

core/federated/RTI/rti.Dockerfile
```diff
@@ -1,13 +1,32 @@
...
+RUN mkdir -p build
+WORKDIR /lingua-franca/core/federated/RTI/build
+RUN cmake -DAUTH=on ..
+RUN make
+RUN make install
+
+WORKDIR /lingua-franca
+
+# application stage
+FROM base-${TARGETARCH} as app
```
Depetrol, line 25:
```suggestion
FROM BASEIMAGE as app
```

core/federated/RTI/rti.Dockerfile
Depetrol: Thanks for the comment! I've updated the suggestion accordingly.

core/federated/RTI/rti.Dockerfile
```diff
@@ -1,13 +1,24 @@
...
+LABEL maintainer="lf-lang"
```
lhstrh, line 18: I'm not sure what "maintainer" is supposed to refer to, but if its the the org on Docker Hub then it should be `lflang`.
```suggestion
LABEL maintainer="lflang"
```

core/federated/RTI/rti.Dockerfile
```diff
@@ -1,13 +1,24 @@
...
-# Use ENTRYPOINT not CMD so that command-line arguments go through
-ENTRYPOINT ["RTI"]
+RUN apk add --no-cache \
+        gcc musl-dev cmake make \
+        libressl-dev
+RUN mkdir -p build
+WORKDIR /lingua-franca/core/federated/RTI/build
+RUN cmake -DAUTH=on ..
+RUN make
+RUN make install
```
Depetrol, line 12:
```suggestion
RUN set -ex && apk add --no-cache gcc musl-dev cmake make && \
    mkdir container && \
    cd container && \
    cmake ../ && \
    make && \
    make install
```

PR #468, opened 2024-07-15
Support zero-delay cycles

edwardalee:
Support zero-delay cycles by having the RTI notify downstream federates when a transient fed joins or leave and by setting ports from absent federates to known absent upon receiving a PTAG.


core/federated/RTI/rti_remote.c
```diff
@@ -2330,6 +2332,7 @@ void* lf_delayed_grants_thread(void* nothing) {
           free(next);
         }
       }
+      LF_MUTEX_UNLOCK(&rti_mutex);
```
edwardalee, line 2335: This unlock may occur with no matching lock.

core/federated/RTI/rti_remote.c
```diff
@@ -344,22 +344,23 @@ static int get_num_absent_upstream_transients(federate_info_t* fed) {
 }
 
 /**
- * @brief Send MSG_TYPE_UPSTREAM_CONNECTED to the specified federate.
+ * @brief Send MSG_TYPE_UPSTREAM_CONNECTED to the specified federate, only if it is connected.
```
edwardalee, line 347: There are two specified federates, so "it" is ambiguous here. Suggestion:
```
 * @brief Send MSG_TYPE_UPSTREAM_CONNECTED to the specified `destination` if it is connected to the RTI,
 * telling it that the specified `upstream` federate is also now connected.
```

core/federated/RTI/rti_remote.c
```diff
@@ -1063,32 +1064,52 @@ void handle_address_ad(uint16_t federate_id) {
 }
 
 /**
- * Send to the start time to the federate my_fed.
- * This function assumes the caller does not hold the mutex.
+ * Send the start time and tag to the federate my_fed.
```
edwardalee, line 1067: Suggestion:
```
 * @brief Send the global federation start time and the federate-specific starting tag to the specified federate.
```

core/federated/RTI/rti_remote.c
```diff
@@ -1063,32 +1064,52 @@ void handle_address_ad(uint16_t federate_id) {
...
  *
- * If it is the startup phase, the start_time will be the maximum received timestamps
- * plus an offset. The federate will then receive identical federation_start_time
- * and federate_start_tag.time (the federate_start_tag.microstep will be 0).
- * If, however, the startup phase is passed, the federate will receive different
- * values than stated above.
+ * During the startup phase, the start_time is calculated as the maximum received timestamp, plus an offset.
+ * The federate will then receive identical values for federation_start_time` and `federate_start_tag.time` (with
+ * `federate_start_tag.microstep` set to 0). After the startup phase, the federate will receive different values for
+ * these parameters.
```
edwardalee, line 1072: Suggestion:
```
 * For persistent federates and transient federates that happen to join during federation startup, the
 * `federation_start_time` will match the time in the `federate_start_tag`, and the microstep will be 0.
 * For a transient federate that joins later, the time in the `federate_start_tag` will be greater than the
 * federation_start_time`.
```

core/federated/RTI/rti_remote.c
```diff
@@ -1063,32 +1064,52 @@ void handle_address_ad(uint16_t federate_id) {
...
  *
- * This will also notify federates downstream of my_fed that this federate is now
- * connected.  This is important when there are zero-delay cycles.
+ * Before sending the start time and tag, this function performs the following actions:
+ *  - If my_fed is transient, notify federates downstream of its connection, ensuring proper handling of zero-delay
+ *    cycles.
```
edwardalee, line 1076: There is no need for this to be done before the messages to `my_fed`.  Note that there is no ordering semantics between messages sent to different federates.  I suggest moving this code to _after_ sending sending the start tag so that the notification does not occur if sending the start tag fails.

core/federated/RTI/rti_remote.c
```diff
@@ -1063,32 +1064,52 @@ void handle_address_ad(uint16_t federate_id) {
...
+ *  - Notify my_fed of all upstream transient federates that are connected.
```
edwardalee, line 1077: Notice that this notification does need to occur before sending the start tag.  These upstream connected messages are traveling on the same socket to `my_fed`, so, here, there _is_ ordering semantics.  You need for `my_fed` to receive these upstream connected messages before the start tag so that it doesn't start executing while thinking its upstream federates are not connected.

core/federated/RTI/rti_remote.c
```diff
@@ -1063,32 +1064,52 @@ void handle_address_ad(uint16_t federate_id) {
...
+      send_upstream_connected_locked(GET_FED_INFO(my_fed->enclave.downstream[i]), my_fed);
+    }
+  }
+  // A corner case occurs when an upstream transient joins at tag (0, 0), but `my_fed`
+  // either misses the notification or receives it late. The following ensures that
+  // `my_fed` is informed of all currently connected upstream transients.
+  // This also prevents `my_fed` from receiving the start time and starting execution
+  // before the upstream connection message is received.
+  // This also deals with an even less likely corner case where two transients are joining simultaneously and one is
+  // upstream of the other.
```
edwardalee, line 1098: I would simplify this comment. It's not really a corner case. Suggestion:
```
  // Notify my_fed of any upstream transient federates that are connected.
  // This has to occur before sending the start tag so that my_fed does not begin executing thinking that these
  // upstream federates are not connected.
```

edwardalee (commented):
This is better, but I've made suggestions.

core/federated/RTI/rti_remote.c
```diff
@@ -2307,17 +2307,19 @@ void* lf_connect_to_transient_federates_thread(void* nothing) {
...
+  // Hold the mutex only when accessing rti_remote->delayed_grants pqueue
   while (!rti_remote->all_federates_exited) {
     if (pqueue_delayed_grants_size(rti_remote->delayed_grants) > 0) {
       // Do not pop, but rather peek.
+      LF_MUTEX_LOCK(&rti_mutex);
       pqueue_delayed_grant_element_t* next = pqueue_delayed_grants_peek(rti_remote->delayed_grants);
       instant_t next_time = next->base.tag.time;
+      LF_MUTEX_UNLOCK(&rti_mutex);
       // Wait for expiration, or a signal to stop or terminate.
       if (lf_clock_cond_timedwait(&updated_delayed_grants, next_time)) {
```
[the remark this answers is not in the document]
ChadliaJerad: This is outdated and the problem does not exist anymore.

core/federated/RTI/rti_remote.c
ChadliaJerad: Added to the upcoming commit.

core/federated/RTI/rti_remote.c
ChadliaJerad: Yeah, makes sense. Added to the upcoming commit.

core/federated/RTI/rti_remote.c
ChadliaJerad: Added to the upcoming commit.

core/federated/RTI/rti_remote.c
ChadliaJerad: The upcoming commit notifies `my_fed` about its connected upstreams before sending the start time and tag. And its downstream are notified after sending the start time and tag.

core/federated/RTI/rti_remote.c
ChadliaJerad: Added to the upcoming commit.

core/federated/RTI/rti_remote.c
ChadliaJerad: Added to the upcoming commit.

core/federated/RTI/rti_remote.c
ChadliaJerad: This is also outdated and the problem is solved.

PR #469, opened 2024-07-15
Immediate start of federates with STA offset under decentralized coordination

Depetrol:
In decentralized coordination, the behavior used to be to wait for `STP_offset` before starting the worker threads. With this change, the federates start immediately without waiting for `STP_offset`. 

When the user sets to `STP_offset` to `FOREVER`, it enables dataflow-like behavior. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization during startup by handling start time offsets, reducing spurious violations in federated decentralized execution.

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -615,16 +615,16 @@ void _lf_initialize_start_tag(environment_t* env) {
 
 #if defined FEDERATED_DECENTRALIZED
   // If we have a non-zero STA offset, then we need to allow messages to arrive
-  // prior to the start time.  To avoid spurious STP violations, we temporarily
+  // at the start time.  To avoid spurious STP violations, we temporarily
   // set the current time back by the STA offset.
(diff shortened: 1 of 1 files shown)


edwardalee:
> Why was the waiting done so far and why is it safe to simply remove it?

An STA_offset of T (which should be called STA, safe to advance) means that at physical time t+T the federate can assume it has received all messages with tags _less than_ t.  If t is the start time, the federate already knows it has received all messages with tags less than t (because there are no messages with tags less than t).  Hence, at the start time, there is no need to wait.

This change enables very simple use of decentralized coordination for a subset of LF programs that do not deal with time, but rather just deal with data flow.

edwardalee (approved):
This looks right to me.

lhstrh:
I adjusted the title of the PR to match what I think this does, but I'm not sure I got it right...

lhstrh:
@edwardalee please also update the description (it doesn't seem like this PR enables data-flow like behavior as it only affects the behavior at startup?)

PR #477, opened 2024-08-09
Prevent hanging on waiting for timeout time

edwardalee:
This PR fixes a corner case where a federate with decentralized coordination that had no events at the timeout tag and has a large STA would hang even when its upstream federates exit and close the socket connection. The fix is to set the last known tag for input federates to `FOREVER_TAG` when a socket connection is closed.

It has a [companion PR in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/2399)

### DIFF
--- a/core/federated/federate.c
+++ b/core/federated/federate.c
@@ -282,12 +282,38 @@ static void update_last_known_status_on_input_port(environment_t* env, tag_t tag
     // federate that is far ahead of other upstream federates in logical time.
     lf_update_max_level(_fed.last_TAG, _fed.is_last_TAG_provisional);
     lf_cond_broadcast(&lf_port_status_changed);
+    lf_cond_broadcast(&env->event_q_changed);
   } else {
     // Message arrivals should be monotonic, so this should not occur.
     lf_print_warning("Attempt to update the last known status tag "
                      "of network input port %d to an earlier tag was ignored.",
                      port_id);
   }
+}
+
+/**
+ * @brief Mark all the input ports from this federate as known to be absent until FOREVER.
+ * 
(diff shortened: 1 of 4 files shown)


lhstrh:
Is this still meant to be a draft?

core/federated/federate.c
```diff
@@ -282,6 +282,7 @@ static void update_last_known_status_on_input_port(environment_t* env, tag_t tag
     // federate that is far ahead of other upstream federates in logical time.
     lf_update_max_level(_fed.last_TAG, _fed.is_last_TAG_provisional);
     lf_cond_broadcast(&lf_port_status_changed);
+    lf_cond_broadcast(&env->event_q_changed);
```
lhstrh, line 285: Why is this broadcast added here? It doesn't seem like this function should have any effect on the event queue...

core/federated/federate.c
```diff
@@ -290,6 +291,34 @@ static void update_last_known_status_on_input_port(environment_t* env, tag_t tag
   }
 }
 
+/**
+ * @brief Mark all the input ports from this federate as known to be absent until FOREVER.
```
lhstrh, line 295:
```suggestion
 * @brief Mark all the input ports connected to the given federate as known to be absent until FOREVER.
```

lhstrh (commented):
I left a question, but aside from that question, this looks good.

core/federated/federate.c
edwardalee: It is added because there are threads blocking on a change in an empty event queue that will never occur because the federation is shutting down.  In effect, once you know that there will be no more events, this _is_ a change in the status of the event queue.

lhstrh (approved):

PR #479, opened 2024-09-16
Fix compiler warnings in Zephyr and FlexPRET support files

erlingrj:
See title.

### DIFF
--- a/low_level_platform/impl/src/lf_atomic_gcc_clang.c
+++ b/low_level_platform/impl/src/lf_atomic_gcc_clang.c
@@ -11,9 +11,9 @@
 #include "platform/lf_atomic.h"
 #include "low_level_platform.h"
 
-int32_t lf_atomic_fetch_add(int* ptr, int value) { return __sync_fetch_and_add(ptr, value); }
+int lf_atomic_fetch_add(int* ptr, int value) { return __sync_fetch_and_add(ptr, value); }
 int64_t lf_atomic_fetch_add64(int64_t* ptr, int64_t value) { return __sync_fetch_and_add(ptr, value); }
-int32_t lf_atomic_add_fetch(int* ptr, int value) { return __sync_add_and_fetch(ptr, value); }
+int lf_atomic_add_fetch(int* ptr, int value) { return __sync_add_and_fetch(ptr, value); }
 int64_t lf_atomic_add_fetch64(int64_t* ptr, int64_t value) { return __sync_add_and_fetch(ptr, value); }
 bool lf_atomic_bool_compare_and_swap(int* ptr, int oldval, int newval) {
   return __sync_bool_compare_and_swap(ptr, oldval, newval);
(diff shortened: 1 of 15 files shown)


erlingrj:
This PR addresses some new compiler warnings that arose when we changed the FlexPRET compiler. It triggered me to change a little the atomics API. I removed the functions explicitly targeting int32_t. Because it is used on many variables which are of type `int` and then they need to be casted. What is really needed is to have an atomics API capable of modifying 64bit values.

There is a test failure on FlexPRET, this is because reactor-c CI is referring to the CI jobs in `master` of lingua-franca, which is outdated. It will pass when we merge: https://github.com/lf-lang/lingua-franca/pull/2405

low_level_platform/impl/src/lf_atomic_gcc_clang.c
```diff
@@ -11,17 +11,17 @@
 #include "platform/lf_atomic.h"
 #include "low_level_platform.h"
 
-int32_t lf_atomic_fetch_add32(int32_t* ptr, int32_t value) { return __sync_fetch_and_add(ptr, value); }
+int32_t lf_atomic_fetch_add(int* ptr, int value) { return __sync_fetch_and_add(ptr, value); }
```
edwardalee, line 14: Should this be returning `int`?

low_level_platform/impl/src/lf_atomic_gcc_clang.c
```diff
@@ -11,17 +11,17 @@
...
 int64_t lf_atomic_fetch_add64(int64_t* ptr, int64_t value) { return __sync_fetch_and_add(ptr, value); }
-int32_t lf_atomic_add_fetch32(int32_t* ptr, int32_t value) { return __sync_add_and_fetch(ptr, value); }
+int32_t lf_atomic_add_fetch(int* ptr, int value) { return __sync_add_and_fetch(ptr, value); }
```
edwardalee, line 16: Should this be returning `int`?

low_level_platform/impl/src/lf_atomic_gcc_clang.c
```diff
@@ -11,17 +11,17 @@
...
 int64_t lf_atomic_add_fetch64(int64_t* ptr, int64_t value) { return __sync_add_and_fetch(ptr, value); }
-bool lf_atomic_bool_compare_and_swap32(int32_t* ptr, int32_t oldval, int32_t newval) {
+bool lf_atomic_bool_compare_and_swap(int* ptr, int oldval, int newval) {
   return __sync_bool_compare_and_swap(ptr, oldval, newval);
 }
 bool lf_atomic_bool_compare_and_swap64(int64_t* ptr, int64_t oldval, int64_t newval) {
   return __sync_bool_compare_and_swap(ptr, oldval, newval);
 }
-int32_t lf_atomic_val_compare_and_swap32(int32_t* ptr, int32_t oldval, int32_t newval) {
+int32_t lf_atomic_val_compare_and_swap(int* ptr, int oldval, int newval) {
```
edwardalee, line 24: Should this be returning `int`?

edwardalee (approved):
Looks OK to me.  I added three questions.

erlingrj:
@edwardalee thanks for the review, sloppy of me to overlook the missing int32 -> int update. It is done now and I did a final self-review. Will merge when issues with the corresponding PR in `lingua-franca` is solved.

PR #480, opened 2024-09-17
Rename "version" directory to avoid Zephyr macOS issue

erlingrj:
The Zephyr build system looks for a file named `VERSION` in the root of a project. Since macOS, for some reason, has a default case-insensitive file system, the `version` directory in reactor-c gets interpreted as the `VERSION` file and it breaks the build.

To resolve this conflict I moved the `version` folder to `~/core/version`. I think this is acceptable.


erlingrj:
@petervdonovan, the reason I asked for your review Peter is that you know how this version library is being used in proprietary plugins. Will moving the location of this folder break any such plugins?

petervdonovan:
You are right that that was my initial intention. However, I don't think it is being used for that purpose yet, and I am not aware of current plans to use it for that purpose.

petervdonovan (approved):

PR #482, opened 2024-09-24
Fix action is_present field not being reset

erlingrj:
See https://github.com/lf-lang/lingua-franca/pull/2409

### DIFF
--- a/core/reactor_common.c
+++ b/core/reactor_common.c
@@ -1103,6 +1103,13 @@ void initialize_global(void) {
   // Call the code-generated function to initialize all actions, timers, and ports
   // This is done for all environments/enclaves at the same time.
   _lf_initialize_trigger_objects();
+
+#if !defined(NDBEUG)
+  // If we are testing, verify that environment with pointers is correctly set up.
+  for (int i = 0; i < num_envs; i++) {
+    environment_verify(&envs[i]);
+  }
+#endif
 }
 
 /**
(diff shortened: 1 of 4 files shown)


core/reactor_common.c
```diff
@@ -1096,6 +1103,13 @@ void initialize_global(void) {
   // Call the code-generated function to initialize all actions, timers, and ports
   // This is done for all environments/enclaves at the same time.
   _lf_initialize_trigger_objects();
+
+#if !defined(LF_SINGLE_THREADED) && !defined(NDBEUG)
```
edwardalee, line 1107:
```suggestion
#if !defined(LF_SINGLE_THREADED) && !defined(NDEBUG)
```

edwardalee (approved):
LGTM!  I'll look at the lingua-franca PR next.

erlingrj:
@edwardalee there is still some issue with that I am debugging. I can ping you when it is resolved!

edwardalee:
> @edwardalee there is still some issue with that I am debugging. I can ping you when it is resolved!

Looking more closely at this:

```
temp.pr(
    enclaveStruct
        + ".is_present_fields["
        + CUtil.bankIndex(instance)
        + " + "
        + enclaveInfo.numIsPresentFields
        + " + count] = &"
        + portRef
        + con
        + "is_present;");
```

I think maybe instead of `bankIndex(instance)` it should be `bankIndex(port.getParent())`.

PR #424, opened 2024-05-12
Set worker thread scheduling policy and priority

edwardalee:
This draft PR is not ready to be merged but is here to track progress.

This PR sets the thread scheduling policy and priorities of worker threads based on deadlines. It is an initial exploration on achieving preemptive scheduling when the number of worker threads is greater than the number of cores.

PR #484, opened 2024-09-27
A EDF scheduler based on enclaves

erlingrj:
This PR is based of this: https://github.com/lf-lang/reactor-c/pull/424 but I merged in an updated reactor-c and I want to keep the existing PR for reference.


erlingrj:
@edwardalee  and @fra-p, I have been looking at the proposed algorithm today and I have some general comments:

1. We should improve the readability by splitting into more functions, maybe use a generic linked list also.

2. I am worried that this approach might lead to starvation. This is because worker threads are doing work in the runtime/scheduler between invoking reactions while retaining the priority associated with the deadline of the reaction. I believe this could lead to starvation in special circumstances and that in general it can lead to a little unexpected timing.
I believe that we must give the workers the maximum priority when they are doing runtime-stuff. This is unfortunate as they will preempt more important work, but if our runtime has little overhead it might be OK?

3. It appears to be potentially alot of syscalls (`lf_thread_set_priority`) in the worst case when we do scheduling because we might need to shift the priorities of many threads. Since we are working with SCHED_FIFO we really only need three priorities:`PRIO_RUNTIME=98` used when sleeping and possibly doing runtime work, `PRIO_RUN=97` which means the thread should be executing, we should only ever have `N` threads with this priority, and finally `PRIO_WAIT=96` which will always be preempted by any thread with `PRIO_RUN`.

What does this mean? It means that when a worker fetches a reaction or completes a reaction (which is the time it adds itself and removes itself from the linked list), he must verify that we have N threads with `PRIO_RUN` by either bumping others up or down.

A downside is that it wont work well SCHED_FAIR, and I think it can be valuable to use this to set niceness values for soft real-time systems. 

Have you considered this approach? Am I missing something crucial here?

edwardalee:
All good points.
I think it's worth trying out the version with just a few priorities.
It would be much simpler...

PR #486, opened 2024-09-29
Various fixes to silence warnings on Windows

edwardalee:
This PR silences a large number of warnings that show up in the Windows CI, which must be using C23, which, among other things, deprecates a number of widely used standard C functions.

### DIFF
--- a/low_level_platform/impl/src/lf_windows_support.c
+++ b/low_level_platform/impl/src/lf_windows_support.c
@@ -170,9 +170,8 @@ int lf_thread_create(lf_thread_t* thread, void* (*lf_thread)(void*), void* argum
   // _beginthreadex requires a function that returns unsigned rather than void*.
   // So the following double cast suppresses the warning:
   // '_beginthreadex_proc_type' differs in levels of indirection from 'void *(__cdecl *)(void *)'
-  uintptr_t handle = _beginthreadex(NULL, 0,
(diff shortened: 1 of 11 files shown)


low_level_platform/impl/src/lf_windows_support.c
```diff
@@ -278,10 +299,20 @@ int _lf_cond_timedwait(lf_cond_t* cond, instant_t wakeup_time) {
   }
 
   // convert ns to ms and round up to closest full integer
-  DWORD wait_duration_ms = (wait_duration + 999999LL) / 1000000LL;
+  interval_t wait_duration_ms = (wait_duration + 999999LL) / 1000000LL;
+  DWORD wait_duration_saturated;
+  if (wait_duration_ms > 0xFFFFFFFFLL) {
```
erlingrj, line 304: Does this mean that we have a maximum sleep duration for windows = 2e32ms = 49 weeks (if I calculated correctly)?

erlingrj (approved):
This looks good to and the tests all pass, so LGTM

low_level_platform/impl/src/lf_windows_support.c
edwardalee: Yes.  Also, it means that on Windows, you can't get better 1ms time resolution in real time, but I guess that's OK. It's not an RTOS.

PR #487, opened 2024-10-03
Prevent a warning when the second timer event is after the timeout time.

edwardalee:
This prevents a warning when the second timer event is after the timeout time.
This also makes some semantically meaningless cleanups in lf_token.c/h.
There is a [companion PR in lingua-franca](https://github.com/lf-lang/lingua-franca/pull/2420).

### DIFF
--- a/core/lf_token.c
+++ b/core/lf_token.c
@@ -28,7 +28,7 @@ int _lf_count_token_allocations;
 
 /**
  * @brief List of tokens created within reactions that must be freed.
- * 
+ *
  * Tokens created by lf_writable_copy, which is automatically invoked
  * when an input is mutable, must have their reference count decremented
  * at the end of a tag (or the beginning of the next tag).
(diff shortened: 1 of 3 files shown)


core/lf_token.c
```diff
@@ -352,8 +367,7 @@ token_freed _lf_done_using(lf_token_t* token) {
 
 void _lf_free_token_copies() {
   while (_lf_tokens_allocated_in_reactions != NULL) {
-    lf_token_t* next = _lf_tokens_allocated_in_reactions->next;
```
erlingrj, line 355: Looking at `_lf_done_using` I cannot see why this reordering makes any difference?

erlingrj (approved):
This looks good. Just a questions. LGTM

core/lf_token.c
edwardalee: You are right, this is not the fix for the bug. This is just a cleanup. In fact, all the changes in this file are just cleanups.  The fix is in the lingua-franca PR.

Issue #393, opened 2024-03-14
Interrupts may cause deadlocks with threaded runtime


sberkun:
Currently, the threaded runtime uses mutexes to maintain exclusive access to key data structures (namely, environments). On embedded systems, lingua franca also supports scheduling physical actions from interrupts, via `lf_schedule`. However, the following sequence of events may occur:

 - Core 1 aquires the mutex (i.e. `LF_MUTEX_LOCK(&env->mutex)`)
 - Core 1 begins modifying the environment 
 - An interrupt occurs on core 1. The interrupt handler calls  `lf_schedule`, which also modifies the environment.
 - Interrupt handler finishes.
 - Core 1 finishes modifying environment
 - Core 1 releases mutex
 
In this case, the mutex was unhelpful, and the environment could be corrupted by the interrupt handler. The situation is even worse if the `lf_schedule` tries to acquire the mutex (is this the case? I could not find a `LF_MUTEX_LOCK` line). Then, deadlock occurs inside the interrupt handler.

There are a few potential solutions to this:
 - (1) Specify that mutexes should disable interrupts on embedded platforms. This is heavy-handed, but should work. The main concern is that interrupts would become disabled when any reaction is running, since reactions acquire a reactor-level mutex ([here](https://github.com/lf-lang/reactor-c/blob/7427d987d725111141139819c5341921edd0fef1/core/reactor_common.c#L806)). Interrupts would therefore only be enabled when all cores are sleeping.
 - (2) Create a separate platform API for regular mutexes, and mutexes that disable interrupts. (i.e. there would be `lf_mutex_init` and `lf_mutex_init_with_interrupts`). The environments would be locked with "interrupt mutexes", and reactors locked with regular mutexes. On most platforms, the two APIs would just be equivalent. However, this would be a 3rd type of locking-related API (in addition to regular mutexes and critical sections), which may clutter the platform API.
 - (3) Use critical sections instead of mutexes for environments. Very similar to above.
 - (4) Create a separate function, `lf_schedule_external`, for scheduling from interrupts. Instead of modifying the environment, this function could have it's own queue for external events. Then whenever an event occurs, the scheduler would need to transfer external events from the external event queue onto the regular event queue. The advantage of this approach is that interrupts only need to be disabled when the external event queue is being modified. The main issue with this approach would be "waking up" the scheduler whenever the external event queue has something new; Currently, waking up is handled by `_lf_cond_timedwait`, which assumes a mutex.

[...]

PR #488, opened 2024-10-07
disable interrupts for global mutex

sberkun:
Possible fix for https://github.com/lf-lang/reactor-c/issues/393 

Untested; to make sure this is viable, would need to demonstrate that (1) deadlocks actually occur without this fix, (2) this solves the deadlocks, and (3) there are no forseeable adverse affects from this additional logic. 

On most platforms, disabling/enabling interrupts is a no-op, so there should be no performance impact on these platforms.

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -1152,6 +1152,13 @@ int lf_notify_of_event(environment_t* env) {
  */
 int lf_critical_section_enter(environment_t* env) {
   if (env == GLOBAL_ENVIRONMENT) {
+    // disabling interrupts prevents possible deadlock
+    // when reaction is scheduled from an interrupt 
+    // on a core that holds the global mutex
+    int ret_code = lf_disable_interrupts_nested();
+    if (ret_code != 0) {
+      return ret_code;
+    }
     return lf_mutex_lock(&global_mutex);
   } else {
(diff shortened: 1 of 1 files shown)


core/threaded/reactor_threaded.c
```diff
@@ -1164,7 +1171,11 @@ int lf_critical_section_enter(environment_t* env) {
  */
 int lf_critical_section_exit(environment_t* env) {
   if (env == GLOBAL_ENVIRONMENT) {
-    return lf_mutex_unlock(&global_mutex);
+    int ret_code = lf_mutex_unlock(&global_mutex);
+    if (ret_code != 0) {
+      return ret_code;
```
edwardalee, line 1176:
```suggestion
      lf_enable_interrupts_nested();
      return ret_code;
```

edwardalee (changes requested):
This looks to me like an important change.  Is there a reason this a draft PR?  I’ve made one small suggestion, but otherwise, I think we should merge this.

erlingrj (approved):
This looks good to me also.

erlingrj:
We need to make disable/enable ISR part of the platform API for threaded applications also.

edwardalee (approved):
LGTM.  Let's merge this when the tests passs.

Issue #489, opened 2024-10-10
Duplicated Actions


Depetrol:
When two actions are triggered at about the same time, there is a chance that the later action value will override the previous action value. Example:

```python
target Python {
    coordination: decentralized,
    // logging: debug
  }
  
  preamble {=
    from fastapi import FastAPI, Request, HTTPException
    from fastapi.middleware.cors import CORSMiddleware
    import threading
    import uvicorn
    import asyncio
    import uuid
  =}
  
  reactor WebServer(bank_index=0, STA=0) {
    state app
    state logs
    physical action addlog_action
  
    reaction(startup) -> addlog_action {=
      self.logs = []
      self.app = FastAPI()
      self.app.add_middleware(
          CORSMiddleware,
          allow_origins=["*"],
          allow_credentials=True,
          allow_methods=["*"],
          allow_headers=["*"],
      )
      @self.app.post("/log")
      async def addlog(request: Request):
          log_message = (await request.json())["log"]
          print(f"Received log: {log_message}")
          addlog_action.schedule(0, log_message)
          return {"status": "success"}
  
      def run_fastapi_app():
          print(f"[WebServer{self.bank_index}] FastAPI server starting")
          uvicorn.run(self.app, host="127.0.0.1", port=5000+self.bank_index, log_level="warning")
      fastapi_thread = threading.Thread(target=run_fastapi_app)
      fastapi_thread.start()
    =}
  
    reaction(addlog_action) {=
        if addlog_action.value in self.logs:
            print(f"Duplicate Action: {addlog_action.value}")
            raise Exception("Duplicate Action")
        else:
            print(f"Action: {addlog_action.value}")
            self.logs.append(addlog_action.value)
    =}
  }
  
  main reactor {
    server = new WebServer()
  }
```
trigger the physical action with this python script:
```python
import requests
import time
import threading

def send_log(log_message, port=5000):
    url = f"http://127.0.0.1:{port}/log"
    log_data = {"log": f"{log_message}"}

    def send_request():
        try:
            requests.post(url, json=log_data, timeout=100)  # Send request without waiting for response
        except Exception as e:
            print(f"Error sending log: {e}")

    thread = threading.Thread(target=send_request)
    thread.start()

if __name__ == "__main__":
    for i in range(50):
        send_log(f"{i}", port=5000)
```

PR #490, opened 2024-10-11
Fix Unintended Action Override

Depetrol:
This PR fixes the unintended action override when two actions are triggered at about the same time in issue #489. The bug was in `convert_C_action_to_py`, the token value in the trigger is suspectable to concurrency overrides. The solution was to use the token value in the action. The companion PR in lingua-franca is https://github.com/lf-lang/lingua-franca/pull/2423

### DIFF
--- a/python/lib/pythontarget.c
+++ b/python/lib/pythontarget.c
@@ -542,8 +542,8 @@ PyObject* convert_C_action_to_py(void* action) {
   }
 
   // Actions in Python always use token type
-  if ((( generic_action_instance_struct*)action)->token!=NULL)
-  ((generic_action_capsule_struct*)cap)->value = (( generic_action_instance_struct*)action)->token->value;
+  if (((generic_action_instance_struct*)action)->token != NULL)
+    ((generic_action_capsule_struct*)cap)->value = ((generic_action_instance_struct*)action)->token->value;
 
   return cap;
 }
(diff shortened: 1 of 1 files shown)


erlingrj:
Could you explain how the race condition occurs, i.e. what is the sequence of events causing it. Also what exactly is this change doing?

erlingrj:
Well done on tracking it down BTW!

Depetrol:
Also huge thanks to @edwardalee and @erlingrj for helping me look into this!

Depetrol:
Not sure why the windows tests are failing -- this seems unrelated to the change in this PR since tests are also failing in the main branch.

edwardalee (approved):
Wow, great catch! Indeed, using the template token is not safe, so this looks like the right fix to me!

erlingrj (approved):
Great find Shulu! I think we can safely merge this and address Windows issues separately

Depetrol:
`reactor-c` requires that all tests pass before merging, so we might have to wait for https://github.com/lf-lang/lingua-franca/pull/2424 to me merged.

Depetrol:
There is another concurrency bug that is triggered probabilistically with the newly introduced test `ConcurrentAction.lf`. Running the test 20 times could trigger the bug quite consistently.

Depetrol:
Since the newly introduced test in https://github.com/lf-lang/lingua-franca/pull/2423 has been marked as failing in https://github.com/lf-lang/lingua-franca/pull/2422, the checks are passing and automatic merge was triggered. I'll start another PR for the other concurrency bug.

PR #492, opened 2024-10-30
Subtract time with care for overflow

edwardalee:
Subtract time with care for overflow.


erlingrj (approved):
The underflow and overflow handling looks correct!

PR #494, opened 2024-11-06
Stringify CMake command-line arguments before they are added as compile definitions

erlingrj:
Needed for: https://github.com/lf-lang/lingua-franca/pull/2432

The paths to the source directory, binary directory and src-gen directory are passed to CMake as command line arguments. They are originally passed as strings using quotation marks. In the LF change linked I remove the quotation marks because it means you cannot easily use them in CMake. I think these paths are most useful for the user when he adds a cmake-include. 

So in this PR I add quotation marks to the paths before they are made into compile definitions so that they conveniently can be used in the user C code as well.


lingua-franca-ref.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+cmake-init-include
```
edwardalee, line 1: Reset to master when merging.

edwardalee (approved):
LGTM.

core/CMakeLists.txt
```diff
@@ -153,6 +153,13 @@ macro(define X)
     endif(DEFINED ${X})
 endmacro()
 
+macro(defineString X)
```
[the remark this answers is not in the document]
erlingrj: Thanks

lingua-franca-ref.txt
erlingrj: Yes!

PR #495, opened 2024-11-07
[STATIC SCHEDULER] Use lf_sleep instead of busywait

erlingrj:
Use lf_sleep instead of busywaiting for max precision on FlexPRET


erlingrj:
Hold on, still some issues. Will ping you when its read

erlingrj:
@lsk567 can you review this? It is blocking the WCET analysis of hte pretVM instructions

lsk567 (approved):
Looks good to me!

I wonder what would happen if the argument to `lf_sleep` is negative. But I assume this is taken care of in `lf_sleep`.

PR #496, opened 2024-11-11
Let user specify the start-time of the program

erlingrj:
This PR adds a new command line argument `-s --start-time` where the user can specify the start time of the program. This currently only works for non-federated programs. In a federated setting the program will then wait until the specified start time (if it is in the future) before reaching out to the RTI and negotiating a start time.

### DIFF
--- a/core/reactor.c
+++ b/core/reactor.c
@@ -329,7 +329,15 @@ int lf_reactor_c_main(int argc, const char* argv[]) {
     LF_PRINT_DEBUG("Initializing.");
     initialize_global();
     // Set start time
-    start_time = lf_time_physical();
+    if (!start_time_specified) {
+      start_time = lf_time_physical();
+    } else {
+      instant_t now = lf_time_physical();
+      if(now < star_time) {
+        LF_PRINT_LOG("Sleeping " PRINTF_TIME " ns until start time", start_time - now);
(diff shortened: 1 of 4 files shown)


edwardalee (changes requested):
This looks very useful, but there are two issues:
1. The start time is not a duration, so the usage message and variable names should be updated.  It's an absolute start time.
2. It would be useful to have some hints about how to specify a start time. It requires units, but on a POSIX machine, it will be specifying the number of nanoseconds since Jan. 1, 1970.    Maybe there is a utility function that will tell me what, say, 2PM today is in seconds?

erlingrj:
Thanks for your feedback Edward! I have done a couple of changes:

1. Better help message:
>   -s, --start-time <duration> <units> 
   The logical start time of the program, expressed as a duration since the epoch of the underlying system clock.
   The units are nsec, usec, msec, sec, minute, hour, day, week or the plurals of those.
   On linux, to compute a start time in 2 minutes you can do:.
     `date -d "2 minutes" +%s`.
I still went with duration, because it is the duration from the epoch of the system clock. (Which I guess is the way that one would define an absolute time point also. 

2. Dont wait for start_time to arrive if fast mode is enabled
3. Refactor the setting and waiting for start time into a function that can be used by both unthreaded and threaded execution

PR #497, opened 2024-12-05
Fix lingua-franca-ref to point to master instead of fix-concurrency

ChadliaJerad:
This PR simply points `lingua-franca-ref` to `master` instead of `fix-concurrency`.

### DIFF
--- a/lingua-franca-ref.txt
+++ b/lingua-franca-ref.txt
@@ -1,1 +1,1 @@
-fix-concurrency
+master


ChadliaJerad:
No more needed, as solved in [this commit](https://github.com/lf-lang/reactor-c/commit/cdec23d3549a265f6c4ede0f0e9f29a0d9b3867c).

PR #498, opened 2024-12-05
Update trace_impl.c

MoezBHH:
Fix the compile error on issue [2398 in lingua-franca](https://github.com/lf-lang/lingua-franca/issues/2398)


edwardalee (approved):
Nice simple fix! Let's merge.

PR #506, opened 2024-12-20
Add `shutdown_socket()` function for follow up of #505

Jakio815:
This PR follow ups #505.
Please merge with lf-lang/lingua-franca#2474.

## New Integrated Function: `shutdown_socket()`
`int shutdown_socket(int* socket, bool read_before_closing);`

**Arguments**
- `socket`: A pointer to the socket descriptor that needs to be shut down and closed.
- `read_before_closing`: A boolean indicating whether the socket should read any remaining incoming data before closing:
  - true: Initiates a graceful shutdown by sending a `FIN` packet (`SHUT_WR`) and waits for `EOF` (0-length message) from the peer.
  - false: Immediately shuts down both reading and writing directions (`SHUT_RDWR`).

**Return Values**
- `0`: Indicates successful shutdown and closure of the socket.
- `-1`: Indicates a failure during shutdown or closure, with errno set to describe the specific error.

**Function Description**
This function gracefully shuts down and closes a socket.
- When read_before_closing is false:
  - The function calls `shutdown(SHUT_RDWR)` to immediately stop both sending and receiving data. Then calls `close()`.
- When read_before_closing is true:
  - The function calls `shutdown(SHUT_WR)` to signal the end of writing but allows reading to continue.
  - It waits for the peer to send an `EOF` (indicated by `read()` returning 0), and discards all received bytes.

This function holds a new mutex named `shutdown_mutex`, which ensures preventing deadlocks.

### `close_inbound_socket()`
The original code looked not updated. There were no use case when the argument `flag` was not 1. I removed the `flag` argument, and all unused code.
### `close_outbound_socket()`
The original flags were only set by this one line code.
`int flag = _lf_normal_termination ? 1 : -1;`
So it depends on `_lf_normal_termination`. However, this function can directly use `_lf_normal_termination`, so I removed the `flag`, and directly used `_lf_normal_termination`.

## Minor implementation details for macOS.
When the federate sends the `MSG_TYPE_RESIGN` signal, the RTI calls `shutdown(SHUT_WR)`. Then, in the federate side, the thread from `listen_to_rti_TCP()`'s `read_from_socket()`'s `read()` returns with a EOF. Then, it calls `shutdown_socket()`.

However, the `shutdown()` fails, with an `errno` `Socket is not connected` on only MacOS. even though the connection is half open. However, since the connection is half p[em, the federate should still send a `FIN_ACK` packet to the RTI, because the RTI is blocked on the `read()`.

Thus, the federate does not return on failure of `shutdown()`, and calls the `close()`, to ensure the `FIN_ACK` packet is sent.

To summarize, even when `shutdown()` fails, the code will try `close()`, to ensure the socket is closed.

## Minor addition of logic in `federate.c`'s `handle_tagged_message()`
In `federate.c`'s `handle_tagged_message()`, when the received message already missed the tag, it closes the inbound socket, however does not return -1 for failure of reading the message. This is logic is added.

## Some mutex locks removed.
`socket_mutex` was only used as federate's inbound socket mutex. Especially, it was only used when closing inbound sockets. The new `shutdown_socket()` function holds its own mutex, so removed the duplication of mutex locks.

### DIFF
--- a/core/federated/RTI/rti_remote.c
+++ b/core/federated/RTI/rti_remote.c
@@ -871,14 +871,7 @@ static void handle_federate_failed(federate_info_t* my_fed) {
   // Indicate that there will no further events from this federate.
   my_fed->enclave.next_event = FOREVER_TAG;
 
-  // According to this: https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket,
-  // the close should happen when receiving a 0 length message from the other end.
-  // Here, we just signal the other side that no further writes to the socket are
-  // forthcoming, which should result in the other end getting a zero-length reception.
-  shutdown(my_fed->socket, SHUT_RDWR);
-
-  // We can now safely close the socket.
-  close(my_fed->socket); //  from unistd.h
+  shutdown_socket(&my_fed->socket, false);
 
   // Check downstream federates to see whether they should now be granted a TAG.
   // To handle cycles, need to create a boolean array to keep
@@ -917,21 +910,7 @@ static void handle_federate_resign(federate_info_t* my_fed) {
   // Indicate that there will no further events from this federate.
   my_fed->enclave.next_event = FOREVER_TAG;
 
-  // According to this: https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket,
-  // the close should happen when receiving a 0 length message from the other end.
-  // Here, we just signal the other side that no further writes to the socket are
-  // forthcoming, which should result in the other end getting a zero-length reception.
-  shutdown(my_fed->socket, SHUT_WR);
-
-  // Wait for the federate to send an EOF or a socket error to occur.
-  // Discard any incoming bytes. Normally, this read should return 0 because
-  // the federate is resigning and should itself invoke shutdown.
-  unsigned char buffer[10];
-  while (read(my_fed->socket, buffer, 10) > 0)
-    ;
-
-  // We can now safely close the socket.
-  close(my_fed->socket); //  from unistd.h
+  shutdown_socket(&my_fed->socket, true);
 
   // Check downstream federates to see whether they should now be granted a TAG.
   // To handle cycles, need to create a boolean array to keep
@@ -1030,9 +1009,7 @@ void send_reject(int* socket_id, unsigned char error_code) {
     lf_print_warning("RTI failed to write MSG_TYPE_REJECT message on the socket.");
   }
   // Close the socket.
-  shutdown(*socket_id, SHUT_RDWR);
-  close(*socket_id);
-  *socket_id = -1;
+  shutdown_socket(socket_id, false);
   LF_MUTEX_UNLOCK(&rti_mutex);
 }
 
@@ -1420,9 +1397,7 @@ void lf_connect_to_federates(int socket_descriptor) {
       if (!authenticate_federate(&socket_id)) {
         lf_print_warning("RTI failed to authenticate the incoming federate.");
         // Close the socket.
-        shutdown(socket_id, SHUT_RDWR);
-        close(socket_id);
-        socket_id = -1;
+        shutdown_socket(&socket_id, false);
         // Ignore the federate that failed authentication.
         i--;
         continue;
@@ -1490,8 +1465,7 @@ void* respond_to_erroneous_connections(void* nothing) {
       lf_print_warning("RTI failed to write FEDERATION_ID_DOES_NOT_MATCH to erroneous incoming connection.");
     }
     // Close the socket.
-    shutdown(socket_id, SHUT_RDWR);
-    close(socket_id);
+    shutdown_socket(&socket_id, false);
   }
   return NULL;
 }
@@ -1554,21 +1528,10 @@ void wait_for_federates(int socket_descriptor) {
   // Shutdown and close the socket that is listening for incoming connections
   // so that the accept() call in respond_to_erroneous_connections returns.
   // That thread should then check rti->all_federates_exited and it should exit.
-  if (shutdown(socket_descriptor, SHUT_RDWR)) {
-    LF_PRINT_LOG("On shut down TCP socket, received reply: %s", strerror(errno));
-  }
-  // NOTE: In all common TCP/IP stacks, there is a time period,
-  // typically between 30 and 120 seconds, called the TIME_WAIT period,
-  // before the port is released after this close. This is because
-  // the OS is preventing another program from accidentally receiving
-  // duplicated packets intended for this program.
-  close(socket_descriptor);
+  shutdown_socket(&socket_descriptor, false);
 
   if (rti_remote->socket_descriptor_UDP > 0) {
-    if (shutdown(rti_remote->socket_descriptor_UDP, SHUT_RDWR)) {
-      LF_PRINT_LOG("On shut down UDP socket, received reply: %s", strerror(errno));
-    }
-    close(rti_remote->socket_descriptor_UDP);
+    shutdown_socket(&rti_remote->socket_descriptor_UDP, false);
   }
 }
 
--- a/core/federated/RTI/rti_remote.c
+++ b/core/federated/RTI/rti_remote.c
@@ -1563,7 +1563,7 @@ void initialize_RTI(rti_remote_t* rti) {
 
   // Initialize thread synchronization primitives
   LF_MUTEX_INIT(&rti_mutex);
-  LF_MUTEX_INIT(&shutdown_mutex);
+  init_shutdown_mutex();
   LF_COND_INIT(&received_start_times, &rti_mutex);
   LF_COND_INIT(&sent_start_time, &rti_mutex);
 
--- a/core/federated/RTI/rti_remote.c
+++ b/core/federated/RTI/rti_remote.c
@@ -1510,7 +1510,7 @@ void initialize_federate(federate_info_t* fed, uint16_t id) {
 int32_t start_rti_server(uint16_t port) {
   _lf_initialize_clock();
   // Create the TCP socket server
-  if (create_TCP_server(port, &rti_remote->socket_descriptor_TCP, &rti_remote->final_port_TCP)) {
+  if (create_TCP_server(port, &rti_remote->socket_descriptor_TCP, &rti_remote->final_port_TCP, true)) {
     lf_print_error_system_failure("RTI failed to create TCP server: %s.", strerror(errno));
   };
   lf_print("RTI: Listening for federates.");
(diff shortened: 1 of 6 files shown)


core/federated/network/socket_common.c
```diff
@@ -155,7 +155,7 @@ static int create_server(uint16_t port, int* final_socket, uint16_t* final_port,
     return -1;
   }
   set_socket_timeout_option(socket_descriptor, &timeout_time);
-  int used_port = set_socket_bind_option(socket_descriptor, port, increment_port_on_retry);
+  uint16_t used_port = set_socket_bind_option(socket_descriptor, port, increment_port_on_retry);
```
Jakio815, line 158: Minor fix on port's data type.

core/federated/federate.c
```diff
@@ -1943,9 +1918,11 @@ void lf_connect_to_rti(const char* hostname, int port) {
 
 void lf_create_server(int specified_port) {
   assert(specified_port <= UINT16_MAX && specified_port >= 0);
-  if (create_TCP_server(specified_port, &_fed.server_socket, (uint16_t*)&_fed.server_port, false)) {
+  uint16_t port;
+  if (create_TCP_server(specified_port, &_fed.server_socket, &port, false)) {
```
Jakio815, line 1922: Minor fix on port's data type.

core/federated/RTI/rti_remote.c
```diff
@@ -1030,9 +1009,7 @@ void send_reject(int* socket_id, unsigned char error_code) {
     lf_print_warning("RTI failed to write MSG_TYPE_REJECT message on the socket.");
   }
   // Close the socket.
```
hokeun, line 1011: Minor: how about adding more information for this call? For example, "Close the socket without reading until EOF."

core/federated/RTI/rti_remote.c
```diff
@@ -1418,9 +1395,7 @@ void lf_connect_to_federates(int socket_descriptor) {
       if (!authenticate_federate(&socket_id)) {
         lf_print_warning("RTI failed to authenticate the incoming federate.");
         // Close the socket.
```
hokeun, line 1397: Ditto.

core/federated/RTI/rti_remote.c
```diff
@@ -1488,8 +1463,7 @@ void* respond_to_erroneous_connections(void* nothing) {
       lf_print_warning("RTI failed to write FEDERATION_ID_DOES_NOT_MATCH to erroneous incoming connection.");
     }
     // Close the socket.
```
hokeun, line 1465: Here as well.

hokeun (approved):
Very nice refactoring. Just left minor suggestions.

core/federated/RTI/rti_remote.c
Jakio815: Thank you, I fixed it as suggested.

core/federated/RTI/rti_remote.c
Jakio815: Thank you, I fixed it as suggested.

core/federated/RTI/rti_remote.c
Jakio815: Thank you, I fixed it as suggested.

Jakio815:
The macOS test is not passing for some unknown reason.

The CI test log seems to not work starting from ` Running: src/federated/DistributedCountPhysical.lf (28%)`, however if you run the federated tests on the local machine, it actually doesn't work starting from `Dataflow.lf`.

The reason for not passing is that the program blocks on the final shutdown phase.
 
First, this is how the LF code is intended to work.
1. The federate sends a `RESIGN` signal.
2. The RTI calls `shutdown(socket, SHUT_WR)`, and then immediately calls `read()`
    - The `shutdown()` sends a `FIN` packet to the federate.
3. The federate was blocked on `read()` to listen messages from the RTI.
    - The federate receives the `FIN` packet, and finishes reading the data in the socket buffer, and then returns EOF.
4. The federate will also send a `FIN` packet and the RTI's `read()` will return 0.

This normally works on Linux, thus passing all tests. However, this fails on Mac. 
Where it fails is exactly in step 3. The federate received a `FIN` packet, and the `read()` should return EOF, however it does not, and gets blocked forever.

I also checked that the `FIN` packet was properly sent, and the `ACK` from the federate to the RTI was also sent.

```
// netstat -ant | grep 15045
tcp4       0      0  127.0.0.1.15045        127.0.0.1.64842        FIN_WAIT_2
tcp4       0      0  127.0.0.1.64842        127.0.0.1.15045        CLOSE_WAIT
```

The RTI(15045) is in `FIN_WAIT_2` state, which means that it sent the `FIN`, and received the `ACK`.
The federate(64842) is in `CLOSE_WAIT` which means that the remote side's(RTI) connection is closed, and waiting to connect its own connection.



This error can be reproduced by the lingua-franca's repo on branch `refactor-only-comm-type` and reactor-c branch on `shutdown`, and it should be done on a MacOS machine.

hokeun (approved):
Very nice and clean!

Jakio815:
@edwardalee Would you please review this?

core/federated/federate.c
```diff
@@ -665,7 +652,9 @@ static int handle_tagged_message(int* socket, int fed_id) {
                      env->current_tag.time - start_time, env->current_tag.microstep, intended_tag.time - start_time,
                      intended_tag.microstep);
       // Close socket, reading any incoming data and discarding it.
-      close_inbound_socket(fed_id, 1);
+      close_inbound_socket(fed_id);
+      LF_MUTEX_UNLOCK(&env->mutex);
+      return -1;
```
Jakio815, line 657: Please check description on PR.

core/federated/federate.c
```diff
@@ -403,28 +403,15 @@ static trigger_handle_t schedule_message_received_from_network_locked(environmen
...
-      if (flag > 0) {
-        shutdown(_fed.sockets_for_inbound_p2p_connections[fed_id], SHUT_RDWR);
-      } else {
-        // Have received EOF from the other end. Send EOF to the other end.
-        shutdown(_fed.sockets_for_inbound_p2p_connections[fed_id], SHUT_WR);
-      }
-    }
-    close(_fed.sockets_for_inbound_p2p_connections[fed_id]);
-    _fed.sockets_for_inbound_p2p_connections[fed_id] = -1;
+    shutdown_socket(&_fed.sockets_for_inbound_p2p_connections[fed_id], true);
```
Jakio815, line 414: This changes logic to `read()` the incoming bytes, then `close()`. Please check PR descriptions.

include/core/federated/network/socket_common.h
```diff
@@ -240,4 +240,14 @@ int write_to_socket_close_on_error(int* socket, size_t num_bytes, unsigned char*
...
                                    char* format, ...);
 
+/**
+ * Shutdown and close the socket. If read_before_closing is false, it just immediately calls shutdown() with SHUT_RDWR
+ * and close(). If read_before_closing is true, it calls shutdown with SHUT_WR, only disallowing further writing. Then,
+ * it calls read() until EOF is received, and discards all received bytes.
+ * @param socket Pointer to the socket descriptor to shutdown and close.
+ * @param read_before_closing If true, read until EOF before closing the socket.
+ * @return int 0 for success and -1 for an error.
+ */
```
edwardalee, line 250: It looks like this function assumes the `lf_outbound_socket_mutex` mutex lock is held when the function is called.  This needs to be stated in the documentation and checked for all calls to the function.

edwardalee (approved):
I've requested what I think is just a documentation change, but also a sanity check to make sure the right mutex is held by all callers of this new function.

include/core/federated/network/socket_common.h
Jakio815: This function doesn't really assume the mutex lock is held **always**.

In critical socket shutdown cases, they do assume there is a mutex lock, for example, `federate.c`'s `close_inbound_socket()`, `close_outbound_socket()`, and `rti_remote.c`'s `handle_federate_failed()`, `handle_federate_resign()` does hold the mutexes.

However, some corner cases that handle abnormal termination, do not assume that there are mutex locks.
For example, in `federate.c`'s `listen_to_rti_TCP()` function, when the 1byte header read fails, it does not lock any mutexes, and just shutdowns the socket.
Also, in `rti_remote.c`'s `respond_to_erroneous_connections()` function, which is running in a separate thread and 

I think that the user should handle the mutex if it is needed. Also, in this PR, I did not want to add any mutex handling logic.

Of course, I could add one line in the documentation of this function, such as 
`Mutex locks should be considered before calling this function.`
or would there be any other suggestions?

edwardalee: Is the code correct without this specific mutex?  Note that it can't be any mutex...  All callers have to use the same mutex...

Jakio815: Sorry, professor, I think I didn't understand your question in the first place.

What I understood was, "Don't you need this `lf_outbound_socket_mutex` inside this function?" Am I correct? For that question, as I explained, some cases need mutexes, and some cases do not. So, I thought that the function caller should set the mutex on their own.

If I misunderstood your question, could you please clarify your question?

edwardalee: Among other things, the function does this:
```
int shutdown_socket(int* socket, bool read_before_closing) {
  if (*socket == -1) {
    lf_print_log("Socket is already closed.");
    return 0;
  }
  ...
  *socket = -1;
  return 0;
}
```
If this function can be called from multiple threads, then two threads can _both_ get past the first check and then proceed to try to simultaneously close the socket.  The only way this code is safe is if you can assure that no more than one thread can call this function.

If more than one thread can call the function, then the above code is only correct if _all_ threads that can call the function acquire the same mutex before calling it.  Otherwise, the code is not correct.

One way to ensure safety is to assure that anywhere this is called without acquiring a mutex, only one thread is running.  Alternatively, you would need to assure that no other thread that might be running can call this function.  This may be quite difficult to assure.

One solution would be to acquire a mutex within this function.  But this may be inefficient because some callers may already hold the relevant mutex.  In the absence of such a local mutex, you would need to assure that all callers hold the mutex, and it would have to be clearly documented that this is required to call this function.

Jakio815: @edwardalee 

I just added another mutex in the `shutdown_socket()` function, named as `socket_mutex`. The new changes are updated under the PR description's UPDATE section, and I also copied the same content below. Would the be fine?

## Minor change of name of mutex.
`socket_mutex` was only used as federate's inbound socket mutex. So, the name was changed to `lf_inbound_socket_mutex`, and moved from `socket_common.c` to `federate.c`.

## Add `shutdown_mutex` inside `shutdown_socket()`.
To ensure that the `shutdown_socket()` function does not create a deadlock, `shutdown_mutex` is added. This introduces some inefficiency in acquiring mutexes; however, it is safer.

edwardalee: OK, this change looks good to me.  It looks like there are no blocking calls nor acquisition of any other mutex within `shutdown_socket`, and the `shutdown_mutex` is used only within `shutdown_socket` so there should be no deadlock risk between this mutex and others.

edwardalee (approved):
LGTM!

Jakio815:
@edwardalee Please merge with lf-lang/lingua-franca#2474.

PR #508, opened 2025-01-15
Addition of network abstraction layer to separate socket implementation code from network communication logic

Jakio815:
This should be merged with lf-lang/lingua-franca#2455.

# Summary
This PR adds a network interface layer to support interoperability in terms of network protocols and security using the `net_abstraction`.

The original code is very tied up with sockets, making it hard to add different network features.
However, I clarify that this PR **does not** remove all socket related functions out of the main flow from the RTI and federate, for two reasons.

1. The RTI and federate support user-specified port numbers and also IP addresses for the federate.
2. The `MSG_TYPE_ADDRESS_ADVERTISEMENT` and `MSG_TYPE_ADDRESS_QUERY` uses the port and IP address in the protocol itself.

We can completely remove socket-related code using `#ifdef COMM_TYPE_TCP`; however, I highlight that this PR is more concentrated on supporting end-to-end pluggable and interchangeable network security based on TCP. Thus, in this PR, I **have not** added the `#ifdef` guards, and **have not** changed the protocol for `MSG_TYPE_ADDRESS_ADVERTISEMENT` and `MSG_TYPE_ADDRESS_QUERY`.

One more important thing is that the RTI now does not increment the port number, when the `DEFAULT_PORT` `15045` is unavailable.
# Features
## Plugin API for `network`
This PR creates a separately-compiled library for the `network` instead of compiling it as a part of the core runtime.
I followed the prior work on `low_level_platform.h` and `platform.h`.
All source files related to networking are moved to `network/impl/src`, and all headers are under `network/api`. There are separate `CMakelists.txt` for each.

## Add `COMM_TYPE` target property.
The `comm-type` keyword is available in the `C` target as follows.
```
target C {
  comm-type: TCP
}
```
Currently only supports `TCP`, and plan to support `SST` for security.

## Refactoring on clock-synchronization.
Cloxk synchronization still uses UDP. So, I left all clock-sync functions to directly use UDP sockets, and refactored these functions.

`rti_remote.c`
- `send_physical_clock()` and `handle_physical_clock_sync_message()` : Uses boolean flag `use_UDP` when UDP socket used.

`clock-sync.c`
- `handle_T4_clock_sync_message()` : Uses boolean flag `use_UDP` when UDP socket used.
- `handle_T1_clock_sync_message()`: Use `void* socket_or_net_abstraction` as parameter to support both socket and network abstraction.

## `handle_address_ad()` and `handle_address_query()` 
**There are no changes in the protocol.**

As explained in the summary, `MSG_TYPE_ADDRESS_ADVERTISEMENT` and `MSG_TYPE_ADDRESS_QUERY` use port numbers and IP addresses in the protocol itself. To explain further, for physical connections or decentralized coordination, the `federateA` has to know the peer `federateB`'s port number and IP address to directly connect to it. This is done by `federateA` sending a `MSG_TYPE_ADDRESS_ADVERTISEMENT` to the `RTI` it's port, and the `RTI` sends a `MSG_TYPE_ADDRESS_QUERY_REPLY` message to the peer `federateB`, including the port number and IP address. Thus, the port number and IP address itself cannot be encapsulated under the network abstraction layer, as it is included in the protocol.

Therefore, there are some `get()` and `set()` calls to the network interface. 
1. `FedA` - > `RTI`: `MSG_TYPE_ADDRESS_ADVERTISEMENT`: `FedA` calls `get_my_port()` to send its own port to the `RTI`.
4. `FedB` -> `RTI`: `MSG_TYPE_ADDRESS_QUERY` : No changes.
5. `RTI` -> `FedB`: `MSG_TYPE_ADDRESS_QUERY_REPLY` : `RTI` calls `get_server_port()` and `get_ip_addr()` to encode it to the message to send to `FedB`.
6. `FedB`: Sets the received port and IP address by `set_server_port()` and `set_server_host_name()` to directly connect to `FedA`.

## Minor logic change: Move `getpeername()` logic to `accept_net_abstraction()`
The RTI should know the connected federate(FedA)'s IP address, to pass the IP address to the other federate(FedB), as explained above. 

Before, this was done in `rti_remote.c`'s `receive_and_check_fed_id_message()`. I moved this to `lf_socket_support.c`'s `accept_net_abstraction()`, because it looks more appropriate to set the connected peer's information inside this function.

One inefficiency that happens is that in decentralized coordination, the server federate does not need to save the connected peer federate's IP address. However, I think this will barely affect the performance.

## Add default UDP port number as 15061.
The UDP port was usually set to the `RTI's port + 1`, in `rti_remote.c`'s `create_server()` function call. However, I did not want to expose the port number in the `create_server()` interface, so I took the parameter `port` out from `create_server()`.

## Port increment logic removed.
When the default port is not available, it used to increment ports and used that one.
However, we decided to remove this feature, and use a specified port as the hash number of the federation, but this will be done in a separate PR.

## Minor changes
- The `start_rti_server()` function does not get the `port` as a parameter. The `port` will be saved in `rti_remote_t`;s `user_specified_port`, with a default value when not set up.


core/federated/federate.c
```diff
@@ -1759,17 +1724,25 @@ void lf_connect_to_federate(uint16_t remote_federate_id) {
...
-  if (connect_to_socket(socket_id, (const char*)hostname, uport) < 0) {
+
+  // Create a network driver.
+  netdrv_t* netdrv = initialize_netdrv();
+  // Set the received host name and port to the network driver.
+  set_server_port(netdrv, uport);
+  set_server_host_name(netdrv, hostname);
+  // Create the client network driver.
+  create_client(netdrv);
+  if (connect_to_netdrv(netdrv) < 0) {
```
Jakio815, line 1735: Check the PR description please.

core/federated/federate.c
```diff
@@ -1835,9 +1808,15 @@ void lf_connect_to_rti(const char* hostname, int port) {
...
-  if (connect_to_socket(_fed.socket_TCP_RTI, hostname, port) < 0) {
+  // Create a network driver.
+  _fed.netdrv_to_RTI = initialize_netdrv();
+  // Set the user specified host name and port to the network driver.
+  set_server_port(_fed.netdrv_to_RTI, port);
+  set_server_host_name(_fed.netdrv_to_RTI, hostname);
+
+  // Create the client network driver.
+  create_client(_fed.netdrv_to_RTI);
+  if (connect_to_netdrv(_fed.netdrv_to_RTI) < 0) {
```
Jakio815, line 1819: Check the PR description please.

core/federated/RTI/rti_remote.h
```diff
@@ -113,22 +106,23 @@ typedef struct rti_remote_t {
   const char* federation_id;
 
   /************* TCP server information *************/
-  /** The desired port specified by the user on the command line. */
+  /** The desired port specified by the user on the command line.
+   * This should be not moved to the net_driver, because the user can configure this as -p or --port.
+   */
   uint16_t user_specified_port;
```
Jakio815, line 112: Please check minor changes section in PR description for more details.

edwardalee (changes requested):
I haven't made it all the way though, but it's looking pretty good! One issue is that while I think "network abstraction" is a good term to use for the general architecture, I find the resulting names unnecessarily long.  I suggest dropping the "abstraction" in most cases.  E.g.:

fed->fed_net_abstraction -> fed->net
write_to_net_abstraction -> write_to_net
write_to_net_abstraction_fail_on_error -> write_to_net_fail_on_error
read_from_net_abstraction_fail_on_error -> read_from_net_fail_on_error

network/impl/src/socket_common.c
```diff
@@ -10,16 +21,16 @@
 #include <stdarg.h> //va_list
 #include <string.h> // strerror
 
-#include "util.h"
-#include "socket_common.h"
+#include "util.h" // LF_MUTEX_UNLOCK(), logging.h
```
hokeun:
```suggestion
#include <logging.h>
#include "util.h" // LF_MUTEX_UNLOCK()
```

core/federated/clock-sync.c
```diff
@@ -192,28 +192,28 @@ void synchronize_initial_physical_clock_with_rti(int* rti_socket_TCP) {
...
     // Handle the message.
-    handle_T4_clock_sync_message(buffer, *rti_socket_TCP, receive_time);
+    handle_T4_clock_sync_message(buffer, (void*)rti_net, receive_time, false);
   }
 
   LF_PRINT_LOG("Finished initial clock synchronization with the RTI.");
 }
 
-int handle_T1_clock_sync_message(unsigned char* buffer, int socket, instant_t t2) {
+int handle_T1_clock_sync_message(unsigned char* buffer, void* socket_or_net, instant_t t2, bool use_udp) {
```
hokeun: The argument name here is inconsistent with the one above: use_UDP vs. use_udp.

core/CMakeLists.txt
```diff
@@ -96,10 +95,19 @@ include(${LF_ROOT}/platform/impl/CMakeLists.txt)
...
 
+if(DEFINED FEDERATED)
+    if(NOT DEFINED COMM_TYPE)
+        set(COMM_TYPE TCP)
+    endif()
+    include(${LF_ROOT}/network/api/CMakeLists.txt)
+    include(${LF_ROOT}/network/impl/CMakeLists.txt)
+    target_link_libraries(reactor-c PUBLIC lf::network-api)
+    target_link_libraries(reactor-c PRIVATE lf::network-impl)
+endif()
```
hokeun: Shouldn't this be merged into a single `if(DEFINED FEDERATED)` ?

hokeun (approved):
Looks great to me! Just left one minor comment.

core/CMakeLists.txt
Jakio815: I had a look again at this file, and it looks like these conditionals need to remain separated because they apply at different stages of the CMake configuration process.

The CMake process is
1. Assemble the a list of the source of `reactor-c`
2. Create the `reactor-c` library using the assembled sources.
3. Link the additional libraries to the `reactor-c` library.

The first part on line 19, which is a part of step 1.
```
# Add sources for either threaded or single-threaded runtime
if(DEFINED FEDERATED)
    include(federated/CMakeLists.txt)
endif()
```
This part is where the source files are added to the `REACTORC_SOURCES`. The CMake is still assembling the list of sources yet.

However on line 98, this part is step 3.
```
if(DEFINED FEDERATED)
    if(NOT DEFINED COMM_TYPE)
        set(COMM_TYPE TCP)
    endif()
    include(${LF_ROOT}/network/api/CMakeLists.txt)
    include(${LF_ROOT}/network/impl/CMakeLists.txt)
    target_link_libraries(reactor-c PUBLIC lf::network-api)
    target_link_libraries(reactor-c PRIVATE lf::network-impl)
endif()
```
This part is responsible for linking the network related funcitons to the `reactor-c` library, which was created on line 51 `add_library(reactor-c)`, which is step 2.

So, taking the first part (line 19) to the second part (line 98) wouldn't work, and also vice versa.

Do you still suggest fixing this part?

Jakio815:
> > I haven't made it all the way though, but it's looking pretty good! One issue is that while I think "network abstraction" is a good term to use for the general architecture, I find the resulting names unnecessarily long. I suggest dropping the "abstraction" in most cases. E.g.:
> > fed->fed_net_abstraction -> fed->net write_to_net_abstraction -> write_to_net write_to_net_abstraction_fail_on_error -> write_to_net_fail_on_error read_from_net_abstraction_fail_on_error -> read_from_net_fail_on_errorHello
> 
> Hello Prof. Lee, thank you for your comment. I shortened most of the names. The one thing I haven't changed was the name of the type, `net_abstraction_t`. Besides this, most of them are shortened.
> 
> net_abstractions_for_inbound_p2p_connections -> net_for_inbound_p2p_connections net_abstractions_for_outbound_p2p_connections -> net_for_outbound_p2p_connections lf_outbound_net_abstraction_mutex -> lf_outbound_net_mutex fed->fed_net_abstraction --> fed->net net_abstraction_to_RTI -> net_to_RTI RTI_net_abstraction_listener -> RTI_net_listener

@edwardalee Hello, Professor. Do you mind taking a look once again? I finally fixed all CI tests failing problems.

network/api/net_abstraction.h
```diff
@@ -0,0 +1,290 @@
...
+ * @return net_abstraction_t Initialized network abstraction.
+ */
+net_abstraction_t initialize_net();
+
+/**
+ * @brief Create a server network abstraction that will accept incoming connections.
+ * @ingroup Network
+ *
+ * Create a network abstraction server. This is such as a server socket which accepts connections.
+ * However this is only the creation of the server network abstraction.
```
edwardalee:
```suggestion
 * Create a network abstraction server. For example, this might be a server socket that accepts connections.
```

network/api/net_abstraction.h
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * @param net_abs Server's network abstraction.
+ * @param increment_port_on_retry Whether to increment the port on retry if binding fails.
```
edwardalee: This argument suggests a "leaky abstraction". The argument only applies to socket connections and is only used by the RTI on its rather funky mechanism where it increments the port if the `DEFAULT_PORT` is not available. This functionality could be moved to `rti_remote.c`, which has this:
```
if (create_server(rti_remote->rti_net, true)) {
    lf_print_error_system_failure("RTI failed to create TCP server: %s.", strerror(errno));
    return -1;
  };
```
If the retry functionality were put here and removed from deep inside `socket_common.c`, then this API could be cleaned up by removing this argument.

network/api/net_abstraction.h
```diff
@@ -0,0 +1,290 @@
...
+net_abstraction_t initialize_net();
+
+/**
+ * @brief Create a server network abstraction that will accept incoming connections.
+ * @ingroup Network
+ *
+ * Create a network abstraction server. This is such as a server socket which accepts connections.
+ * However this is only the creation of the server network abstraction.
+ *
+ * @param net_abs Server's network abstraction.
```
edwardalee:
```suggestion
 * @param net_abs Server's network abstraction as returned by `initialize_net`.
```

network/api/net_abstraction.h
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * @brief Accept an incoming connection on a server network abstraction.
+ * @ingroup Network
+ *
+ * Wait for an incoming connection request on the specified server network abstraction.
+ * The implementation should include three steps.
+ * 1. Initialize the network abstraction of the connected federate.
+ * 2. Wait for the incoming connection request. This should block until the connection is successfully accepted.
+ * 3. Save the information in the connected network abstraction, such as the address of the connected peer, for future
+ * querying address.
```
edwardalee: These steps are unclear. The first step implies there is already a connected federate.
Also, why is this even referencing a federate or RTI? Shouldn't this abstraction be just about establishing a connection between two entities, without regard to what they are?

Also, this comment should state that `create_server` must be called first.

Also, I think the `rti_chan` argument should be removed and its functionality put where it is actually needed, which is in `federate.c` where `accept_net` is called.  Shouldn't this handle checking the RTI and retrying? This may require another function like `is_connected(net_abstraction_t)` in your `net_abstraction`.

edwardalee (changes requested):
Apologies for the delay.

This is really a significant piece of work. This time, I focused on `net_abstraction.h` and I have a number of questions. Some of those I've entered in the code. Overall, this looks like a good first step, but the abstraction still looks pretty "leaky" to me.  There are a lot of specifics that leak into the abstraction, like arguments that are there only because the RTI and the federates handle error conditions differently.  It seems like this abstraction should not be making any distinction between RTI, federate, or any other user of the abstraction. Let the caller handle errors however they like.

Most importantly, `net_abstraction.h` needs some documentation on how to use it.
Writing such documentation might reveal some weaknesses in the abstraction design.

What you currently do is this, for example:

```
  // Create a network abstraction.
  net_abstraction_t net = initialize_net();
  // Set the received host name and port to the network abstraction.
  set_server_port(net, uport);
  set_server_hostname(net, hostname);
  // Create the client network abstraction.
  create_client(net);
  if (connect_to_net(net) < 0) {
    lf_print_error_and_exit("Failed to connect to federate.");
  }
```

(there is nearly identical code in `connect_to_RTI`)

What this is really doing is this:

```
  // Allocate a network abstraction.
  net_abstraction_t net = initialize_net();
  // Set the received host name and port to the network abstraction.
  ((socket_priv_t*)net)->server_port = uport;
  ((socket_priv_t*)net)->server_hostname = hostname;
  // Create the client network abstraction.
  create_client(net);
  if (connect_to_net(net) < 0) {
    lf_print_error_and_exit("Failed to connect to federate.");
  }
```

It is confusing to me to initialize, then initialize some more, then create, then connect.
Also, `create_client` is always followed by `connect_to_net`, so why are these two different functions?  Maybe it be better to do something like this:

```
  socket_connection_parameters_t params;
  params.type = "TCP";
  params.port = uport;
  params.server_hostname = hostname;
  net_abstraction_t net = connect_to_net((net_params_t*)&params);
  if (net == NULL) {
    lf_print_error_and_exit("Failed to connect to federate.");
  }
```

The base type `net_params_t` would have whatever fields are needed by all uses, such as the `type`.

Better yet, this could be abstracted for TCP usage, so the user just does this:

```
  net_abstraction_t net = connect_via_TCP(hostname, uport);
  if (net == NULL) {
    lf_print_error_and_exit("Failed to connect to federate.");
  }
```

This function would then do what the previous code does.
The function `connect_via_TCP` would not be part of your API, hence not declared in `net_abstraction.h`.


More questions:

Is `get_socket_priv_t` really necessary? All it does is a cast, and it seems to obscure the code. It doesn't even avoid the segfault if the argument is NULL.

It's not clear what `check_net_closed` will do if called on a network interface that has not been opened.  Maybe a better function would be `is_net_open` or `is_connection_open`?

Regarding `MSG_TYPE_ADDRESS_ADVERTISEMENT` and `MSG_TYPE_ADDRESS_QUERY`, perhaps in a future PR these could also be abstracted. The protocol could be augmented (or just changed) to support an advertisement that could be a simple string message, to be interpreted appropriately by the client. This could in fact allow for adaptable connections. E.g., the advertisement could be a string like: "TCP:my_ip_address:my_port"

For now, I don't think `get_my_port` or `get_server_port` or `get_ip_addr` or `get_server_hostname` should be part of this abstraction. The users of these functions already know what struct to cast to. Same for all the corresponding set functions.
If you want functions, make them static functions in `rti_remote.c`.

network/api/net_abstraction.h
```diff
@@ -0,0 +1,290 @@
...
+ * connection management, robust read/write with error handling, graceful shutdown, and
+ * querying local/peer addressing information.
+ */
+
+#ifndef NET_ABSTRACTION_H
+#define NET_ABSTRACTION_H
+
+#include "socket_common.h"
+
+typedef void* net_abstraction_t;
```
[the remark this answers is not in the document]
Jakio815: Added this comment.

Jakio815:
Question 2. 
> Is `get_socket_priv_t` really necessary? All it does is a cast, and it seems to obscure the code. It doesn't even avoid the segfault if the argument is NULL.

Yes, you are totally correct, I'll fix this.

Jakio815:
Question 5.
> For now, I don't think `get_my_port` or `get_server_port` or `get_ip_addr` or `get_server_hostname` should be part of this abstraction. The users of these functions already know what struct to cast to. Same for all the corresponding set functions. If you want functions, make them static functions in `rti_remote.c`.

Okay, then in this point, I'll first remove them, and just directly set it to the struct. 
This is done here. [24aec1d](https://github.com/lf-lang/reactor-c/pull/508/commits/24aec1db3b71a1c4b02ae4279ba1e4afda6dfd4a)

Jakio815:
Non-deterministic failing tests:

`src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf`
`src/enclave/ContainedEnclaveToContainedEnclave.lf`
`src/federated/DistributedStopDecentralized.lf`
`src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf`

edwardalee:
> Non-deterministic failing tests:
> 
> `src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf` `src/enclave/ContainedEnclaveToContainedEnclave.lf`

You can move the enclave test to `failing`. The other one is flaky (by design, unfortunately). I've been just re-running it, but maybe there is a way to make it less flaky.

network/api/net_abstraction.h
```diff
@@ -0,0 +1,222 @@
...
+#ifndef NET_ABSTRACTION_H
+#define NET_ABSTRACTION_H
+
+#include "socket_common.h"
+
+/**
+ * @brief Pointer to whatever data structure is used to maintain the state of a network connection or service.
+ * @ingroup Network
+ *
+ * Pointer to whatever data structure is used to maintain the state of a network connection or service.
```
edwardalee:
```suggestion
```

(no need to repeat `@brief`.)

network/api/net_abstraction.h
```diff
@@ -0,0 +1,222 @@
...
+ * For example, it may point to a socket_connection_params_t structure
+ * when using a socket-based network implementation.
+ */
+typedef void* net_params_t;
+
+/**
+ * @brief Allocate and initialize a network abstraction handle.
+ * @ingroup Network
+ *
+ * Allocate memory for the network abstraction.
```
edwardalee:
```suggestion
```

network/api/net_abstraction.h
```diff
@@ -0,0 +1,222 @@
...
+ * @return net_abstraction_t Initialized network abstraction.
+ */
+net_abstraction_t initialize_net();
+
+/**
+ * @brief Create a server network abstraction that will accept incoming connections.
+ * @ingroup Network
+ *
+ * Create a network abstraction server. For example, this might be a server socket that accepts connections.
```
edwardalee:
```suggestion
 * @brief Create a server network server that will accept incoming connections.
 * @ingroup Network
 *
 * For example, this might be a server socket that accepts connections.
```

edwardalee (approved):
LGTM!  I've requested a Copilot review, which will probably have a lot of suggestions. Feel free to dismiss them if they are out of scope or questionable. I made small suggested changes in the docs.

edwardalee (changes requested):
This review from Copilot identifies some potential bugs that may not show up in testing.

core/federated/federate.c
```diff
@@ -1959,42 +1931,47 @@ void lf_connect_to_rti(const char* hostname, int port) {
...
+                             "Failed to send the UDP port number to the RTI.");
 }
 
 void lf_create_server(int specified_port) {
   assert(specified_port <= UINT16_MAX && specified_port >= 0);
-  uint16_t final_port;
-  if (create_server(specified_port, &_fed.server_socket, &final_port, TCP, false)) {
-    lf_print_error_system_failure("RTI failed to create TCP server: %s.", strerror(errno));
+
+  net_abstraction_t* server_net = initialize_net();
```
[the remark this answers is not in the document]
Jakio815: Fixed.

Jakio815:
> This review from Copilot identifies some potential bugs that may not show up in testing.

Could you let me know which one? I went through all of them, fixed things that looked correct, but also dismissed a few. Just want to make sure I did fix the one you thought was a potential bug.

edwardalee:
> > This review from Copilot identifies some potential bugs that may not show up in testing.
> 
> Could you let me know which one? I went through all of them, fixed things that looked correct, but also dismissed a few. Just want to make sure I did fix the one you thought was a potential bug.

I was concerned about the "free before use" issue that it reported.

core/federated/federate.c
```diff
@@ -1531,40 +1507,27 @@ static void* listen_to_rti_TCP(void* args) {
...
-      }
+      lf_print_error("Connection to the RTI was closed by the RTI with an error. Considering this a soft error.");
+      shutdown_net(_fed.net_to_RTI, false);
+      return NULL;
     } else if (read_failed > 0) {
       // EOF received.
       lf_print_info("Connection to the RTI closed with an EOF.");
-      shutdown_socket(&_fed.socket_TCP_RTI, false);
+      shutdown_net(_fed.net_to_RTI, false);
       return NULL;
```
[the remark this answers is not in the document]
Jakio815: Fixed this and all following usages of shutdown_net().
Done here. [08d7207](https://github.com/lf-lang/reactor-c/pull/508/commits/08d72077b56d933bead575bd5ccfee92262ef175)

Jakio815:
> I was concerned about the "free before use" issue that it reported.

I guess you meant this one.

>> `listen_to_rti_net()` calls `shutdown_net(_fed.net_to_RTI, ...)`, which frees the underlying network abstraction, but `_fed.net_to_RTI` is not set to NULL afterward. Subsequent checks/use of `_fed.net_to_RTI` can become use-after-free. Set `_fed.net_to_RTI = NULL` immediately after `shutdown_net()` (and consider making `shutdown_net()` accept a `net_abstraction_t*` so the callee can null it).

Fixed this and all following usages of `shutdown_net()`.
Done here. [08d7207](https://github.com/lf-lang/reactor-c/pull/508/commits/08d72077b56d933bead575bd5ccfee92262ef175)

edwardalee (approved):
Looks good! Let's merge!

PR #509, opened 2025-01-20
Move CI to ubuntu-24.04

erlingrj:
See https://github.com/lf-lang/lingua-franca/pull/2458

### DIFF
--- a/.github/workflows/build-rti.yml
+++ b/.github/workflows/build-rti.yml
@@ -7,7 +7,7 @@ jobs:
   native-build:
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


edwardalee (approved):
I'm OK with this, but do we really want to freeze the ubuntu version we use?

erlingrj:
I think the best strategy is to pin the OS version and then update it when the new releases have stabilized. The advantage is that we decide when to port our CI workflows to the new version of Ubuntu, without it affecting any other PR. 

The current status is that our CI is broken and no PR's will pass CI until these fixes are merged.

ChadliaJerad:
@Jakio815, @erlingrj, I think this PR #510 will fix the issue.

Issue #513, opened 2025-01-28
Optimistic `MIN_SLEEP_DURATION` causes a problem


byeonggiljun:
[`MIN_SLEEP_DURATION`](https://github.com/lf-lang/reactor-c/blob/856e98a86cc4af30b9891ee0a47c5d9b04832670/include/core/reactor_common.h#L45) causes a flaky error in [DecentralizedP2PUnbalancedTimeoutPhysical.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/Python/src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf) (see this [CI test result](https://github.com/lf-lang/reactor-c/actions/runs/12959482104/job/36151776068)).

In the below logs, you can see that the federate doesn't wait until the physical time of `1 msec` because the waiting time is less than `MIN_SLEEP_DURATION` (`10 usec`). However, this leads to advancing the current tag to `1 msec` before the physical time exceeds `1 msec` (`994.0 usec`). However, a message came through the physical connection, was scheduled at `996.6 usec`, and invoked a fatal error.

```
DEBUG: -------- Waiting until physical time 1000000.
DEBUG: Wait time 6608 is less than MIN_SLEEP_DURATION 10000. Skipping wait.
DEBUG: Waiting is not interrupted. Current physical time is 993743.
DEBUG: Advanced (elapsed) tag to (1000000, 0) at physical time 993983.
Fed 1 (d): FATAL ERROR: DEBUG: get_next_event_tag(): Earliest event on the event queue ((996593, 0)) is earlier than the current tag ((1000000, 0)).
```

@edwardalee Do you think we can fix this by simply decreasing the amount of `MIN_SLEEP_DURATION`? Then, what should be the new value?

erlingrj:
I think we should remove MIN_SLEEP_DURATION, it was a premature optimization.

byeonggiljun:
I agree. But I found the PR https://github.com/lf-lang/reactor-c/pull/403 related to `MIN_SLEEP_DURATION`. This PR will fix the issue because it adds a busy wait when the wait duration is less than `MIN_SLEEP_DURATION`. So, why don't we merge that PR?

edwardalee:
Wow, great detective work!  The goal of `MIN_SLEEP_DURATION` and #403 was to reduce lag.  I'm wondering if this bug could also be the cause for the [watchdog failures](https://github.com/lf-lang/lingua-franca/issues/2395).

[...]

PR #514, opened 2025-01-28
Perform busy waiting when the wait duration is less than `MIN_SLEEP_DURATION`

byeonggiljun:
This fixes https://github.com/lf-lang/reactor-c/issues/513. 

Add busy waiting when the wait duration is less than `MIN_SLEEP_DURATION` to prevent the reactor from advancing to a logical time greater than the current physical time.

### DIFF
--- a/core/threaded/reactor_threaded.c
+++ b/core/threaded/reactor_threaded.c
@@ -199,6 +199,11 @@ bool wait_until(instant_t wait_until_time, lf_cond_t* condition) {
     LF_PRINT_DEBUG("-------- Waiting until physical time " PRINTF_TIME, wait_until_time - start_time);
     // Check whether we actually need to wait, or if we have already passed the timepoint.
     interval_t wait_duration = wait_until_time - lf_time_physical();
+    if (wait_duration < MIN_SLEEP_DURATION) {
+      LF_PRINT_DEBUG("Wait time " PRINTF_TIME " is less than MIN_SLEEP_DURATION " PRINTF_TIME ". Skipping wait.",
+                     wait_duration, MIN_SLEEP_DURATION);
+      return true;
+    }
 
     // We do the sleep on the cond var so we can be awakened by the
     // asynchronous scheduling of a physical action. lf_clock_cond_timedwait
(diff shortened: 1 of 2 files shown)


byeonggiljun:
@erlingrj I ended up creating a new PR instead of https://github.com/lf-lang/reactor-c/pull/403 because that PR is WIP. I have two questions for you.

1. What do you prefer between removing `MIN_SLEEP_DURATION` and adding busy waiting?
2. If we're going to add busy waiting, should we adjust the value of `MIN_SLEEP_DURATION`? In other words, is busy waiting for at most `10 usec` too much?

erlingrj:
Hmm, I think doing busy-waiting is fine, but I think it should be viewed as a real-time optimization, the purpose of it is to reduce lag. That might not be always what is wanted, however, we only do this a short period of time anyways. Another argument is that if the users are writing real-time programs they have to configure the underlying platform to achieve it. If they are using Linux, this means enabling PREEMPT_RT and using a real-time scheduling policy for the process that runs the LF program.

byeonggiljun:
Surprisingly, either removing MIN_SLEEP_DURATION ([CI184b55f#](https://github.com/lf-lang/reactor-c/actions/runs/13035385263/job/36364423963)) or adding busy waiting ([CI test#fd7bd22](https://github.com/lf-lang/reactor-c/actions/runs/13017669821/job/36336767221)) in `wait_until` led to test failures in `window-latest` while the error didn't occur when I revert the change.

erlingrj:
My guess here is that we try going to sleep for a negative duration and the Windows platform implementation does not support that. I suggest that we verify that the sleep duration is greater than zero. (Basically setting MIN_SLEEP_DURATION to zero).

The first failures when adding the busy-wait seems to be different than the ones when you removed it altogheter

erlingrj:
Very wierd Java errors, almost seems like its something with the network?

byeonggiljun:
Yes, it is very weird. https://github.com/RomanIakovlev/timeshape/issues/1 Maybe the busy waiting causes a stack or heap overflow?

> I suggest that we verify that the sleep duration is greater than zero. (Basically setting MIN_SLEEP_DURATION to zero).

Anyway, I'll add your suggestion. Thanks!

byeonggiljun:
Hmm... I found that there's already the code to handle negative time delays below in [`lf_windows_support.c`](https://github.com/lf-lang/reactor-c/blob/main/low_level_platform/impl/src/lf_windows_support.c). Let's first see whether the test passes now.

https://github.com/lf-lang/reactor-c/blob/856e98a86cc4af30b9891ee0a47c5d9b04832670/low_level_platform/impl/src/lf_windows_support.c#L293-L299

byeonggiljun:
Why don't we set `MIN_SLEEP_DURATION` to 0 for now to avoid the flaky error? We can improve the design for real-time optimization in other PRs.

erlingrj (approved):
LGTM

PR #516, opened 2025-02-06
Add option for providing an external implementation of clock functions

erlingrj:
This PR adds support for letting the user provide their own implementation of functions found in `clock.c`. This enables locking the physical time the runtime sees, to a simulator.

### DIFF
--- a/core/clock.c
+++ b/core/clock.c
@@ -17,7 +17,6 @@ void clock_sync_add_offset(instant_t* t) { (void)t; }
 void clock_sync_subtract_offset(instant_t* t) { (void)t; }
 #endif // defined(FEDERATED)
 
-#ifndef LF_SIMULATE_PHYSICAL_TIME
 static instant_t last_read_physical_time = NEVER;
 int lf_clock_gettime(instant_t* now) {
   instant_t last_read_local;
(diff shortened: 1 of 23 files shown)


core/clock.c
```diff
@@ -57,3 +62,8 @@ int lf_clock_cond_timedwait(lf_cond_t* cond, instant_t wakeup_time) {
   return _lf_cond_timedwait(cond, wakeup_time);
 }
 #endif // !defined(LF_SINGLE_THREADED)
+#else // defined(LF_EXTERNAL_CLOCK_PLUGIN)
+// The following empty "dummy" function is here to avoid an "Empty translation unit" compiler
+// warning if the user has defined LF_EXTERNAL_CLOCK_PLUGIN to provide their own implementation.
+void __clock_dummy_function(void) {}
```
erlingrj, line 68: We are using `Werror=pedantic` which throws an error on empty translation units. This dummy definition avoids that problem.

edwardalee (approved):
It would be nice to add specific instructions for how to make use of this. E.g., "Add X to your target properties, put Y into your cmake include file, then provide your own foo.c file and link it in as follows..."  These could go into a comment in clock.c.

PR #517, opened 2025-02-10
Avoid a double join on the sensor_simulator output thread

erlingrj:
This makes sure that we dont try to join a thread which has already been joined, it is undefined behavior and has sometimes lead to a segmentation fault.


edwardalee (approved):
Wow, another good catch!  Let's merge!

PR #518, opened 2025-02-10
Fix memory bug when obtaining a port number

erlingrj:
The problem here was that the port number is an int32, it was initialized to -1. Then its address was casted to an `uint16_t *` and then written to from the `create_server` function.  However, when a port number was obtained and written, it would only overwrite two of the four bytes. So you were left with the sign bit (set when initialized to -1)


edwardalee (approved):
Wow, good catch! Let's merge this ASAP.

PR #520, opened 2025-04-07
Handle STP violation over deadline when both occur

edwardalee:
This PR makes the following changes to decentralized coordination:
1. When a network input has both an STP violation and a deadline violation, we used to invoke the deadline violation handler only. This PR changes it to invoke the STP violation handler only. The reason for this is that the current tag is incorrect, which is not normally the case in a deadline violation handler. The previous strategy would mask the STP violation, making it impossible to distinguish between lateness caused by local overhead and lateness caused by network issues.
2. Removes spurious warnings "Attempt to update the last known status tag" generated when the background thread updates the last known status tag without taking into account after delays on incoming links.

This has a companion PR: https://github.com/lf-lang/lingua-franca/pull/2487

### DIFF
--- a/core/federated/federate.c
+++ b/core/federated/federate.c
@@ -262,9 +262,10 @@ static void update_last_known_status_on_input_ports(tag_t tag, environment_t* en
  *
  * @param env The top-level environment, whose mutex is assumed to be held.
  * @param tag The tag on which the latest status of the specified network input port is known.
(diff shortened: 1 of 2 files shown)


erlingrj (approved):
This looks good! I agree that in the case that an STP violation has occurred, and the reaction is executed at the wrong tag, then we should prioritize executing STP handler, not reaction body.

PR #521, opened 2025-04-08
Add semantic versioning of the RTI

erlingrj:
See https://github.com/lf-lang/lingua-franca/pull/2490

### DIFF
--- a/core/federated/RTI/CMakeLists.txt
+++ b/core/federated/RTI/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.12)
-project(RTI VERSION 1.0.0 LANGUAGES C)
+project(RTI VERSION 0.0.2 LANGUAGES C)
 
 set(CoreLib ../../../core)
 set(LF_ROOT ${CMAKE_CURRENT_LIST_DIR}/../../..)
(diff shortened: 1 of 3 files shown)


core/federated/RTI/rti.Dockerfile
```diff
@@ -2,7 +2,7 @@ ARG BASEIMAGE=alpine:latest
 FROM ${BASEIMAGE} AS builder
 COPY . /lingua-franca
 WORKDIR /lingua-franca/core/federated/RTI
-RUN set -ex && apk add --no-cache gcc musl-dev cmake make && \
+RUN set -ex && apk add --no-cache gcc musl-dev cmake make git && \
```
erlingrj, line 5: We need git because the build process uses it to get the commit ID of the build which is printed when -v/--version is passed

core/federated/RTI/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ target_include_directories(${RTI_LIB} PUBLIC ${IncludeDir}/modal_models)
 target_include_directories(${RTI_LIB} PUBLIC ${IncludeDir}/utils)
 
 if (NOT DEFINED LOG_LEVEL)
-    set(LOG_LEVEL 0)
+    set(LOG_LEVEL 2)
```
erlingrj, line 37: This change has nothing to do with semantic versioning. But I don't believe we should disable all prints by default. Even the help string when calling `RTI --help` is dropped unless LOG_LEVEL is greater than or equal to 2.

core/federated/RTI/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.12)
-project(RTI VERSION 1.0.0 LANGUAGES C)
+project(RTI VERSION 0.2.0 LANGUAGES C)
```
erlingrj, line 2: I propose we start out with version 0.2.0 instead of 0.1.0 because there are older versions of the RTI, but they just don't have semantic versioning. 0.2.0 indicates that this is not the initial version

edwardalee (approved):
These are important improvements, but I wonder if we should go further.  Perhaps we should add a new message type for initial contact with the RTI that specifies the protocol version that the federate is using.  If a federate contacts the RTI using the old `MSG_TYPE_FED_IDS` message, then the RTI should assume version 0.1.0 of the protocol and disable the DNET message enhancement. Otherwise, it should check whether it supports the version specified and give an informative error message if not.  Perhaps this could be another PR?

erlingrj:
> These are important improvements, but I wonder if we should go further. Perhaps we should add a new message type for initial contact with the RTI that specifies the protocol version that the federate is using. If a federate contacts the RTI using the old `MSG_TYPE_FED_IDS` message, then the RTI should assume version 0.1.0 of the protocol and disable the DNET message enhancement. Otherwise, it should check whether it supports the version specified and give an informative error message if not. Perhaps this could be another PR?

I agree that the proper solution to this issue is to include a notion of a federated protocol version, this should be the first thing that is exchange between federates and RTI. The protocol version must not necessarily be connected to the RTI version. I would say that lfc v0.9.0 uses v1 of the protocol and the DNET message enhancement introduces a v2. 

Adding some logic in reactor-c where each federate requests a protocol version in the initial message is something I can look at. Making the RTI switch its functioning between DNET and not DNET, might be a bigger task, depending on how this feature is architected.

In any case, having a version string printed by the RTI is an important feature that will help diagnostics etc in the future, so I will merge this.

PR #523, opened 2025-04-14
Compile RTI for each federation

erlingrj:
This PR removes the use of CMake options in the RTI CMakeLists.txt. We should be consistent, and the other arguments are handled as normal CMake variables. CMake options are more complicated and we dont need them for this simple use-case.

See: https://github.com/lf-lang/lingua-franca/pull/2492

### DIFF
--- a/core/federated/RTI/CMakeLists.txt
+++ b/core/federated/RTI/CMakeLists.txt
@@ -88,8 +88,10 @@ target_compile_options(${RTI_LIB} PUBLIC -Werror)
 find_package(Threads REQUIRED)
 target_link_libraries(${RTI_LIB} PUBLIC Threads::Threads)
 
-# Option for enabling federate authentication by RTI.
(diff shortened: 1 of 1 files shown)


core/federated/RTI/CMakeLists.txt
```diff
@@ -88,8 +88,10 @@ target_compile_options(${RTI_LIB} PUBLIC -Werror)
 find_package(Threads REQUIRED)
 target_link_libraries(${RTI_LIB} PUBLIC Threads::Threads)
 
-# Option for enabling federate authentication by RTI.
-option(AUTH "Federate authentication by RTI enabled." OFF)
+if (NOT DEFINED AUTH)
```
edwardalee, line 91:
```suggestion
# If AUTH is defined, then the RTI will be able to authenticate federates. By default, this is off.
if (NOT DEFINED AUTH)
```

edwardalee (approved):
Modulo a suggested comment, this looks good.

PR #528, opened 2025-04-29
New `update` policy

lsk567:
This PR introduces a new `update` policy: dropping the already scheduled event, and replacing it with a newly scheduled event.

The new `update` policy enables a useful pattern with minimum spacing — renewing the delay interval of an already scheduled event, which can be used to implement the election timeout mechanism essential to implementing the [Raft](https://raft.github.io) protocol.

### DIFF
--- a/core/clock.c
+++ b/core/clock.c
@@ -11,7 +11,7 @@
  * implementation of the clock functions. This allows controlling the
  * physical time the runtime sees. It is useful for integration with simulators
  * or for repeatable test environments.
- * 
+ *
  * Steps to provide an external clock plugin:
  * 1. Use the cmake-include target property to add a custom CMake file to the build.
  * 2. Add `target_compile_definition(reactor-uc PUBLIC LF_EXTERNAL_CLOCK_PLUGIN)` to the custom CMake file.
--- a/include/core/lf_types.h
+++ b/include/core/lf_types.h
@@ -72,13 +72,13 @@ typedef struct _lf_tag_advancement_barrier {
  * The default policy is `defer`: adjust the tag to that the minimum
  * interarrival time is satisfied.
  * The `drop` policy simply drops events that are scheduled too early.
- * The `replace` policy drops the preceding event and replaces it with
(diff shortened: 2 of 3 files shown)


lsk567:
The CI failure is expected because it is running against an old test suite where `replace` has not been changed to `update`. This change is made in https://github.com/lf-lang/lingua-franca/pull/2499.

edwardalee:
> The CI failure is expected because it is running against an old test suite where `replace` has not been changed to `update`. This change is made in [lf-lang/lingua-franca#2499](https://github.com/lf-lang/lingua-franca/pull/2499).

I think you have to change `lingua-franca-ref.txt` to point to the branch of lingua-franca, then ensure all the tests pass, merge with main, then merge the lingua-franca branch, then update `lingua-franca-ref.text` to point to master again, then merge again into main, then update the reactor-c pointer in master, and merge that update there.  Yes, it takes four merges to make one change.  We really shouldn't have separated out reactor-c into a separate repo.  Maybe we should revisit that and put it back in lingua-franca?

lsk567:
Yes, I think bringing reactor-c (and other runtimes) back into the lingua-franca repo is a great idea. The monorepo style makes things easier to maintain, as shown in reactor-uc.

edwardalee (changes requested):
As I stated in my review of the companion lingua-franca PR, maybe we should stay backward compatible. The documentation could read as follows:
```
 * The `update` policy drops the preceding event and queues the new event, thereby
 * updating both the tag and the payload of the event.
 * The `replace` policy will attempt to replace the payload of
 * the previously scheduled event. Unless the previous event has already been
 * handled, its gets assigned the value (the payload) of the new event, but conservs
 * the tag of the previous event. If the
 * previous event has already been popped off the event queue, the
 * `defer` policy is used.
```

lsk567:
Yes, I think this is a good consideration. Just revert the changes.

edwardalee (approved):
Modulo minor wording changes, LGTM!

PR #533, opened 2025-06-11
Use of snprintf instead of vulerable sprintf for code security best practices.

hokeun:


core/federated/network/socket_common.c
```diff
@@ -235,7 +235,7 @@ int connect_to_socket(int sock, const char* hostname, int port) {
     }
     // Convert port number to string.
     char str[6];
-    sprintf(str, "%u", used_port);
+    snprintf(str, sizeof(str), "%u", used_port);
```
edwardalee, line 238: This doesn't look right to me.  `str` is a pointer, so `sizeof(str)` is the size of a pointer, which is 8 bytes on a 64-bit machine. This will cause a buffer overrun.

trace/impl/src/trace_impl.c
```diff
@@ -261,9 +261,9 @@ void lf_tracing_global_init(char* process_name, char* process_names, int fedid,
   process_id = fedid;
   char filename[100];
   if (strcmp(process_name, "rti") == 0) {
-    sprintf(filename, "%s.lft", process_name);
+    snprintf(filename, sizeof(filename), "%s.lft", process_name);
```
edwardalee, line 264: Same problem.

trace/impl/src/trace_impl.c
```diff
@@ -261,9 +261,9 @@ void lf_tracing_global_init(char* process_name, char* process_names, int fedid,
...
   } else {
-    sprintf(filename, "%s_%d.lft", process_name, process_id);
+    snprintf(filename, sizeof(filename), "%s_%d.lft", process_name, process_id);
```
edwardalee, line 266: Same problem.

util/tracing/influxdb.h
```diff
@@ -357,13 +357,12 @@ int post_curl(influx_v2_client_t* c, ...) {
   curl_easy_setopt(curl, CURLOPT_URL, url_string);
   free(url_string);
 
-  char* token_string = (char*)malloc(120 * sizeof(char));
-  sprintf(token_string, "Authorization: Token %s", c->token);
+  char token_string[120];
+  snprintf(token_string, sizeof(token_string), "Authorization: Token %s", c->token);
```
edwardalee, line 361: Same problem.

edwardalee (changes requested):
I don't think `sizeof` is being used correctly here.

core/federated/network/socket_common.c
edwardalee: Hmm.  I couldn't find anything more authoritative about this. Seems brittle.  Would it be better to use a `#define` instead?

hokeun: The behavior `sizeof` operator in the C language is well defined. It universally works as it's defined in the C language standard - thus, it is nothing hacky, brittle, or platform-dependent. This may be a more authoritative document: https://learn.microsoft.com/en-us/cpp/c-language/sizeof-operator-c?view=msvc-170 

Using #define constant like #define PORT_STR_SIZE 6 also works. Still, I believe it's safer to do:
```
#define PORT_STR_SIZE 6
char str[PORT_STR_SIZE];
snprintf(str, sizeof(str), "%u", used_port);
```
than doing:
```
#define PORT_STR_SIZE 6
char str[PORT_STR_SIZE];
snprintf(str, PORT_STR_SIZE, "%u", used_port);
```
I'm fine with either option above, with a preference to the former.

edwardalee (approved):
Given the explanations, LGTM!

PR #540, opened 2025-08-01
Fix for race condition creating a possible segfault

### LINKED ISSUE #539: Federate can segfault during shutdown
Occasionally, when a federate is shutting down normally, a segfault can occur, apparently while shutting down socket connections. It happens quite rarely, but can be reliably reproduced in master and in release 0.10.0 by using the `run_until_fail.sh` script below to repeatedly run the `test/C/src/federated/SmallDelayDecentralized.lf` test case. It takes me a few hundred runs.

It appears that the background thread `update_ports_from_staa_offsets` was continuing to execute after the data structures it was using had been freed during shutdown. I also found some race conditions that could result in a negative socket ID being passed to socket read and write functions. This PR corrects those race conditions, ensuring that a negative ID is never passed to such functions.  PR #540 fixes these and also removes an unused `mutex` argument to the function `read_from_socket_fail_on_error`.

The `run_until_fail.sh` script:

```sh
#!/bin/bash

# Check if a program is provided
if [ -z "$1" ]; then
    echo "Usage: $0 <command>"
    exit 1
fi

# Initialize run counter
count=0

# Run the program in a loop
while true; do
        echo "***************************************** $count"
    # Execute the program
    "$@"
    
    # Check if the program failed
    if [ $? -ne 0 ]; then
        echo "************************* Program failed after $count successful runs."
        exit 1
    fi
    
    # Increment the run counter
    ((count++))
done
```

edwardalee:
This seems to fix #539.


Jakio815:
I had a look at the issue. I am currently running the script you gave to reproduce the seg fault. I did get the script to exit without the `logging: DEBUG` option, but didn't get it when I'm using it (Done 400 runs).

The changes look good. I probably think what happened is that the referenced value got changed while the `shutdown()` function. 

But I saw the sourcecode of shutdown() from `glibc-2.42`, `shutdown.c`,
```
#include <sys/socket.h>
#include <socketcall.h>

int
__shutdown (int fd, int how)
{
#ifdef __ASSUME_SHUTDOWN_SYSCALL
  return INLINE_SYSCALL_CALL (shutdown, fd, how);
#else
  return SOCKETCALL (shutdown, fd, how);
#endif
}
weak_alias (__shutdown, shutdown)
```

And this means, even referenced value changes shouldn't change the `fd` inside the function (because it should be a copy).
So, actually that can't happen...

Anyway the PR looks good to me. Do you have any further ideas? @edwardalee

Jakio815 (approved):

PR #554, opened 2026-01-01
Use `find_package` to link against external trace plugin.

lsk567:
The overall goal of this PR is to support a [new trace plugin](https://github.com/lf-lang/lf-trace-xronos) that streams data to the Xronos dashboard.

The existing mechanism to link against a trace plugin library is through `find_library`, which requires the plugin library to be located at the root of the generated C project (i.e., `LF_ROOT`). This requires using the `files` target property to copy the trace plugin library to `LF_ROOT`.

While this approach works for a small, self-contained library, this approach would not work if a library has its own dependencies, as the library needs to link against them, which `find_library` does not do.

A more robust approach is to pull in the plugin library dependencies using cmake's `find_package()` or using a custom cmake file via the `cmake-include` target property (more dev-friendly).

This PR:
1. uses `find_package` instead of `find_library` and adds a message for the user to provide a cmake file if `find_package` does not succeed;
2. makes the default plugin compilable using `build.sh`.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -53,49 +53,7 @@ add_library(reactor-c)
 target_sources(reactor-c PRIVATE ${REACTORC_SOURCES})
 lf_enable_compiler_warnings(reactor-c)
 
-if(DEFINED LF_TRACE)
-    include(${LF_ROOT}/trace/api/CMakeLists.txt)
-    target_link_libraries(reactor-c PUBLIC lf::trace-api)
-    # If LF_TRACE_PLUGIN is set, treat it as a CMake package name and try to locate it via find_package().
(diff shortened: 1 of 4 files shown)


edwardalee (approved):
This looks good to me, although without an example, it is hard to see how a user could figure out how to use the trace plugin mechanism.

lhstrh:
Is it really true that the `files` target property must be used? I haven't seen that in examples. When you build a shared library and install it (i.e., put it in a location where `find_library` will find it, such as `/usr/local/lib`), no files need to be copied.

If transitive dependencies are the concern, would `find_package` do the trick?

edwardalee (approved):
LGTM, modulo a small bug in the cmake file which I've identified.

PR #559, opened 2026-02-10
Fix for the wrong return value of `lf_check_deadline`

byeonggiljun:
This PR fixes an error that `lf_check_deadline()` with the argument `invoke_deadline_handler` of `false` returns `false` even if the reaction has passed the deadline.


hokeun (approved):
Thanks for spotting and fixing this error!

PR #573, opened 2026-04-01
Continuation of Fix length calculation for federate received array

edwardalee:
The merge of https://github.com/lf-lang/reactor-c/pull/572 was premature.
This is a continuation with the following change:
Set element_size to 1 on opaque byte payloads


edwardalee:
Abandoning this approach. It's a dead end.

PR #591, opened 2026-07-27
Refactored freeing of federate struct to address FIXME

edwardalee:
This PR consolidates freeing of the federate struct into a single function and addresses FIXMEs where there was incomplete freeing.

This also fixes an error accidentally introduced in transient-fed to allow -nt 0 and correct a seg faulting RTI.

This PR also fixes the logic in the neighborhood validation code, where the checks being performed were actually no-ops (spotted by Copilot).

### DIFF
--- a/core/federated/RTI/main.c
+++ b/core/federated/RTI/main.c
@@ -364,8 +364,8 @@ int process_args(int argc, const char* argv[]) {
     usage(argc, argv);
     return 0;
   }
-  if (rti.number_of_transient_federates > rti.base.number_of_scheduling_nodes) {
-    lf_print_error("--number_of_transient_federates cannot be higher than the number of federates.");
+  if (rti.number_of_transient_federates >= rti.base.number_of_scheduling_nodes) {
+    lf_print_error("--number_of_transient_federates must be less than the number of federates.");
     usage(argc, argv);
     return 0;
   }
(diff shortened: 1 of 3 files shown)


core/federated/RTI/rti_remote.c
```diff
@@ -2072,10 +2070,14 @@ static int receive_connection_information(net_abstraction_t fed_net, uint16_t fe
       }
       if (reject) {
         if (temp_fed != hot_swap_federate) {
-          free(temp_fed);
+          free_federate_info(temp_fed);
         }
         return 0;
```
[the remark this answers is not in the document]
edwardalee: Fixed in aceb44c21.

core/federated/RTI/main.c
ChadliaJerad: I think that a transient-only federation should not be allowed because the `start_time` is defined by persistent federates.
Supporting this corner case would require more careful thought.

ChadliaJerad (changes requested):
These are great memory cleanups, thanks!

Regarding the transient-only federation support (see the comment), it should be done independently.

core/federated/RTI/main.c
edwardalee: Agreed. Addressed in 7020175be

ChadliaJerad (approved):

PR #592, opened 2026-08-06
Transient fed refinement

edwardalee:
This PR attempts to clarify some comments and refactor some code for code duplication. It also fixes a number of typos and rewords comments.

### DIFF
--- a/core/federated/RTI/rti_common.c
+++ b/core/federated/RTI/rti_common.c
@@ -9,7 +9,6 @@
  */
 #if defined STANDALONE_RTI || defined LF_ENCLAVES
 #include "rti_common.h"
-#include <string.h> // for memset
 
 /**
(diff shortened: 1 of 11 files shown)


core/federated/RTI/rti_common.c
```diff
@@ -9,6 +9,7 @@
  */
 #if defined STANDALONE_RTI || defined LF_ENCLAVES
 #include "rti_common.h"
+#include <string.h> // for memset
```
ChadliaJerad: There is no call to memeset.

ChadliaJerad (approved):
LGTM!

core/federated/RTI/rti_common.c
edwardalee: Good catch. Fixed with 923b86642

PR #594, opened 2026-08-11
Priority-based deadline-monotonic scheduler

fra-p:
This pull request is the work @edwardalee  and I have been doing to implement priority scheduling of reactions with deadline for applications with the reactor-c runtime.

The scheduler can be used with federated and enclaved applications, as well as classic LF applications where independent reactions can be executed concurrently. The scheduler works with Linux-based platforms that provide the SCHED_FIFO and SCHED_RR priority-based thread schedulers.

# Algorithm description and usage
This PR introduces an intermediate layer of scheduling that configures the priority of the LF worker threads through the POSIX pthread API so that to prioritize the execution of reactions with tight deadlines. The priority assignment follows a deadline-monotonic rule for which smaller deadlines correspond to higher priority values. The conversion of deadlines into priority values is realized with the priority function that has the shape of a decreasing exponential function, whose parameters are chosen to minimize the number of collisions (similar deadline values that are mapped to the same priority value). The co-domain of the function is [2,98], because 1 is reserved for reactions with no deadline (lowest priority value) and 99 for sleeping threads (highest priority value to make sure their wake up is not delayed by other threads).

The scheduler is enabled with the following attribute:
`@platform(value = "posix", scheduler = "rt-fifo")`

Only the POSIX platform is currently supported. The supported values for the `scheduler` parameters are:

- `rt-fifo`: use the SCHED_FIFO scheduling class (priority-based, requires to be launched with sudo rights)
- `rt-rr`: use the SCHED_RR scheduling class (priority-based, requires to be launched with sudo rights)
- `normal`: use the default Linux scheduler (CFS, not real-time). `normal` is the default scheduler when no scheduler parameter is provided.

If the `platform` attribute is specified before the main reactor, the whole federation will use that scheduler; if it is specified before a federate, that federate will use that scheduler (enables support for multi-platform federation).

# Solved CI problems
## Deadline inheritance algorithm
This PR includes an enhancement of the deadline inheritance algorithm to avoid priority inversion. Consider this federated application:
Screenshot from 2026-08-13 11-16-41

The NetworkSender reactor added through the AST transformation between the FastLoop and the StateEstimator inherited a deadline of 60 ms from reaction se2. At tag 110 ms, the sensor timer fires and the NetworkSender is triggered, but this cannot execute because the execution of reaction se2, triggered at tag 100 ms, hasn't completed yet and cannot be preempted because the two reactions have the same priority (same deadline). The Actuator, since it has a higher level than the NetworkSender, cannot execute until the NetworkSender finishes. A deadline violation on the actuator occurs, because se2 executes for 40 ms.

We decided to modify the algorithm so that a reaction takes the minimum deadline of all reactions within the federate (or enclave) that have a higher level: reactions like the actuator will not be blocked by reactions with lower level and lower priority.

This modification broke the following CI test that makes use of the old GEDF dispatcher in a non-federated application:
Screenshot from 2026-08-13 11-24-53

Task2 inherits the 25 ms deadline, so the test expects it to be executed before Task1, which inherits a deadline of 50 ms. With the modified deadline inheritance algorithm, Task1 inherits the 25 ms deadline too, because Detector2 has a higher level but a tighter deadline. Task1 and Task2 end up having the same deadline, and the LF scheduler picks Task1 as first, which makes the test fail. The application is not federated and does not use the new priority scheduler, but the old GEDF dispatcher.

We restricted the new extra tightening pass of the deadline inheritance algorithm to only federated applications that use the new priority scheduler, because it looks like these two conditions are necessary to trigger the priority inversion problem found in the controller with state example: federation is required to have a current tag reaction be blocked by a previous tag one, and the new priority scheduler forces a deterministic ordering of reactions based on their deadline (the default CFS/EEVDF are fair and would interleave the NetworkSender with se2). With this restriction, the test passes.

## New core pinning attribute
The PR features the new `@cores` attribute that configures the OS thread scheduler with the core IDs where the LF application is allowed to run. In addition to a simple list of core IDs separated by commas, we defined a "range" syntax to map an application to a contiguous set of cores in a convenient way: `@cores(1..4)` means that the application can execute on cores 1, 2, 3, and 4 (equivalent to `@cores(1, 2, 3, 4)`). The new syntax is defined in the Lingua Franca xtext file.

An issue was encountered with C++ tests that were using the "..." syntax in target code blocks. The lexer would match ".." before the single ".", thus creating parsing problems. ".." was added to the Tokens list allowed in target code blocks, otherwise the parsing would fail.

A similar problem occurred with "import" statements using relative paths, where ".." was interpreted as "range" instead of "previous folder". ".." was then added to the allowed FSName tokens.


low_level_platform/api/platform/lf_arduino_support.h
```diff
@@ -113,4 +113,9 @@ typedef void* lf_thread_t;
 // Arduinos are embedded platforms with no command line interface
 #define NO_CLI
 
+// Priority values for the abstract LF scheduling API (no-ops on Arduino).
+// Range matches LF_SCHED_MIN_PRIORITY..LF_SCHED_MAX_PRIORITY.
+#define LF_SLEEP_PRIORITY 99      // Highest priority when waiting for physical time
+#define LF_NO_DEADLINE_PRIORITY 0 // Lowest priority for reactions without deadlines
```
edwardalee: Do these make sense for Arduino? Also, the POSIX support sets this priority to 1.

low_level_platform/api/platform/lf_flexpret_support.h
```diff
@@ -48,6 +48,11 @@ typedef fp_cond_t lf_cond_t;
 #define NO_CLI
 #define MINIMAL_STDLIB
 
+// Priority values for the abstract LF scheduling API (no-ops on FlexPRET).
+// Range matches LF_SCHED_MIN_PRIORITY..LF_SCHED_MAX_PRIORITY.
+#define LF_SLEEP_PRIORITY 99      // Highest priority when waiting for physical time
+#define LF_NO_DEADLINE_PRIORITY 0 // Lowest priority for reactions without deadlines
```
edwardalee: Do these make sense for FlexPRET? Also, the POSIX support sets this priority to 1.

low_level_platform/api/platform/lf_zephyr_support.h
```diff
@@ -35,4 +35,9 @@ void _lf_initialize_clock_zephyr_common();
 
 #endif // !LF_SINGLE_THREADED
 
+// Priority values for Zephyr thread scheduling
+// Zephyr uses lower numbers for higher priority (negative for cooperative, positive for preemptive)
+#define LF_SLEEP_PRIORITY -16      // Highest priority when waiting for physical time
+#define LF_NO_DEADLINE_PRIORITY 14 // Lowest priority for reactions without deadlines
```
edwardalee: Given the reversed polarity here (lower numbers have higher priority), will the priority evaluation function still work?

core/threaded/reactor_threaded.c
```diff
@@ -35,6 +35,26 @@
...
+ */
+static inline interval_t lf_reaction_scheduling_deadline_ns(const reaction_t* reaction) {
+  return (interval_t)(reaction->index >> 16);
+}
+
+/**
+ * If two consecutive lf_time_physical() reads differ by more than this, assume the worker was
+ * preempted between the reads and recompute time-until-deadline with the later timestamp.
+ */
+#define LF_PREEMPTION_DETECT_THRESHOLD MSEC(1)
```
edwardalee: Is this still used? I thought we weren't using time-until-deadline anymore.

edwardalee (commented):
Looks pretty good, but I've registered some questions.

low_level_platform/impl/src/lf_zephyr_support.c
```diff
@@ -147,7 +147,33 @@ int lf_thread_id() { return *((int*)k_thread_custom_data_get()); }
...
+    if (core_ids[i] < 0 || core_ids[i] >= available) {
+      continue; // Skip invalid core IDs
+    }
+    ret = k_thread_cpu_mask_enable(thread, core_ids[i]);
+    if (ret != 0) {
+      return ret;
+    }
+  }
+
+  return 0;
```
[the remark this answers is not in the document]
fra-p: Fixed as per suggestion

low_level_platform/api/low_level_platform.h
```diff
@@ -185,14 +185,22 @@ typedef struct {
...
- * @param thread The thread
- * @param cpu_number the CPU ID
- * @return 0 on success, platform-specific error number otherwise.
+ * On platforms that support CPU affinity (e.g., Linux), the calling thread's
+ * affinity mask is set to include ALL cores listed in `core_ids`. The OS
+ * scheduler then decides which of those cores to run the thread on.
+ *
+ * On platforms that only support pinning to a single core (e.g., Zephyr),
+ * a round-robin selection is used: `core_ids[lf_thread_id() % num_core_ids]`.
+ *
```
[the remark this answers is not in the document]
fra-p: The comment referred to an old implementation of `lf_thread_set_cpu`, so it was wrong. The suggestion is wrong as well, since Zephyr is implemented as a mask now. Updated the comment to reflect the actual behavior on Zephyr that is the same as Linux

low_level_platform/api/platform/lf_zephyr_support.h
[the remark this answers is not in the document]
fra-p: Those macros were removed from every platform-specific low-level platform header and defined in the generic low-level platform header, so that they apply to all platforms. These values are the logical priority values that LF uses. Each platform-specific low-level platform source file uses the `map_priorities` function that maps the LF range into a platform-specific one.

low_level_platform/api/platform/lf_windows_support.h
```diff
@@ -43,4 +43,9 @@ typedef HANDLE lf_thread_t;
 // Use 64-bit times and 32-bit unsigned microsteps
 #include "lf_tag_64_32.h"
 
+// Priority values for Windows thread scheduling
+// Windows thread priorities range from -15 to 15 (THREAD_PRIORITY_IDLE to THREAD_PRIORITY_TIME_CRITICAL)
+#define LF_SLEEP_PRIORITY 15        // Highest priority when waiting for physical time
+#define LF_NO_DEADLINE_PRIORITY -15 // Lowest priority for reactions without deadlines
```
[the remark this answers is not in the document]
fra-p: Those macros were removed from every platform-specific low-level platform header and defined in the generic low-level platform header, so that they apply to all platforms. These values are the logical priority values that LF uses. Each platform-specific low-level platform source file uses the `map_priorities` function that maps the LF range into a platform-specific one.

low_level_platform/impl/src/lf_windows_support.c
```diff
@@ -177,16 +177,17 @@ int lf_thread_join(lf_thread_t thread, void** thread_return) {
...
   return -1;
 }
 
 int lf_thread_set_priority(lf_thread_t thread, int priority) {
   (void)thread;   // Suppress unused variable warning.
   (void)priority; // Suppress unused variable warning.
-  return -1;
+  // Real-time scheduling API not implemented for Windows, return success as no-op
+  return 0;
 }
```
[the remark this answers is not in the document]
fra-p: Fixed as suggested

core/threaded/reactor_threaded.c
```diff
@@ -864,6 +901,51 @@ static void _lf_worker_do_work(environment_t* env, int worker_number) {
...
+    int num = (LF_NUMBER_OF_CORES > available) ? available : LF_NUMBER_OF_CORES;
+    int auto_core_ids[num];
+    for (int i = 0; i < num; i++) {
+      auto_core_ids[i] = available - 1 - i;
+    }
+    int ret = lf_thread_set_cpu(auto_core_ids, num);
+    if (ret != 0 && ret != -1) {
+      LF_PRINT_LOG("Warning: Could not set CPU affinity (error %d).", ret);
+    }
+  }
```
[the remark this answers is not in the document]
fra-p: LF_NUMBER_OF_CORES has been removed, so the comment does not apply anymore.

low_level_platform/impl/src/lf_macos_support.c
```diff
@@ -21,9 +21,18 @@
...
+  return -1;
+}
 
-int lf_thread_set_priority(lf_thread_t thread, int priority) { return -1; }
+int lf_thread_set_priority(lf_thread_t thread, int priority) {
+  (void)thread;   // Suppress unused variable warning.
+  (void)priority; // Suppress unused variable warning.
+  // Real-time scheduling API not implemented for macOS, return success as no-op
+  return 0;
+}
```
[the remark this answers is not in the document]
fra-p: Fixed as suggested

core/threaded/reactor_threaded.c
fra-p: You're right, we're not using time-until-deadline anymore, so this macro is obsolete. I removed it.

low_level_platform/api/platform/lf_zephyr_support.h
fra-p: The implementation of `lf_thread_set_priority` in lf_zephyr_support.c uses a utility function called `map_priorities` that maps the Linux [0,99] range into a Zephyr-compatible one, and it also takes care of reversing the polarity. This function already exists in the main branch, it's not an addition of my PR. I guess it should work, but I haven't tried it.

low_level_platform/api/platform/lf_arduino_support.h
fra-p: Those macros have been removed from every platform-specific low-level platform header and defined in the generic low-level platform header, so that they apply to all platforms. These values are the "abstract" priority values that LF uses. Each platform-specific low-level platform source file uses the `map_priorities` function that maps the LF range into a platform-specific one. LF_NO_DEADLINE_PRIORITY has been standardized to 1. Arduino does not support priority setting, so it does not use these macros.

low_level_platform/api/platform/lf_flexpret_support.h
fra-p: Those macros have been removed from every platform-specific low-level platform header and defined in the generic low-level platform header, so that they apply to all platforms. These values are the "abstract" priority values that LF uses. Each platform-specific low-level platform source file uses the `map_priorities` function that maps the LF range into a platform-specific one. LF_NO_DEADLINE_PRIORITY has been standardized to 1. FlexPRET does not support priority setting, so it does not use these macros.

core/threaded/reactor_threaded.c
```diff
@@ -35,6 +35,20 @@
...
 extern tag_t effective_start_tag;
 
+#ifdef LF_THREAD_POLICY
+// Function generated by the C code generator to compute priority from deadline.
+extern int get_priority_value(interval_t rel_deadline);
+
+/**
+ * Scheduling deadline (ns) for OS priority: upper bits of reaction->index, set by codegen via
+ * lf_combine_deadline_and_level from the graph’s inferred deadline (same as declared when nothing
+ * tighter is inherited from downstream).
```
edwardalee: This comment is unclear. 
```suggestion
 * Return the deadline of the specified reaction in ns.
 * This is extracted from the upper bits of `reaction->index`, which is set by code generator via
 * `lf_combine_deadline_and_level` from the reaction's inferred deadline (which is the same as
 * the declared deadline when nothing tighter is inherited from downstream).
```

core/threaded/reactor_threaded.c
```diff
@@ -182,6 +196,11 @@ bool wait_until(instant_t wait_until_time, lf_cond_t* condition) {
       return true;
     }
 
+    // Set highest priority before sleeping so we wake up quickly
+    lf_thread_set_priority(lf_thread_self(), LF_SLEEP_PRIORITY);
+    // Ignoring the return value: optimization for platforms that support priority setting,
+    // but it does not affect platforms that don't support it.
+
```
edwardalee: Should this also be wrapped in `#ifdef LF_THREAD_POLICY`?

core/threaded/reactor_threaded.c
```diff
@@ -864,6 +897,37 @@ static void _lf_worker_do_work(environment_t* env, int worker_number) {
...
+  // Configure thread scheduling policy and CPU affinity if specified
+#ifdef LF_THREAD_POLICY
+  {
+    lf_scheduling_policy_t policy;
+    policy.policy = LF_THREAD_POLICY;
+    policy.priority = LF_SCHED_MAX_PRIORITY;
+    policy.time_slice = 0;
+    int ret = lf_thread_set_scheduling_policy(lf_thread_self(), &policy);
+    if (ret != 0) {
+      if (LF_THREAD_POLICY != LF_SCHED_FAIR) {
```
edwardalee, line 913: This is comparing two compile-time definitions, which need not be compared at runtime. Use `#if ... #else`? But also, this raises the question of whether all the `#ifdef LF_THREAD_POLICY` should be `#if LF_THREAD_POLICY != LF_SCHED_FAIR`?

edwardalee (changes requested):
Minore changes requested plus some questions.

core/CMakeLists.txt
```diff
@@ -181,6 +181,14 @@ define(SCHEDULER)
...
 define(WORKERS_NEEDED_FOR_FEDERATE)
 define(LF_ENCLAVES)
+define(LF_THREAD_POLICY)
+
+# LF_CORE_IDS_INIT needs special handling since it contains braces and commas.
+# It is set to e.g. {0,1,2,3} by the @cores attribute.
+if(DEFINED LF_CORE_IDS_INIT)
+    message(STATUS "LF_CORE_IDS_INIT=${LF_CORE_IDS_INIT}")
+    target_compile_definitions(reactor-c PUBLIC "LF_CORE_IDS_INIT=${LF_CORE_IDS_INIT}")
+endif()
```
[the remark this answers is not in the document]
fra-p: The `define()` macro expands to

```
macro(define X)
    if(DEFINED ${X})
        message(STATUS ${X}=${${X}})
        target_compile_definitions(reactor-c PUBLIC ${X}=${${X}})
    endif(DEFINED ${X})
endmacro()
```

and it does not wrap values in quotes (that's what `defineString()` does). LF applications in fact compile correctly and the scheduling policy selection works; it wouldn't if the scheduling policy values were set as strings.

core/threaded/reactor_threaded.c
```diff
@@ -864,6 +897,37 @@ static void _lf_worker_do_work(environment_t* env, int worker_number) {
...
+  }
+#endif
+
+#ifdef LF_CORE_IDS_INIT
+  {
+    static int core_ids[] = LF_CORE_IDS_INIT;
+    int ret = lf_thread_set_cpu(core_ids, sizeof(core_ids) / sizeof(core_ids[0]));
+    if (ret != 0 && ret != -1) {
+      LF_PRINT_LOG("Warning: Could not set CPU affinity (error %d).", ret);
+    }
```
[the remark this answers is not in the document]
fra-p: The platforms that do not support core pinning return -1, so that value is ignored in reactor_threaded.c for code optimization purposes. I defined the `LF_THREAD_CPU_INVALID_CORE` macro that is the error code (-512) returned by the core pinning when an invalid core ID is entered. I then left line 925 of reactor_threaded.c unchanged.

test/scheduling/scheduling_api_test.c
```diff
@@ -13,10 +13,13 @@
...
-  if (res != 0) {
-    lf_print_error_and_exit("lf_thread_set_cpu failed with %d", res);
+  // Set the CPU affinity using specific core IDs
+  {
+    int core_ids[] = {0, 1};
+    res = lf_thread_set_cpu(core_ids, 2);
+    if (res != 0 && res != -1) {
+      lf_print_error_and_exit("lf_thread_set_cpu failed with %d", res);
+    }
   }
```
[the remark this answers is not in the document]
fra-p: Fixed the test as suggested

low_level_platform/api/low_level_platform.h
```diff
@@ -185,14 +189,20 @@ typedef struct {
...
+ * OS scheduler then decides which of those cores to run the thread on.
+ * Platforms that do not support affinity return -1.
+ *
+ * @param core_ids An array of CPU core IDs to allow the thread to run on.
+ * @param num_core_ids The number of entries in the core_ids array.
+ *                     If 0 or core_ids is NULL, no pinning is done.
+ * @return 0 on success, -1 if no pinning needed or invalid, platform-specific error otherwise.
  */
-int lf_thread_set_cpu(lf_thread_t thread, size_t cpu_number);
+int lf_thread_set_cpu(int* core_ids, size_t num_core_ids);
```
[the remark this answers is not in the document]
fra-p: Fixed as suggested

core/threaded/reactor_threaded.c
fra-p: Changed as suggested

core/threaded/reactor_threaded.c
fra-p: LF_SCHED_FAIR, LF_SCHED_PRIORITY, and LF_SCHED_TIMESLICE are enum members, not macros, so the pre-processor does not evaluate expressions containing them: `#if LF_THREAD_POLICY != LF_SCHED_FAIR` evaluates both sides as 0 and always takes the FAIR branch. The standard `if` is evaluated when the enum already exists. The runtime `if` is already constant-folded at -O2. 

On the broader question, I followed the principle that the scheduling code is enabled when the `@platform` attribute defines any scheduling policy, even the "fair" one that is set by passing `normal` to the `scheduler` parameter of the attribute. `LF_THREAD_POLICY` is then defined whenever the `scheduler` parameter is set.  The scheduling API implementation then does not set priorities when the policy is `normal`.

core/threaded/reactor_threaded.c
fra-p: Wrapped as suggested

PR #596, opened 2026-08-19
Retry connect() on a fresh socket instead of reusing the failed fd

### LINKED ISSUE #595: Federated: connect_to_socket() retries connect() on the same fd after failure; retry loop never recovers on macOS
## Summary

`connect_to_socket()` in `network/impl/src/socket_common.c` retries `connect()` on the **same file descriptor** after a failed attempt. POSIX leaves a socket's state unspecified after a failed `connect()`, and on macOS the practical consequence is that one transient failure makes every subsequent retry fail: the federate loops `"Could not connect. Will try again every 500000000 nanoseconds."` for the entire `CONNECT_TIMEOUT` and then gives up with `Failed to connect to RTI` — while the RTI is demonstrably reachable from the same host the whole time. On Linux the same-fd retry happens to recover, which masks the bug in single-machine testing.

## The code

`network/impl/src/socket_common.c`, `connect_to_socket()` (present in the v0.13.0 release and unchanged on `main` @ b6fd23c):

```c
while (1) {
    if (CHECK_TIMEOUT(start_connect, CONNECT_TIMEOUT)) { ... break; }
    ...
    ret = connect(sock, result->ai_addr, result->ai_addrlen);
    ...
    if (ret < 0) {
      lf_sleep(CONNECT_RETRY_INTERVAL);
      lf_print_warning("Could not connect. Will try again every ...");
      continue;            // <-- retries connect() on the SAME fd
    }
}
```

POSIX (`connect`, System Interfaces): *"If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect."*

## Observed behavior

Federation of 2 federates across a LAN: RTI + one federate on Linux aarch64 (Ubuntu, kernel 6.17), the other federate on macOS arm64 (M2, Darwin 25.5). lfc 0.13.0, C and Python targets (both use this code path).

| Scenario | Result |
|---|---|
| macOS federate, first `connect()` to the RTI fails once (cause: transient; instrumented `errno` showed e.g. `ETIMEDOUT`) | retry loop never recovers: 30–70 warnings over minutes, then `FATAL ERROR: Failed to connect to RTI` |
| Concurrently, from the same macOS host, while the federate loops | `python socket.connect()` to the same RTI host:port succeeds instantly — single attempt, with `TCP_NODELAY` set, and 30/30 in a rapid burst; `nc` succeeds too |
| Linux federates started **before** the RTI (guaranteed first-connect failure), RTI started 6 s later | federates recover on a subsequent same-fd retry; federation starts and runs to completion (29/29 messages) — Linux's leniency masks the bug |

So: every fresh-socket connect from the affected machine succeeds while the federate's same-fd retries all fail. Restarting the federate process ("retry the whole program") also typically succeeds, which is the workaround we use.

## Reproduction

Any 2-federate program. Deterministic first-connect failure: start a federate before the RTI (or firewall one SYN). On Linux it recovers (masking); to see the failure you need a platform where a failed `connect()` poisons the fd — macOS in our testing. A cross-machine Linux RTI + macOS federate over a real LAN reproduced it for us within a few join attempts.

## Suggested fix

Recreate the socket on each retry instead of reusing the fd. Since callers hold the fd number, either restructure `connect_to_socket` to own socket creation and return the fd, or preserve the caller's fd number with `dup2`:

```c
if (ret < 0) {
  int fresh = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
  if (fresh >= 0) {
    int one = 1;
    setsockopt(fresh, IPPROTO_TCP, TCP_NODELAY, &one, sizeof(one));
    dup2(fresh, sock);   // sock now refers to a clean socket; old state discarded
    close(fresh);
  }
  lf_sleep(CONNECT_RETRY_INTERVAL);
  ...
  continue;
}
```

We ran with this patch applied to the generated sources; the retry loop then uses a clean socket per attempt as intended. (In one episode on macOS even fresh-socket retries from the federate process received `ETIMEDOUT` while concurrent Python connects succeeded — we could not identify that cause and it may be a separate issue; the same-fd retry defect above stands independently.)

Two small adjacent notes seen while debugging, mentioned for context only: the accept loops log `"Failed to accept the socket. Resource temporarily unavailable. Trying again."` in a busy spin on `EAGAIN` (log spam + CPU), and adding `errno`/`strerror` to the `"Could not connect"` warning (as we did locally) makes field diagnosis of this code path much easier.

remifan:
Fixes #595.

`connect_to_socket()` retried `connect()` on the same file descriptor after a failure. POSIX leaves the socket's state unspecified after a failed `connect()` ("Conforming applications should close the file descriptor and create a new socket before attempting to reconnect"), and on macOS the observed consequence is that every subsequent retry on that fd fails: one transient failure makes a federate loop `"Could not connect. Will try again..."` for the entire `CONNECT_TIMEOUT` and then give up with `Failed to connect to RTI` — while the RTI is demonstrably reachable (concurrent fresh-socket connects from the same host succeed throughout; details and the evidence table are in #595). On Linux the same-fd retry happens to recover, which masks the bug in single-machine testing.

This change:
- recreates the socket on each retry via the existing `create_real_time_tcp_socket_errexit()` (so `TCP_NODELAY`, and `TCP_QUICKACK` on Linux, are preserved), using `dup2()` to keep the fd number that callers hold;
- captures `errno` before it can be clobbered and includes `errno`/`strerror` in the retry warning, which made field diagnosis of this path possible for us.

Tested on a two-machine federation (RTI + federate on Linux aarch64, federate on macOS arm64, C target, lfc v0.13.0 with this runtime via `--external-runtime-path`): with the fix, a federate whose first connect fails recovers on a subsequent retry; we also run this patch in a production federation with transient join/leave/rejoin. The late-RTI scenario (federates started before the RTI) continues to work on Linux as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/network/impl/src/socket_common.c
+++ b/network/impl/src/socket_common.c
@@ -278,7 +278,12 @@ int connect_to_socket(int sock, const char* hostname, const struct in_addr* ip_a
       // Retry on a fresh socket, preserving the fd number that the caller
       // holds via dup2(). See #595.
       int fresh = create_real_time_tcp_socket_errexit();
-      dup2(fresh, sock);
+      if (dup2(fresh, sock) < 0) {
+        int dup_errno = errno;
+        close(fresh);
+        errno = dup_errno;
+        lf_print_error_system_failure("Failed to recreate socket with dup2() after connect() failure.");
+      }
       close(fresh);
       lf_sleep(CONNECT_RETRY_INTERVAL);
       lf_print_warning("Could not connect (errno=%d: %s). Will try again every " PRINTF_TIME
--- a/network/impl/src/socket_common.c
+++ b/network/impl/src/socket_common.c
@@ -268,9 +268,22 @@ int connect_to_socket(int sock, const char* hostname, const struct in_addr* ip_a
     }
 
     if (ret < 0) {
+      int connect_errno = errno;
+      // POSIX leaves the state of a socket unspecified after connect() fails:
+      // "If connect() fails, the state of the socket is unspecified. Conforming
+      // applications should close the file descriptor and create a new socket
+      // before attempting to reconnect." On some platforms (observed on macOS)
+      // retrying connect() on the same fd fails on every attempt, so a single
+      // transient failure otherwise becomes permanent until process restart.
+      // Retry on a fresh socket, preserving the fd number that the caller
+      // holds via dup2(). See #595.
+      int fresh = create_real_time_tcp_socket_errexit();
+      dup2(fresh, sock);
+      close(fresh);
       lf_sleep(CONNECT_RETRY_INTERVAL);
-      lf_print_warning("Could not connect. Will try again every " PRINTF_TIME " nanoseconds. Connecting to port %d.\n",
-                       CONNECT_RETRY_INTERVAL, used_port);
+      lf_print_warning("Could not connect (errno=%d: %s). Will try again every " PRINTF_TIME
+                       " nanoseconds. Connecting to port %d.\n",
+                       connect_errno, strerror(connect_errno), CONNECT_RETRY_INTERVAL, used_port);
       continue;
     } else {
       break;


edwardalee (approved):
LGTM, but the Copilot review suggests what looks like a useful change.  OK to apply this? The Python test failure looks like a spurious one to me. Looks like a transient network failure in the maven repository.

Jakio815 (approved):
LGTM!

